PR #5, opened 2023-09-06
A few renamings and bug fixes

semi-h:
- memblock_t is renamed as field_t because we use allocator and memblock exclusively for allocating scalar fields.
- cudaallocator and cudafield(cudamemblock) is renamed with underscore, I think it looks better and easier to read
- The rest is just few small bug fixes

### DIFF
--- a/src/vector3d.f90
+++ b/src/vector3d.f90
@@ -12,7 +12,8 @@ module m_slab
       class(field_t), pointer :: u1, u2, u3
       real :: xnu
    contains
-      procedure(field_op), deferred :: transport, div
+      procedure(field_op), deferred :: transport
+      procedure(field_op), deferred :: div
       procedure(transport_op), deferred :: transport_dir
       procedure, public :: u => get_component_ptr
    end type slab_t
--- a/src/vector3d.f90
+++ b/src/vector3d.f90
@@ -1,7 +1,7 @@
 module m_slab
    use m_stencil, only: stencil
    use m_tridiagsolv, only: tridiagsolv
-   use m_allocator, only: allocator_t, memblock_t
+   use m_allocator, only: allocator_t, field_t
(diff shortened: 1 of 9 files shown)


tlestang:
@semi-h could you add a description stating what you changed and why?

src/vector3d.f90
```diff
@@ -12,7 +12,8 @@ module m_slab
       class(field_t), pointer :: u1, u2, u3
       real :: xnu
    contains
-      procedure(field_op), deferred :: transport, div
+      procedure(field_op), deferred :: transport
+      procedure(field_op), deferred :: div
```
tlestang, line 16: That's to make nvfortran happy if I remember correctly?

tests/test_diffengine.f90
```diff
@@ -7,22 +7,29 @@ program test_diffengine
...
+  allocate(f_vec(SZ,n))
+  do j = 1, n
+     do i = 1, SZ
+        f_vec(i, j) = sin((j-1)*dx)
+     end do
+  end do
 
   ! First derivative with periodic boundary conditions
-  expected = [(cos((i-1)*dx), i=1,n)]
-  expected_vec = transpose(reshape(expected, [n, SZ], pad = expected))
```
tlestang, line 25: Was it not correct? In any case replacing this by a double loop is probably for the best.

tlestang (approved):

src/vector3d.f90
semi-h: Yes, it doesn't compile when you list them in the same line

tests/test_diffengine.f90
semi-h: My nvfortran compiler was giving an error here for some reason

PR #7, opened 2023-09-12
Backend structure

semi-h:
~First commit to showcase the backend structure in the new framework. It's possible that this won't compile, but it demonstrates the main idea.~

Subsequent commits added a lot of functionality to the codebase and we can run the transport equation on the CUDA backend now.

### DIFF
--- a/src/backend.f90
+++ b/src/backend.f90
@@ -37,24 +37,25 @@ module m_base_backend
       class(dirps_t), pointer :: xdirps, ydirps, zdirps
    contains
       procedure :: transeq
-      procedure :: run
       procedure(transeq_ders), deferred :: transeq_x
       procedure(transeq_ders), deferred :: transeq_y
       procedure(transeq_ders), deferred :: transeq_z
       procedure(transposer), deferred :: trans_x2y
       procedure(transposer), deferred :: trans_x2z
       procedure(sum9into3), deferred :: sum_yzintox
+      procedure(set_fields), deferred :: set_fields
+      procedure(get_fields), deferred :: get_fields
    end type base_backend_t
 
    abstract interface
-      subroutine transeq_ders(self, du, duu, d2u, u, v, w, dirps)
+      subroutine transeq_ders(self, du, dv, dw, u, v, w, dirps)
          import :: base_backend_t
          import :: field_t
          import :: dirps_t
          implicit none
 
          class(base_backend_t) :: self
-         class(field_t), intent(out) :: du, duu, d2u
+         class(field_t), intent(inout) :: du, dv, dw
          class(field_t), intent(in) :: u, v, w
          type(dirps_t), intent(in) :: dirps
       end subroutine transeq_ders
@@ -67,7 +68,7 @@ module m_base_backend
          implicit none
 
          class(base_backend_t) :: self
-         class(field_t), intent(out) :: u_, v_, w_
+         class(field_t), intent(inout) :: u_, v_, w_
          class(field_t), intent(in) :: u, v, w
       end subroutine transposer
    end interface
@@ -84,13 +85,37 @@ module m_base_backend
       end subroutine sum9into3
    end interface
 
+   abstract interface
+      subroutine set_fields(self, u, v, w, u_in, v_in, w_in)
+         import :: base_backend_t
+         import :: dp
+         import :: field_t
+         implicit none
+
+         class(base_backend_t) :: self
+         class(field_t), intent(inout) :: u, v, w
+         real(dp), dimension(:, :, :), intent(in) :: u_in, v_in, w_in
+      end subroutine set_fields
+
+      subroutine get_fields(self, u_out, v_out, w_out, u, v, w)
+         import :: base_backend_t
+         import :: dp
+         import :: field_t
+         implicit none
+
+         class(base_backend_t) :: self
+         real(dp), dimension(:, :, :), intent(out) :: u_out, v_out, w_out
+         class(field_t), intent(in) :: u, v, w
+      end subroutine get_fields
+   end interface
+
 contains
 
    subroutine transeq(self, du, dv, dw, u, v, w)
       implicit none
 
       class(base_backend_t) :: self
-      class(field_t), intent(out) :: du, dv, dw
+      class(field_t), intent(inout) :: du, dv, dw
       class(field_t), intent(in) :: u, v, w
 
       class(field_t), pointer :: u_y, v_y, w_y, u_z, v_z, w_z, &
@@ -154,28 +179,6 @@ contains
       call self%allocator%release_block(dw_z)
 
    end subroutine transeq
-
-   subroutine run(self)
-      !! This is the main subroutine that will start execution. This may be
-      !! moved to the main program but is here for now.
-      implicit none
-
-      class(base_backend_t) :: self
-      class(field_t), pointer :: u, v, w, du, dv, dw, udiv, p, px, py, pz
-
-      real :: dt, alpha
-
-      ! transport equation
-      call self%transeq(du, dv, dw, u, v, w)
-      ! time integration
-
-      !! pressure stuff
-      !call self%divergence(u, v, w, udiv)
-      !call self%poisson(udiv, p)
-      !call self%gradient(p, px, py, pz)
-      !! velocity correction
-
-   end subroutine run
 
 end module m_base_backend
 
--- a/src/backend.f90
+++ b/src/backend.f90
@@ -92,7 +92,8 @@ contains
       class(field_t), intent(out) :: du, dv, dw
       class(field_t), intent(in) :: u, v, w
 
-      class(field_t), pointer :: u_y, v_y, w_y, u_z, v_z, w_z, du_y, dv_y, dw_y, du_z, dv_z, dw_z
+      class(field_t), pointer :: u_y, v_y, w_y, u_z, v_z, w_z, &
+                                 du_y, dv_y, dw_y, du_z, dv_z, dw_z
 
       !1/2(nabla u curl u + u nabla u) + nu nablasq u
 
@@ -111,7 +112,7 @@ contains
       ! reorder data from x orientation to y orientation
       call self%trans_x2y(u_y, v_y, w_y, u, v, w)
       ! similar to the x direction, obtain derivatives in y.
-      call self%transeq_y(dv_y, du_y, dw_y, v_y, u_y, w_y, self%ydirps)
+      call self%transeq_y(du_y, dv_y, dw_y, u_y, v_y, w_y, self%ydirps)
 
       ! we don't need the velocities in y orientation any more, so release
       ! them to open up space.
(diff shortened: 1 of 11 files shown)


src/backend.f90
```diff
@@ -0,0 +1,174 @@
...
+      ! we don't need the velocities in y orientation any more, so release
+      ! them to open up space.
+      ! It is important that this doesn't actually deallocate any memory,
+      ! it just makes the corresponding memory space available for use.
+      call self%allocator%release_block(u_y)
+      call self%allocator%release_block(v_y)
+      call self%allocator%release_block(w_y)
+
+      ! just like in y direction, get some fields for the z derivatives.
+      u_z => self%allocator%get_block()
```
semi-h, line 120: Yes when we're doing the pointer assignment we actually need to specify bounds so that it matches the orientation we want. Its called bounds remapping and was introduced with the Fortran 2003 standard. I tested this with nvfortran and gfortran, haven't got any issues.

mathrack: In this case `u_y` is only a memory block and there is no need for a specific class ?

semi-h: Well actually it has to be u_y%data(1:SZ, 1:n_y, 1:n_ygroups) => ...

semi-h: Hmm actually we won't be able to do this. The data component is different for cuda and cpu backends. So here we just associate a field without any bounds, its just a pointer after all. Then in backends we use a pointer to a regular array or device array and point the data component with defining the bounds.

mathrack: Ok. Then the get_block function probably needs some arguments.

semi-h: x/y/zderps has got the dimensionality info and they're passed to the transeq_? subroutines so I think it is better to arrange it via derps. Otherwise the grid info is like duplicated both in field type and derps type and its extra work to make sure they're the same.

mathrack: Avoiding duplication makes sense. However, fields will be used for derivations / interpolations but also for other operations, such as IO. Thus, it could be relevant to include the grid/dimensionality info inside the field.

Without providing arguments to the subroutine `get_block`, I think the pointer u_y%data will not have the correct shape/bounds by default, at least right after the call to get_block. This could be an issue.

semi-h: https://henriquesd.medium.com/dry-kiss-yagni-principles-1ce09d9c601f

semi-h: You're trying to violate YAGNI and DRY :smiley:

semi-h: The size of the field is actually determined when we instantiate the allocator and is always the same size. And because backend needs to access allocator, allocator needs to be instantiated first.
https://github.com/xcompact3d/x3d2/blob/main/src/allocator.f90#L71
https://github.com/xcompact3d/x3d2/blob/main/tests/test_allocator.f90#L14

Each time we request a field exactly predetermined size of memory is provided to us. The only problem is when we switch pencil orientation the bounds differ a little for different dimensions.
x pencil data is (SZ, n_x, n_x_groups) and y pencil is (SZ, n_y, n_y_groups). And to support this we'll probably need to pad n_x, n_y, and n_z in the global grid sizes. So most of the time we'll have unneccessary regions in the field array. 
And this unnecessary region depends on wheter we're storing velocity or pressure, but we always have a way to access this information. If it is a u_y and then it is a velocity array in y orientation, and we can get this via yderps type. We need the n_y and n_y_groups even for executing a GPU kernel for example. If we try to obtain these sizes from the field type then it is not the straightforward way and more error prone in my opinion. There are many fields passed to transeq_y for example, and obtaining the grid size from one lucky field is not the best. This grid size is very important to transeq_y, and we pass this info already when we call transeq_y via yderps type.

src/backend.f90
```diff
@@ -0,0 +1,179 @@
...
+      end subroutine sum9into3
+   end interface
+
+contains
+
+   subroutine transeq(self, du, dv, dw, u, v, w)
+      implicit none
+
+      class(base_backend_t) :: self
+      class(field_t), intent(out) :: du, dv, dw
```
pbartholomew08, line 92: I think that technically this will create new fields, deallocating their `allocatable` members (which likely get reallocated later when set).

semi-h: I use intent(out) in the kernels as well and haven't really noticed a deallocation/allocation and an associated performance loss due to this. Still worth double checking with using intent(inout) instead, I'll let you know.

semi-h: Actually, do you expect such a deallocation/allocation when you pass an allocatable array into a subroutine and within the subroutine declare intent(out) for the dummy variable corresponding to the allocatable array?

semi-h: or this only applies to user defined types?

semi-h:
So as of now with the latest commit, the cuda backend that extends base backend also compiles (however extremely simplified and basically does nothing inside). Also, we have a time integrator that executes the algorithm within its subroutine called 'run'. All this is managed in the main file xcompact.f90, which is ultimately used for creating an executable.

The issue I noticed at this level is that our allocator is not advanced enough to deal with different nx, ny, nz. To fix this I need to introduce a simple functionality in the allocator using the 'bounds remapping' introduced in fortran 2003. I have already used this 'bounds remapping' technique with the nvidia compiler and gfortran and haven't got any issues.

src/backend.f90
pbartholomew08: It is complicated - if the _argument_ is given the attribute ``allocatable`` and ``intent(out)`` it will be deallocated on entry. An allocatable array that is passed otherwise will not be deallocated, e.g.

```
subroutine foo(a, b)
  integer, dimension(:), allocatable, intent(out) :: a ! Deallocated on entry
  integer, dimension(:), intent(out) :: b ! Not deallocated
  ...
```

For user defined types I think they will technically be reinitialised with ``intent(out)`` so if they contain an ``allocatable`` field it will be deallocated.

semi-h: Alright then we're fine for the kernels because I never use allocatable attribute for the kernel subroutines. But for the user defined types it is better to use intent(inout) to avoid reallocation. I didn't know about this at all, thanks!

src/cuda/backend.f90
```diff
@@ -0,0 +1,404 @@
...
+
+      select type(temp_du)
+      type is (cuda_field_t); du_dev => temp_du%data_d
+      end select
+      select type(temp_duu)
+      type is (cuda_field_t); duu_dev => temp_duu%data_d
+      end select
+      select type(temp_d2u)
+      type is (cuda_field_t); d2u_dev => temp_d2u%data_d
+      end select
```
semi-h, line 207: We obtain a device array pointer here and after this stage we can call kernels just like in the tests

semi-h:
I fixed the issue I mentioned on Monday about kernels not being executed. Now the transport equation runs in the backend structure.

Next few days I'll work on separating a solver class from backend class and then the PR be ready to merge.

src/common.f90
```diff
@@ -4,4 +4,79 @@ module m_common
...
+      integer, intent(in) :: xnproc, ynproc, znproc, nrank
+
+      integer :: i, ix, iy, iz
+
+      ! Set the pprev and pnext for each rank
+      i = 0
+      do iz = 1, znproc
+         do iy = 1, ynproc
+            do ix = 1, xnproc
+               if (nrank == i) then
```
Nanoseb, line 33: The 3 loops here could be avoided with something like:
```
ix = modulo(nrank, xnproc)
iy = modulo(nrank - ix*xnproc, ynproc)
iz = modulo(nrank - ix*xnproc - iy*ynproc*xnproc)
```

src/common.f90
```diff
@@ -4,4 +4,79 @@ module m_common
...
+      real(dp) :: Lx, Ly, Lz
+      real(dp) :: dx, dy, dz
+      integer :: nproc_x = 1, nproc_y = 1, nproc_z = 1
+      character(len=20) :: BC_x_s, BC_x_e, BC_y_s, BC_y_e, BC_z_s, BC_z_e
+   end type globs_t
+
+contains
+
+   subroutine set_pprev_pnext(xprev, xnext, yprev, ynext, zprev, znext, &
+                              xnproc, ynproc, znproc, nrank)
```
Nanoseb, line 20: Not sure how you want to go about tests, but this function would be well suited to have one as it is easy to make a typo with sign/variables and it is isolated from the rest of the codebase.

src/cuda/backend.f90
```diff
@@ -0,0 +1,423 @@
...
+      type is (cuda_tdsops_t); der2nd_sym => tdsops
+      end select
+
+      ! Copy halo data into buffer arrays
+      !self%u_send_s_dev(:, :, :) = u_dev(:, 1:4, :)
+      !self%u_send_e_dev(:, :, :) = u_dev(:, dirps%n - 3:dirps%n, :)
+      !self%v_send_s_dev(:, :, :) = v_dev(:, 1:4, :)
+      !self%v_send_e_dev(:, :, :) = v_dev(:, dirps%n - 3:dirps%n, :)
+      !self%w_send_s_dev(:, :, :) = w_dev(:, 1:4, :)
+      !self%w_send_e_dev(:, :, :) = w_dev(:, dirps%n - 3:dirps%n, :)
```
Nanoseb, line 215: To delete? or explain why it is commented out

src/xcompact.f90
```diff
@@ -0,0 +1,127 @@
...
+   allocate(w(SZ, globs%nx_loc, globs%n_groups_x))
+
+   ! GPU only
+   !allocate(cuda_allocator_t :: allocator)
+   !allocate(cuda_backend_t :: backend)
+
+   !cuda_backend = cuda_backend_t(allocator, xdirps, ydirps, zdirps)
+
+   !allocator = cuda_allocator_t([SZ, 512, 512*512/SZ])
+   !backend = cuda_backend_t(allocator, xdirps, ydirps, zdirps)
```
Nanoseb, line 110: should this be removed? as I understand there shouldn't be any implementation specifics calls in this part of the code

src/cuda/backend.f90
semi-h: We need these assignments for the halo data exchange,  these arrays are the buffer arrays for sending halo data. NVIDIA compiler complained for this assignment, I guess because we're using the members of the class via `self%`. I need to move this into a function and then if I pass the buffer arrays as device arrays the assignment in the function should work fine just like in the tests.

src/xcompact.f90
semi-h: We need to make the main file a bit more elaborate so that it can handle both backends. The problem is if you're not on a GPU cluster you probably don't have the NVIDIA compiler and then all the CUDA Fortran extentions NVIDIA compiler supports are not recognized by standard Fortran compilers. Our Cmake file compiles the cuda backend only when the compiler is NVIDIA compiler. Other times cuda backend is not even compiled and I think we'll need to use 'ifdef' stuff just in a few places in the main file to sort this out.

src/common.f90
Nanoseb: Sorry, just realised I made a mistake as modulo goes from `0` to `xnprox-1`, so there is likely some `+1` to put in there somewhere. Emphasising the need for tests :)

PR #11, opened 2023-10-23
Add a class for managing tridiagonal solver operators

semi-h:
This is an independent class that allows us to preprocess the tridiagonal solver based on typical boundary conditions and type of operations in Incompact3D.

The idea is, if there is a need to add a new type of tridiagonal solver in the codebase, we add all the relevant preprocessing here in this class and then we execute the universal kernel by supplying the coefficient arrays this class store.

~~- I'll add few more commits in this PR to add unit testing and remove the old operator functions.~~

- Added unit tests for the OpenMP backend
- Extended the tdsops_t class for the CUDA backend
- Use cuda_tdsops_t in CUDA tests
- Remove old derparams that was being used as a preprocessor


src/derparams.f90
```diff
@@ -1,10 +1,667 @@
 module m_derparams
-   use m_common, only: dp
+   use m_common, only: dp, pi
 
    implicit none
 
+   type :: tdsops_t
```
pbartholomew08, line 6: Add a comment - what is this type for?

src/derparams.f90
```diff
@@ -1,10 +1,667 @@
...
+      procedure :: deriv_1st, deriv_2nd, interpl_mid, stag_der_1st, preprocess
+   end type tdsops_t
+
+   interface tdsops_t
+      module procedure tdsops
+   end interface tdsops_t
+
 contains
 
+   type(tdsops_t) function tdsops(n, delta, operation, variation, n_halo, &
```
pbartholomew08, line 21: Add a comment - what does this function do?

src/derparams.f90
```diff
@@ -1,10 +1,667 @@
...
+
+   end function tdsops
+
+   subroutine deriv_1st(self, variation, bc_top, bc_bot, paire)
+      implicit none
+
+      class(tdsops_t), intent(inout) :: self
+      character(*), intent(in) :: variation
+      character(*), optional, intent(in) :: bc_top, bc_bot
+      logical, optional, intent(in) :: paire
```
pbartholomew08, line 76: Worth commenting procedure arguments

- what is variation? what values could it take?
- what are bc_top/bot? Are they related to the y-orientation or the top/bottom of the vector when viewed as a system of equations?
- what is `paire` (I know from Xcompact3d familiarity, but i think we should make this clear for the future)?

src/derparams.f90
```diff
@@ -1,10 +1,667 @@
...
+            !             d(wu)/dz, d(wv)/dz, dw/dz
+            self%dist_sa(1) = 0._dp
+            self%dist_sc(1) = 2*self%alpha
+            self%coeffs_t(1, :) = [0._dp, 0._dp, 0._dp, 0._dp, &
+                                   0._dp, &
+                                   2*afi, 2*bfi, 0._dp, 0._dp]
+            self%coeffs_t(2, :) = [0._dp, 0._dp, 0._dp, -afi, &
+                                   bfi, &
+                                   afi, bfi, 0._dp, 0._dp]
+         end if
```
pbartholomew08, line 138: Would it be neater to write these as helper subroutines? This subroutine would then be a "driver" and the utility subroutines easier to understand (and test) in isolation

src/derparams.f90
```diff
@@ -1,10 +1,667 @@
...
+         self%coeffs_b(n_halo-1, :) = [0._dp, 0._dp, 0._dp, -0.75_dp/dx, &
+                                       0._dp, &
+                                       0.75_dp/dx, 0._dp, 0._dp, 0._dp]
+      end select
+
+      call self%preprocess(dist_b)
+
+   end subroutine deriv_1st
+
+   subroutine deriv_2nd(self, variation, bc_top, bc_bot, paire, c_nu, nu0_nu)
```
pbartholomew08, line 200: See comments on `deriv_1st`

src/derparams.f90
```diff
@@ -1,10 +1,667 @@
...
+         self%dist_sa(n-3) = 2._dp/11._dp
+         self%dist_sc(n-3) = 2._dp/11._dp
+         self%coeffs_b(1, :) = self%coeffs_b(2, :)
+      end select
+
+      call self%preprocess(dist_b)
+
+   end subroutine deriv_2nd
+
+   subroutine interpl_mid(self, variation, from_to, bc_top, bc_bot, paire)
```
pbartholomew08, line 396: See comments on `deriv_X` subroutines

src/derparams.f90
```diff
@@ -1,10 +1,667 @@
...
+                           *self%dist_sa(i-1)
+         self%dist_sc(i) = self%dist_fw(i)*self%dist_sc(i)
+      end do
+      do i = self%n - 2, 2, -1
+         self%dist_sa(i) = self%dist_sa(i) - self%dist_sc(i)*self%dist_sa(i+1)
+         self%dist_bw(i) = self%dist_sc(i)
+         self%dist_sc(i) = -self%dist_sc(i)*self%dist_sc(i+1)
+      end do
+      ! this is not good
+      self%dist_fw(1) = 1.d0/(1.d0 - self%dist_sc(1)*self%dist_sa(2))
```
pbartholomew08, line 658: Is the intent to be type `d0` or `_dp`?

src/derparams.f90
semi-h: There is a similarity for the operator matrices accross diffrent types of operations, but we need to edit only a few entries based on the BC and the exact specification may differ in similar cases (for example der_1st, neumann and paire=0 is similar to der_2nd, neumann, paire=1, not paire=0). We can call this helper function to set things as we want but then it is already a single or a few line assignment, so I'm not sure delegating this simple assignment to a separate function.

For the RHS builder matrices it is almost entirely different for the first derivative, second derivative, interpolation, and staggared derivative. Only similarity is between the top and bottom BCs of the same operation if the BCs are the same. So there is a potential simplification here, but then the helper function needs to be a tad more elaborate to assign coefficients considering whether its a top or bottom, so not sure if it will be easier to understand at the end, haven't tried. Can you provide a very simple demonstration here if you have something in mind?

src/tdsops.f90
```diff
@@ -0,0 +1,682 @@
...
+      !! coeff arrays define the specific rules of building the RHS
+      !! corresponding to the tridiagonal system to be solved, and used only in
+      !! the first phase of the distributed algorithm when building the RHS.
+      !! If a boundary condition is defined then coeffs_t and coeffs_b differ
+      !! from coeffs array and define the RHS rule for the first and last 4
+      !! entries in the tridiagonal system (n_halo = 4).
+      !!
+      !! This class does not know about the current rank or its relative
+      !! location among other ranks. All the operator arrays here are used when
+      !! executing a distributed tridiagonal solver phase one or two.
```
mathrack, line 23: What is phase one or two ?

src/tdsops.f90
```diff
@@ -0,0 +1,682 @@
...
+      !! provided. dist_fw and dist_bw are used in the first phase of the
+      !! distributed tridiagonal solver algorithm. dist_sa and dist_sc are used
+      !! in the final substutition phase. See the kernels_dist.f90 files in the
+      !! relevant backend folders.
+      !! coeff arrays define the specific rules of building the RHS
+      !! corresponding to the tridiagonal system to be solved, and used only in
+      !! the first phase of the distributed algorithm when building the RHS.
+      !! If a boundary condition is defined then coeffs_t and coeffs_b differ
+      !! from coeffs array and define the RHS rule for the first and last 4
+      !! entries in the tridiagonal system (n_halo = 4).
```
mathrack, line 19: Could be relevant to include situations with different top / bottom level of halos. Also relevant could be situations with only one cell in a given direction. Current version of incompact3d had to deal with the case nz=1 to run 2D simulations.

src/tdsops.f90
```diff
@@ -0,0 +1,682 @@
...
+      !!
+      !! This class does not know about the current rank or its relative
+      !! location among other ranks. All the operator arrays here are used when
+      !! executing a distributed tridiagonal solver phase one or two.
+      real(dp), allocatable, dimension(:) :: dist_fw, dist_bw, dist_sa, dist_sc
+      real(dp), allocatable :: coeffs(:), coeffs_t(:, :), coeffs_b(:, :)
+      integer :: n, n_halo
+      real(dp) :: alpha, delta
+   contains
+      procedure :: deriv_1st, deriv_2nd, interpl_mid, stag_der_1st, preprocess
```
mathrack, line 29: Can we add a type-bound procedure `fin` to properly release the memory and restore the parameters to default values ?

src/tdsops.f90
```diff
@@ -0,0 +1,682 @@
...
+      !! If a boundary condition is defined then coeffs_t and coeffs_b differ
+      !! from coeffs array and define the RHS rule for the first and last 4
+      !! entries in the tridiagonal system (n_halo = 4).
+      !!
+      !! This class does not know about the current rank or its relative
+      !! location among other ranks. All the operator arrays here are used when
+      !! executing a distributed tridiagonal solver phase one or two.
+      real(dp), allocatable, dimension(:) :: dist_fw, dist_bw, dist_sa, dist_sc
+      real(dp), allocatable :: coeffs(:), coeffs_t(:, :), coeffs_b(:, :)
+      integer :: n, n_halo
```
mathrack, line 26: Do we need default values for `n`, `n_halo`, `alpha` and `delta` ?

src/tdsops.f90
```diff
@@ -0,0 +1,682 @@
...
+      real(dp) :: alpha, delta
```
mathrack, line 27: Is delta the size between nodes ? If yes, it could be removed and integrated in the coefficients ? What about cases with a stretched mesh ?

src/tdsops.f90
```diff
@@ -0,0 +1,682 @@
...
+   end interface tdsops_t
+
+contains
+
+   type(tdsops_t) function tdsops(n, delta, operation, variation, n_halo, &
+                                  from_to, bc_top, bc_bot, paire, c_nu, nu0_nu)
+      implicit none
+
+      integer, intent(in) :: n
+      integer, optional, intent(in) :: n_halo
```
mathrack: It is easier to read when the arguments are in the same order here and in the function description.

I think it is interesting to have optional arguments with a prefix opt_ in their name. Reading the code is easier then.

mathrack (changes requested):
I have performed a small review, I will look at it again later

src/derparams.f90
pbartholomew08: I had in mind ```call set_neuman(..., paire); call set_dirichlet(...)``` etc. The advantage of this is then you can have automated tests that call these directly with a preprepared input. Ultimately this should be easier to debug than calling the "driver" function and checking its output to find where it went wrong (still a worthwhile test, the finer grained ones will help narrow down where exactly the error is).

src/tdsops.f90
semi-h: I noticed in Incompact3D finalisation is a thing and allocatable arrays are freed before the executable terminates. I don't normally do this, and don't understand why it is important to do so. Can you explain a bit?

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+
+      tdsops%n = n
+      tdsops%delta = delta
+      if (present(n_halo)) then
+         tdsops%n_halo = n_halo
+      else
+         tdsops%n_halo = 4
+      end if
+
+      n_stencil = 2*tdsops%n_halo + 1
```
mathrack, line 77: I am not sure to understand this line. It leads to an oversized array coeffs in some cases. And the oversized array will lead to useless memory access ?

src/tdsops.f90
semi-h: putting n_halo a side (I'll start a different thread for that), we always want user to input an 'n', and should compain if they don't supply one. Its essential and I don't see any natural default number for 'n'. Alpha is actually an output, and it depends on the specifics of the operation user wants to execute. If the operation someone ones to execute is not yet available, the idea is that user adds something here in this class to make it available, and then calls the tdsops with correct arguments.
Again for delta, it is something we want user to input, I don't think there is a natural default for such a value and assuming something is just doesn't make sense to me.

src/tdsops.f90
semi-h: I think you're right, we don't need to access delta from the tdsops_t class, so it should be an input and not a member of the type itself. I'll remove if I can't find a use case for this.

src/tdsops.f90
semi-h: Fixed

src/tdsops.f90
semi-h: I feel it is more appropriate to explain this in the kernels_dist.f90 or maybe somewhere at top level documentation. Too irrelevant to talk about this in detail here in the tdsops class in my view. We still explain that there are two phases and the associated arrays for the phases. Maybe we can give a bit more detail though, if you have something in mind please share.

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+      if (operation == 'first-deriv') then
+         call tdsops%deriv_1st(variation, bc_top, bc_bot, paire)
+      else if (operation == 'second-deriv') then
+         call tdsops%deriv_2nd(variation, bc_top, bc_bot, paire, c_nu, nu0_nu)
+      else if (operation == 'interpolate') then
+         call tdsops%interpl_mid(variation, from_to, bc_top, bc_bot, paire)
+      else if (operation == 'staggared-deriv') then
+         call tdsops%stag_der_1st(variation, from_to, bc_top, bc_bot, paire)
+      else
+         error stop 'operation is not defined'
```
mathrack, line 94: I think it would be nice to have a proper warning / error catching mechanism. See the one in decomp2d maybe ? Similarly, a proper log mechanism is probably needed (output to file or stdout ? only master rank will output or all ranks will output ?)

I think it is good to plan this a bit with some verbose / debug flags or variables.

However, this is beyond the scope of the present PR.

src/tdsops.f90
pbartholomew08: Fortran should deallocate an (allocatable) array when it goes out of scope. An explicit deallocate statement can often be a "good practice" but shouldn't be necessary

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+         if (present(paire) .and. paire) then
+            ! paire == 1; d(uu)/dx, dv/dx, dw/dx
+            !             d(vv)/dy, du/dy, dw/dy
+            !             d(ww)/dz, du/dz, dv/dz
+            self%dist_sa(1) = 0._dp
+            self%dist_sc(1) = 0._dp
+            self%coeffs_t(1, :) = [0._dp, 0._dp, 0._dp, 0._dp, &
+                                   0._dp, &
+                                   0._dp, 0._dp, 0._dp, 0._dp]
+            self%coeffs_t(2, :) = [0._dp, 0._dp, 0._dp, -afi, &
```
mathrack, line 157: At this stage, are we sure `n_halo` is above 2 ?

src/tdsops.f90
semi-h: I'm tempted to opt for not bothering with deallocation to be honest. Simulation is going to terminate at the end anyways.

src/tdsops.f90
slaizet: It would be good to be able to run the code with nz=1. It is a nice feature to have. We do not need to deal with it now but we would want the framework to be compatible with this feature, if possible.

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+         ! second line
+         self%dist_sa(2) = 0.25_dp
+         self%dist_sc(2) = 0.25_dp
+         self%coeffs_t(2, :) = [0._dp, 0._dp, 0._dp, -0.75_dp/dx, &
+                                0._dp, &
+                                0.75_dp/dx, 0._dp, 0._dp, 0._dp]
+      end select
+
+      select case (bc_bot)
+      case ('neumann')
```
mathrack, line 189: I think the name is misleading. In case of symmetry, a variable can be odd or even. In the absence of symmetry, one can use Dirichlet / Neumann bounday conditions. However, homogeneous Neumann is a very specific Neumann condition. From a generic point of view, I think we have the options :

1. Periodic
2. Symmetry (odd / even)
3. Dirichlet / Neumann with imposed value or flux

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+         if (present(paire) .and. paire) then
+            ! paire == 1; d(uu)/dx, dv/dx, dw/dx
+            !             d(vv)/dy, du/dy, dw/dy
+            !             d(ww)/dz, du/dz, dv/dz
+            self%dist_sa(n) = 0._dp
+            self%dist_sc(n) = 0._dp
+            self%coeffs_b(4, :) = [0._dp, 0._dp, 0._dp, 0._dp, &
+                                   0._dp, &
+                                   0._dp, 0._dp, 0._dp, 0._dp]
+            self%coeffs_b(3, :) = [0._dp, 0._dp, -bfi, -afi, &
```
mathrack, line 199: See comment on the value of n_halo

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+                                   bfi, &
+                                   afi, 0._dp, 0._dp, 0._dp]
+         else
+            ! paire == 0; d(uv)/dx, d(uw)/dx, du/dx
+            !             d(vu)/dy, d(vw)/dy, dv/dy
+            !             d(wu)/dz, d(wv)/dz, dw/dz
+            self%dist_sa(n) = 2*self%alpha
+            self%dist_sc(n) = 0._dp
+            self%coeffs_b(4, :) = [0._dp, 0._dp, -2*bfi, -2*afi, &
```
mathrack, line 208: See comment on the value of n_halo

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+                                   0._dp, &
+                                   0._dp, 0._dp, 0._dp, 0._dp]
+            self%coeffs_b(3, :) = [0._dp, 0._dp, -bfi, -afi, &
```
mathrack, line 211: See comment on the value of n_halo

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+         end if
+      case ('dirichlet')
+         ! last line
+         self%dist_sa(n) = 2._dp
+         self%dist_sc(n) = 0._dp
+         self%coeffs_b(n_halo, :) = [0._dp, 0._dp, -0.5_dp/dx, -2._dp/dx, &
+                                     2.5_dp/dx, &
+                                     0._dp, 0._dp, 0._dp, 0._dp]
+         ! second last line
+         self%dist_sa(n-1) = 0.25_dp
```
mathrack, line 223: This incompatible with cases such as n=1

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+         self%dist_sc(n-1) = 0.25_dp
+         self%coeffs_b(n_halo-1, :) = [0._dp, 0._dp, 0._dp, -0.75_dp/dx, &
```
mathrack, line 225: See comment on the value of n_halo ?

mathrack: Similar comments apply below. I will skip them from hereon

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+                                   aici, &
+                                   aici + bici, bici + cici, cici + dici, dici]
+            self%coeffs_t(2, :) = [0._dp, 0._dp, 0._dp, bici, &
+                                   aici + cici, &
+                                   aici + dici, bici, cici, dici]
+            self%coeffs_t(3, :) = [0._dp, 0._dp, cici, bici + dici, &
+                                   aici, &
+                                   aici, bici, cici, dici]
+         case('p2v')
+            ! npaire is always 1
```
mathrack, line 508: I am not sure about npaire=1 if the user wants to interpolate the pressure on the velocity grid instead of computing the pressure gradient.

mathrack (changes requested):

src/tdsops.f90
mathrack: Could be relevant to give the location of the description of the phases then.

src/tdsops.f90
mathrack: Avoiding explicit deallocate statements makes the code slightly simpler. It also makes the hunt of a memory leak very difficult. If you think the code will never face a memory leak, then it is ok to avoid explicit deallocate statements.

src/tdsops.f90
semi-h: For the stretched mesh I think we need one more allocatable array like dist_st. We can add this when we need. There'll need to be a relevant change in one of the kernels_dist.f90 functions as well, I think only in the second phase function.

src/tdsops.f90
semi-h: We can maybe assign the z directional operator function pointers to a null function, then we won't need to bother with n=1 and compatibility of this. If we skip the instantiation of z directional operators and never use them it shouldn't be a problem.
But in general distributed algorithm requires at least 32 points so I wasn't paying attention to what happens when n is small

src/tdsops.f90
semi-h: Can you point me to this mechanism in decomp2d? A simple example if possible

mathrack: We use `decomp_2d_abort` and `decomp_2d_warning` to deal with errors and warnings.

The function is defined here : https://github.com/2decomp-fft/2decomp-fft/blob/dev/src/decomp_2d_mpi.f90#L133

It is used here for instance : https://github.com/2decomp-fft/2decomp-fft/blob/dev/src/io.f90#L128

For the log we have the following module. I think it can be improved, but it is not so bad : https://github.com/2decomp-fft/2decomp-fft/blob/dev/src/log.f90

src/tdsops.f90
mathrack: Maybe we could consider that the derivative (on a streched mesh) is another layer sitting on top of this one ?

mathrack: This could be an elegant way to deal with both stretched meshes and specific cases such as nz=1.

src/tdsops.f90
mathrack: We might want to skip this issue if the scope of the PR is limited to "mimick the current features of the current master branch of incompact3d" https://github.com/xcompact3d/x3d2/pull/11#discussion_r1368889905

However, I think the description of the boundary conditions above is sharper than the current one.

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+      !! If a boundary condition is defined then coeffs_t and coeffs_b differ
+      !! from coeffs array and define the RHS rule for the first and last 4
+      !! entries in the tridiagonal system (n_halo = 4).
+      !!
+      !! This class does not know about the current rank or its relative
+      !! location among other ranks. All the operator arrays here are used when
+      !! executing a distributed tridiagonal solver phase one or two.
+      real(dp), allocatable, dimension(:) :: dist_fw, dist_bw, dist_sa, dist_sc
+      real(dp), allocatable :: coeffs(:), coeffs_t(:, :), coeffs_b(:, :)
+      integer :: n, n_halo
```
semi-h, line 26: Okay lets discuss n_halo in this thread

For the first, second, and staggared derivative and interpolation with classic scheme, n_halo=2 can be enough. However if the hyperviscosity is in use or optimized/aggressiveinterpolation schemes are selected, n_halo is 4. Because of this I implemented the universal derivative function assuming that n_halo is 4. This also leads RHS builder stencil to have 9 entries in total and it is also hardcoded in the universal function. If one initiates tdsops_t class with a n_halo other than 4, the existing distributed solver algorithm can't be used with the coeff arrays that tdsops_t provides when n_halo/=4.

However, the tdsops_t class is an independent class that can potentially be used with a kernel implementation that only supports n_halo=2 (or 3, or anything else if needed). Because it doesn't really have an impact on performance we don't need to bother having another implementation of distributed algorithm and use this with the schemes that require only two halo cells. But there is no harm implementing tdsops_t class in a way that supports an arbitrary n_halo number. Basically, thats why I included n_halo parameter here. We'll never instantiate this class with n_halo other than 4 (and therefore the default value for n_halo is 4 and its an optional parameter, so we don't even need to specify anything or consider this). And for some schemes, the halo number cannot be smaller than 2 or 4, so assignments such as coeff(4, :) = [...] is completely fine in my opinion. Maybe we can add conditional checks and warn people if they request incompatible n_halo value with some operations/schemes.

semi-h:
variation is renamed as scheme in the last commit.

src/tdsops.f90
semi-h: If n_halo is 4 it means RHS at point i is a function of points [i-4, i-3, i-2, i-1, i, i+1, i+2, i+3, i+4]. If halo was 2 then it would be points from i-2 to i+2 so 2*2+1=5 in total.

src/tdsops.f90
semi-h: I think for performance it is better if we fuse the stretching operation with the second phase of the distributed algorithm. And if there is no stretching we will be multiplying the result array with an array of ones. Implementing this as a layer on top would require a dedicated function call that would add some extra cost due to moving data back and forth in memory

src/tdsops.f90
mathrack: I think it is misleading to have `n_halo` as a parameter / argument when under the hood, in this class and in some others, plenty of things are hard-coded for `n_halo=4`. At the very least, this should be clearly documented (in this class and in the universal derivative function) and the code should print some kind of warnings when n_halo is present and not 4.

mathrack: As a side remark, I am a bit surprised that the size of the stencil has no real impact on performance.

src/tdsops.f90
mathrack: No real impact on performance according to https://github.com/xcompact3d/x3d2/pull/11#discussion_r1369183495

src/tdsops.f90
```diff
@@ -0,0 +1,700 @@
...
+      real(dp) :: alpha, delta
+   contains
+      procedure :: deriv_1st, deriv_2nd, interpl_mid, stag_der_1st, preprocess
```
mathrack, line 29: I think we could also have a log subroutine. It could take a IO unit as input, and would print the coefficients / parameters in the object. From this perspective, it is nice to provide a name / label (inside the object or as a second argument for the log subroutine).

src/derparams.f90
semi-h: - variation is renamed as scheme
- bc_top/bot was indicating y direction indeed, so renamed as bc_start and bc_end, other places in the framework may refer to this in short form such as .._b/.._e.
- paire is renamed as sym.
And describtion for all this is provided in the class constructor function.

src/tdsops.f90
semi-h: We decided to have unit tests as a error catching mechanism for the moment.

src/tdsops.f90
semi-h: Irrelevant for the current PR as this is a too deep level to infer this info. The driver function that initiates the tdsops class with various settings can determine and output all this info.

PR #16, opened 2023-11-14
Add new subroutines to execute the generic and fused distributed solvers.

semi-h:


src/cuda/exec_dist.f90
```diff
@@ -0,0 +1,188 @@
...
+      real(dp), device, dimension(:, :, :), intent(out) :: du, dud, d2u
+      real(dp), device, dimension(:, :, :), intent(out) :: &
+         du_send_s, du_send_e, du_recv_s, du_recv_e, &
+         dud_send_s, dud_send_e, dud_recv_s, dud_recv_e, &
+         d2u_send_s, d2u_send_e, d2u_recv_s, d2u_recv_e
+
+      type(cuda_tdsops_t), intent(in) :: der1st, der2nd
+      real(dp), intent(in) :: nu
+      integer, intent(in) :: nproc, pprev, pnext
+      type(dim3), intent(in) :: blocks, threads
```
semi-h, line 103: What do you think about the argument ordering here?

semi-h: Some may argue that those temporary storages are relevant only in this subroutine, so maybe should be allocated in there, but we know that allocating temporary arrays each time a function is called is bad for performance.

Alternative would be wrapping this inside a class so that temporary data storages are a member of the class. But then we allocate temporaries for all the instances we create from this class, and this also results in unnecessary memory usage.

semi-h: So my point is grouping `r_u` and `du, dud, d2, ...` together and listing them next to each other just because they happen to have the same attribute `intent(out)` is wrong, because the semantic here is different for these two groups, and we need to figure out a way to communicate this for making it easy to understand.
So maybe the guideline for us could be
```
intent(out) [the ones we actually care about]
intent(in) [the corresponding fields we pass to get the result, and their auxilary components]
facilitators? [temporary storages that happen to have intent(out) attribute to please the compiler]
The rest of the inputs, scalars, coefficient arrays that we need to carry out the operation etc.
```

Any opinions? Would you prefer these `facilitators` between the actual `intent(out)`s and the corresponding `intent(in)`s? Is it a bad idea to seperate some of the `intent(in)`s from the rest of them like the scalar ones? What is the most natural ordering would be in this subroutine in your opinion?

slaizet: let's discuss about this on Monday :-)

slaizet: great, thanks

PR #19, opened 2023-12-01
Add tridiagonal solver to omp backend

### LINKED ISSUE #18: Add tridiagonal solver to omp backend
Use `test_omp_tridiag` as basis for tridiagonal solver

Nanoseb:
closes #18

### DIFF
--- a/tests/omp/test_omp_tridiag.f90
+++ b/tests/omp/test_omp_tridiag.f90
@@ -219,9 +219,9 @@ program test_omp_tridiag
    if (nrank == 0) then
       if (norm_du > tol) then
          allpass = .false.
-        write (stderr, '(a)') 'Check 2nd ders, hyperviscous, dir-neu... failed'
+         write (stderr, '(a)') 'Check 2nd ders, hyperviscous, dir-neu... failed'
       else
-        write (stderr, '(a)') 'Check 2nd ders, hyperviscous, dir-neu... passed'
+         write (stderr, '(a)') 'Check 2nd ders, hyperviscous, dir-neu... passed'
       end if
    end if
 
--- a/tests/omp/test_omp_tridiag.f90
+++ b/tests/omp/test_omp_tridiag.f90
@@ -28,14 +28,11 @@ program test_omp_tridiag
    type(tdsops_t) :: tdsops
 
    character(len=20) :: bc_start, bc_end
-   real(dp), allocatable, dimension(:, :) :: coeffs_s, coeffs_e
-   real(dp), allocatable, dimension(:) :: coeffs, dist_fw, dist_bw, dist_af, &
-                                          dist_sa, dist_sc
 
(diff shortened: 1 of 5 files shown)


tests/omp/test_omp_tridiag.f90
```diff
@@ -216,21 +214,21 @@ program test_omp_tridiag
...
 
    call check_error_norm(du, sin_0_2pi, n, n_glob, n_block, 1, norm_du)
-   if (nrank == 0) print*, 'error norm hyperviscous', norm_du
+   if (nrank == 0) print *, 'error norm hyperviscous', norm_du
 
    if (nrank == 0) then
       if (norm_du > tol) then
          allpass = .false.
-         write(stderr, '(a)') 'Check 2nd ders, hyperviscous, dir-neu... failed'
+        write (stderr, '(a)') 'Check 2nd ders, hyperviscous, dir-neu... failed'
```
semi-h, line 222: seems like fprettify failed here for some reason, indentation level is wrong

tests/omp/test_omp_tridiag.f90
```diff
@@ -246,13 +244,13 @@ program test_omp_tridiag
 
    if (nrank == 0) then
       print'(a, f8.3, a)', 'Available BW:   ', deviceBW/2**30, &
-                           ' GiB/s (per NUMA zone on ARCHER2)'
+         ' GiB/s (per NUMA zone on ARCHER2)'
```
semi-h, line 247: Is this also fprettify's auto fix?

tests/omp/test_omp_tridiag.f90
Nanoseb: indeed, gonna fix that

tests/omp/test_omp_tridiag.f90
Nanoseb: yes, it is also fprettify here. Though I don't mind that one too much

semi-h: It looks okay. But if you're bothered with the whitespace between `print` and `*` you can add
`whitespace-print = false`
to the `.fprettify.ini` file if you want.  But then I think you can't put a space even after comma, so it becomes `print*,'something'` if I remember right. There is really not enough control over minor fprettify settings unfortunately.

Nanoseb: I actually prefer `print *` over `print*`, so I don't mind the current syntax. I think I'd rather have `print '(xyz)'` versus `print'(xyz)'`, but I can live with that. As long as it is somewhat consistent and I don't have to manually tweak it I am fine with most syntax rule.

tests/omp/test_omp_tridiag.f90
Nanoseb: Fixed

PR #26, opened 2023-12-20
Add divergence and gradient

semi-h:
This PR demonstrates the current capability of the framework pretty well. Both the divergence and gradient operations in Xcompact3D algorithm is now supported in the new framework and this feature is added only by adding two corresponding functions in the solver class. There was no need to add anything special in any of the backends to add this support as they already provide a method that solves a single tridiagonal system.

### DIFF
--- a/src/backend.f90
+++ b/src/backend.f90
@@ -34,6 +34,7 @@ module m_base_backend
       procedure(trans_d2d), deferred :: trans_z2y
       procedure(trans_d2d), deferred :: trans_y2x
       procedure(sum9into3), deferred :: sum_yzintox
+      procedure(vecadd), deferred :: vecadd
       procedure(get_fields), deferred :: get_fields
       procedure(set_fields), deferred :: set_fields
       procedure(alloc_tdsops), deferred :: alloc_tdsops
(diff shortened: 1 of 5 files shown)


src/backend.f90
```diff
@@ -147,6 +182,10 @@ subroutine alloc_tdsops(self, tdsops, n, dx, operation, scheme)
          integer, intent(in) :: n
          real(dp), intent(in) :: dx
          character(*), intent(in) :: operation, scheme
+         integer, optional, intent(in) :: n_halo
+         character(*), optional, intent(in) :: from_to, bc_start, bc_end
+         logical, optional, intent(in) :: sym
+         real(dp), optional, intent(in) :: c_nu, nu0_nu
```
Nanoseb, line 188: Could you explain what these new arguments are for?
Also, `alloc_tdsops` hasn't been updated to use these arguments, maybe mention it is a WIP?

Nanoseb (commented):
Beside that it looks good to me

src/backend.f90
semi-h: Thanks for pointing out, alloc_tdsops methods in backends were not passing all these optional arguments down to tdsops_init but now it is all fixed.

PR #27, opened 2024-01-19
Implement transeq in omp backend

### LINKED ISSUE #21: Implement transeq in omp backend

Nanoseb:
closes #21


src/omp/backend.f90
```diff
@@ -147,10 +150,36 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+      call tds_solve_omp(self, duu, uu, dirps, dirps%der1st_sym)
+
+      d2u => self%allocator%get_block()
+      call tds_solve_omp(self, d2u, u, dirps, dirps%der2nd)
+
+
+
+
+   end subroutine transeq_omp_dist
+
```
semi-h, line 182: This is not the way we want to implement transeq stuff on the OpenMP backend. We can't take advantage of the cache and reduce the data movements if we implement in this way.

We should instead implement a subroutine that is similar to exec_dist_tds_compact.
https://github.com/xcompact3d/x3d2/blob/main/src/omp/exec_dist.f90#L14C15-L14C36

And the idea is we call the der_univ_dist to get du/dx, duu/dx, and d2u/dx2 one after another as we loop through batches of lines we have in the domain. (So a loop over the 3rd dimension of the arrays we have, aka the group number.

It should look like the loop below, but with 3 distinct calls to der_univ_dist inside the loop with correct set of tdsops instances.
https://github.com/xcompact3d/x3d2/blob/main/src/omp/exec_dist.f90#L40-L47

This will also make an operation like vecmul_omp unnecessary, saving some data movement.

Nanoseb: ah yes, makes sense indeed. I will do that

Nanoseb:
Just reimplemented following your comment. Note that, for now, I have kept the `velmul_omp` because it will require a rewrite of `der_univ_dist` to do it inplace. I will likely do that in an other branch because I may try to simplify `der_univ_dist` itself at the same time.

Still not ready to be merged, I need to cleanup the naming convention I used, test it and add a test.

src/omp/backend.f90
```diff
@@ -147,10 +150,76 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+      call transeq_dist_component(self, du, u, u, dirps%der1st, dirps%der1st_sym, dirps%der2nd, dirps)
+
+      call transeq_dist_component(self, dv, v, u, dirps%der1st_sym, dirps%der1st, dirps%der2nd_sym, dirps)
+
+      call transeq_dist_component(self, dw, w, u, dirps%der1st_sym, dirps%der1st, dirps%der2nd_sym, dirps)
+
+   end subroutine transeq_omp_dist
+
+   subroutine transeq_dist_component(self, rhs, v, u, tdsops_du, tdsops_dud, tdsops_d2u, dirps)
```
semi-h, line 173: missing implicit none

src/omp/backend.f90
```diff
@@ -147,10 +150,76 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+
+      call copy_into_buffers(self%u_send_s, self%u_send_e, u%data, dirps%n)
+      call copy_into_buffers(self%v_send_s, self%v_send_e, uv%data, dirps%n)
+
+      ! halo exchange
+      call sendrecv_fields(self%u_recv_s, self%u_recv_e, self%u_send_s, self%u_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
+
+      call sendrecv_fields(self%v_recv_s, self%v_recv_e, self%v_send_s, self%v_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
```
semi-h, line 203: I think we can move the halo exchange above this subroutine. After all we need halos for all u, v, w fields, and can make all of them ready before calling any of the transeq_dist_components. Currently there is 6 times halo exchange, but when the communication is handled at transeq_omp_dist level we can get away with only 3 times.

src/omp/backend.f90
```diff
@@ -147,10 +150,76 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
 
+   subroutine transeq_omp_dist(self, du, dv, dw, u, v, w, dirps)
+      implicit none
+
+      class(omp_backend_t) :: self
+      class(field_t), intent(inout) :: du, dv, dw
+      class(field_t), intent(in) :: u, v, w
+      type(dirps_t), intent(in) :: dirps
+
+      call transeq_dist_component(self, du, u, u, dirps%der1st, dirps%der1st_sym, dirps%der2nd, dirps)
```
semi-h, line 165: I like this approach, I'll implement this in the CUDA backend as well.

src/omp/backend.f90
```diff
@@ -147,10 +150,76 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+
+      integer :: n_halo
+
+      ! TODO: don't hardcode n_halo
+      n_halo = 4
+
+      du => self%allocator%get_block()
+      dud => self%allocator%get_block()
+      uv => self%allocator%get_block()
+      call vecmul_omp(uv, u, v, dirps)
```
semi-h, line 192: We don't need this at all, see my comment in exec_dist_transeq_compact.

src/omp/exec_dist.f90
```diff
@@ -61,5 +61,107 @@ subroutine exec_dist_tds_compact( &
...
+
+         call der_univ_dist( &
+            d2u(:, :, k), d2u_send_s(:, :, k), d2u_send_e(:, :, k), u(:, :, k), &
+            u_recv_s(:, :, k), u_recv_e(:, :, k), &
+            tdsops_d2u%coeffs_s, tdsops_d2u%coeffs_e, tdsops_d2u%coeffs, tdsops_d2u%n, &
+            tdsops_d2u%dist_fw, tdsops_d2u%dist_bw, tdsops_d2u%dist_af &
+            )
+
+         call der_univ_dist( &
+            dud(:, :, k), dud_send_s(:, :, k), dud_send_e(:, :, k), ud(:, :, k), &
```
semi-h, line 119: Because we pass a very small section of the domain into the der_univ_dist, the multiplication between convective velocity and primary velocity fields can be carried out just here only for the bit that we're passing. Just to give an idea, its only around 64 KiB when n=1024 and SZ=8.
So just allocate a 2D array with size (SZ, n), then just before calling der_univ_dist third time do
`temp = convective*u`, where convective is either `u`, `v` or `w`. Then the `temp` can be passed to der_univ_dist instead of `ud`. It might worth to do the multiplication with an explicit loop and simd directives.

semi-h (changes requested):

src/omp/exec_dist.f90
Nanoseb: ah yes, ok. I always assume we are playing with large arrays. That will simplify things

src/omp/backend.f90
Nanoseb: actually, no because there is one at the top of the file (I just double checked). I think we can get rid of them in each routine. Not that it matters too much either though.

src/omp/backend.f90
Nanoseb: Yes, that was my plan once I could handle `uv` things. But thanks to your other comment that will be easier than expected. I will do that

semi-h (commented):
Looks great! I think now we get all the benefit from CPU cache.

src/omp/backend.f90
```diff
@@ -147,10 +150,83 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+      call copy_into_buffers(self%w_send_s, self%w_send_e, w%data, dirps%n)
+
+      call sendrecv_fields(self%u_recv_s, self%u_recv_e, self%u_send_s, self%u_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
+      call sendrecv_fields(self%v_recv_s, self%v_recv_e, self%v_send_s, self%v_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
+      call sendrecv_fields(self%w_recv_s, self%w_recv_e, self%w_send_s, self%w_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
+
+   end subroutine
```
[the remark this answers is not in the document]
Nanoseb: yes, it is never a requirement (maybe it used to be in older standards?). It is weird indeed that github syntax highlight doesn't handle it. I can add it if needed.

src/omp/backend.f90
```diff
@@ -147,10 +150,83 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+      class(field_t), intent(inout) :: du, dv, dw
+      class(field_t), intent(in) :: u, v, w
+      type(dirps_t), intent(in) :: dirps
+      integer :: n_halo
+
+      call transeq_halo_exchange(self, u, v, w, dirps)
+
+      call transeq_dist_component(self, du, u, u, dirps%der1st, dirps%der1st_sym, dirps%der2nd, dirps)
+      call transeq_dist_component(self, dv, v, u, dirps%der1st_sym, dirps%der1st, dirps%der2nd_sym, dirps)
+      call transeq_dist_component(self, dw, w, u, dirps%der1st_sym, dirps%der1st, dirps%der2nd_sym, dirps)
```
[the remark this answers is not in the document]
Nanoseb: ah yeah. Indeed I can change that. Though, is there a reason we have to comply with the 79 columns limit? 
If it is mainly for aesthetics reasons, I'd argue that keeping single lines is a lot clearer than breaking it here for example.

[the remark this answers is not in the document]
semi-h: There is not really a reason we just put it in our in our style guide when we're trying to come up with something and fprettify was set to check this as well. I actually favour it because I normally write code in a default size terminal screen and going beyond 79 columns the line gets wrapped and to avoid this I don't go beyond that. Even if one doesn't write code on a terminal a quick edit when running the code on a cluster is what most people do I think and then wrapped lines can make things a little confusing. Also, when looking at the changes on github with the split view too long lines get wrapped which is again not ideal I believe. And these three lines in particular are so long that a horizontal scroll bar appears for me on github in the conversation tab here and even if my window is wide and can't see the whole line without scrolling.

semi-h:
I would only test `exec_dist_transeq_compact` for now. It would be much simpler and targeted. transeq only calls `exec_dist_transeq_compact` 3x3 times after all and testing it requres instantiating the allocator and backend and then setting a few variables. I think its beyond the scope of the current test which should be more low level.
Also, its very likely that allocator will change a bit and then we would need to edit this test program too. Probably there'll be changes in how we set the variables in the backend as well, and mirroring all these here in the transeq test shouldn't be necessary. 
I think any test initiating the backend should be one level up so that we can test all backends picking the right one at runtime and the main procedures in backends can be tested in a better way and in a single program.

Nanoseb:
> I would only test `exec_dist_transeq_compact` for now.

Added a test for it too. Now we are testing both.

src/omp/backend.f90
```diff
@@ -144,10 +147,86 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+      call copy_into_buffers(self%w_send_s, self%w_send_e, w%data, dirps%n)
+
+      call sendrecv_fields(self%u_recv_s, self%u_recv_e, self%u_send_s, self%u_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
+      call sendrecv_fields(self%v_recv_s, self%v_recv_e, self%v_send_s, self%v_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
+      call sendrecv_fields(self%w_recv_s, self%w_recv_e, self%w_send_s, self%w_send_e, &
+                           SZ*n_halo*dirps%n_blocks, dirps%nproc, dirps%pprev, dirps%pnext)
+
+   end subroutine
```
JamieJQuinn, line 195:
```suggestion
   end subroutine transeq_halo_exchange
```

We should keep in line with the rest of the code.

src/omp/backend.f90
```diff
@@ -144,10 +147,86 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+         u%data, self%u_recv_s, self%u_recv_e, &
+         v%data, self%v_recv_s, self%v_recv_e, &
+         tdsops_du, tdsops_dud, tdsops_d2u, self%nu, &
+         dirps%nproc, dirps%pprev, dirps%pnext, dirps%n_blocks)
+
+      call self%allocator%release_block(du)
+      call self%allocator%release_block(dud)
+      call self%allocator%release_block(d2u)
+
+   end subroutine
```
JamieJQuinn, line 228:
```suggestion
   end subroutine transeq_dist_component
```

src/omp/backend.f90
```diff
@@ -157,10 +236,36 @@ subroutine tds_solve_omp(self, du, u, dirps, tdsops)
...
+      implicit none
+
+      class(omp_backend_t) :: self
+      class(field_t), intent(inout) :: du
+      class(field_t), intent(in) :: u
+      type(dirps_t), intent(in) :: dirps
+      class(tdsops_t), intent(in) :: tdsops
+      integer :: n_halo
+
+      ! TODO: don't hardcode n_halo
```
JamieJQuinn, line 253: Can an issue be raised about this so the TODO isn't just lost in the code?

src/omp/exec_dist.f90
```diff
@@ -61,5 +61,134 @@ subroutine exec_dist_tds_compact( &
...
+
+      real(dp), dimension(:, :), allocatable :: ud, ud_recv_s, ud_recv_e
+      real(dp) :: nu
+      integer, intent(in) :: nproc, pprev, pnext
+      integer, intent(in) :: n_block
+
+      integer :: n_data, n_halo
+      integer :: k, i, j, n
+
+      ! TODO: don't hardcode n_halo
```
JamieJQuinn, line 102: Ditto ^

JamieJQuinn (changes requested):
Looks great! Just a few small suggestions.

src/omp/backend.f90
```diff
@@ -144,10 +147,86 @@ subroutine transeq_z_omp(self, du, dv, dw, u, v, w, dirps)
...
+   subroutine transeq_halo_exchange(self, u, v, w, dirps)
+      class(omp_backend_t) :: self
+      class(field_t), intent(in) :: u, v, w
+      type(dirps_t), intent(in) :: dirps
+      integer :: n_halo
+
+      ! TODO: don't hardcode n_halo
+      n_halo = 4
+
+      call copy_into_buffers(self%u_send_s, self%u_send_e, u%data, dirps%n)
```
semi-h, line 184: Can you please add !$omp parallel and simd stuff in `copy_into_buffers` subroutine? It should help a bit with performance.

Nanoseb: Just did it

semi-h:
Distributed solver requires up to 128/256 points per rank based on the particular compact scheme we solve. If you want to do parallel tests the test can fail if you have too few points in a rank.

We don't have a parallel testing enviroment yet but can you confirm that the test passes when you run the executable by hand with multiple ranks? Because you use the default schemes I think 64 points per rank should be more than enough.

Nanoseb:
> Distributed solver requires up to 128/256 points per rank based on the particular compact scheme we solve. If you want to do parallel tests the test can fail if you have too few points in a rank.


Yes, exactly. With 64 cells, it was failing with 4 cores (error ~ 7e-8) and working on 2. That's why I have increased it to 96 now so it works up to 4 cores with the tolerance that was set (1e-8).

Nanoseb:
I lowered the cell count just to run them faster. Went from 2s to 0.4 I think. For now it isn't a big deal, but when we will have 10 or 20 tests, that quickly adds up. Being able to run many tests quickly makes you run them more often.

src/omp/exec_dist.f90
```diff
@@ -61,5 +61,134 @@ subroutine exec_dist_tds_compact( &
...
+                            tdsops_dud%n, tdsops_dud%dist_sa, tdsops_dud%dist_sc)
+
+         call der_univ_subs(d2u(:, :, k), &
+                            d2u_recv_s(:, :, k), d2u_recv_e(:, :, k), &
+                            tdsops_d2u%n, tdsops_d2u%dist_sa, tdsops_d2u%dist_sc)
+
+         do j = 1, n
+            !$omp simd
+            do i = 1, SZ
+               rhs(i, j, k) = -0.5*(v(i, j, k)*du(i, j, k) + dud(i, j, k)) + nu*d2u(i, j, k)
```
semi-h, line 180: `0.5` here is not with `_dp`, I'm not sure what exactly this would result in but I always suffix floating data with precision just to be on the safe side.

Nanoseb: fair point. Just did it

src/omp/exec_dist.f90
```diff
@@ -61,5 +61,134 @@ subroutine exec_dist_tds_compact( &
...
+      call sendrecv_fields(du_recv_s, du_recv_e, du_send_s, du_send_e, &
+                           n_data, nproc, pprev, pnext)
+      call sendrecv_fields(dud_recv_s, dud_recv_e, dud_send_s, dud_send_e, &
+                           n_data, nproc, pprev, pnext)
+      call sendrecv_fields(d2u_recv_s, d2u_recv_e, d2u_send_s, d2u_send_e, &
+                           n_data, nproc, pprev, pnext)
+
+      !$omp parallel do
+      do k = 1, n_block
+         call der_univ_subs(du(:, :, k), &
```
semi-h, line 165: I realised that here we're writing 3 field sized arrays into main memory unnecessarily. It is potentially increasing the runtime %20.

In the second phase of the algorithm here we pass a part of the `du`, `dud`, and `d2u` into der_univ_subs, and they're all rewritten in place. Then later we combine them in `rhs` for the final result. Ideally, we want `du`, `dud`, and `d2u` to be read once and `rhs` to be written only once. However because of the way der_univ_subs work, the updated data in `du` arrays after der_univ_subs call gets written in the main memory, even though we don't need this at all.

There are three ways we can fix this
- In the parallel do loop in the second phase we can copy the relevant parts of `du`, `dud`, and `d2u` arrays into `(SZ, n)` sized temporary arrays. Then we pass temporary arrays into der_univ_subs, and at the end we use these temporaries to obtain final `rhs`. This is the easiest solution but it may not be the best in terms of performance.
- We can write an alternative der_univ_subs and separate input and output arrays. This way we can pass a part of the `du` arrays as we do now, and pass a small temporary array as the output one. Because `du` arrays will be input arrays no data will be written in main memory. Then we can combine the temporaries to get `rhs`.
- If we're writing an alternative der_univ_subs to be used in transeq, we can go one step further and have a fused version of it. This would probably the most performant solution. der_univ_subs is relatively lightweight so it isn't really hard to do so. The new subrotuine can input all `du`, `dud`, and `d2u`, and write the final result `rhs`.

src/omp/exec_dist.f90
```diff
@@ -61,5 +61,134 @@ subroutine exec_dist_tds_compact( &
...
+      ! TODO: don't hardcode n_halo
+      n_halo = 4
+      n = tdsops_d2u%n
+      n_data = SZ*n_block
+
+      allocate(ud(SZ, n))
+      allocate(ud_recv_e(SZ, n_halo))
+      allocate(ud_recv_s(SZ, n_halo))
+
+      !$omp parallel do
```
semi-h, line 111: We make use of `ud`, `ud_recv_e` and `ud_recv_s` arrays defined and allocated above here inside the parallel do loop. We need a minor fix here to ensure we get correct result and not result in race conditions. I suspect that the current state can produce wrong result in some cases, and due to race conditions it is probably not as fast as it can be.

When you enter a OpenMP parallel region you instantly have OMP_NUM_THREADS many processes working together. And the variables and arrays you defined outside a parallel region can be accessed by any of the threads. For example, we only have one `du` field, and all threads in this parallel region can read/write this array. But because pass a unique part of these arrays for each process (via the loop over k), there is no problem with them. But for example, if you write `du(1,1,1)  = 0`, all of the threads will do the exact same operation for you, its a very important thing to consider when using OpenMP.

The problem here as you can guess now is with the temporary arrays, (`ud` ..). Unless we specify, all threads will try to make use of the very same memory when we try to write or read from `ud`. We don't want this because these threads are working on different parts of the domain and need their own temporary storages. It is easy to fix this, just replace 
`!$omp parallel do` with
`!$omp parallel do private(ud, ud_recv_e, ud_recv_s)`
so that all threads have their own copy of these temporary arrays.

I guess the tests pass because we have the exact sin wave in all the lines in the domain, so the temporaries are exactly the same accross domain. If the initial field is not the same accross the domain I don't think the test can pass.

example: https://github.com/xcompact3d/transeq_perf/blob/9849028f940200800b449f80d394fde67729a585/vectranseq.f90#L215

semi-h (changes requested):
Making the temporaries private inside the parallel loops is very important, and once it is fixed I'm happy to merge. We can work on the performance related part in a new PR.

src/omp/exec_dist.f90
Nanoseb: ah yes of course, well spotted. My openMP is a bit rusty and I lost the reflex to check these things. I will correct that

src/omp/exec_dist.f90
semi-h: #40 is relevant here

Nanoseb: Good points, I will have a think about it, but indeed having it in a new PR focusing on performance makes sense.

semi-h: Alright, I think best way to move forward is implementing the first strategy and checking how much of the peak BW we get. If its a good utilisation then maybe we don't need to bother with a new der_univ_subs at all.

Issue #24, opened 2023-12-14
Field size in x/y/z orientations


semi-h:
We minimize the memory requirement via our memory pool strategy.  We have implemented the allocator module to provide this functionality. Allocator module is responsible from allocating memory on the device or host, and then passing a pointer to a memory region when a method requires a temporary field storage.

Currently, allocator module allocates a 3D array with specified sizes.
`allocate (field(SZ, nx, ny*nz/SZ)`
However, the number of vertices can differ in x, y, and z directions. Therefore, when switching from one orientation to another, or when requesting a pointer from the allocator module, we may need to specify a size so that we have a temporary storage with correct shape. A feature introduced in Fortran 2003 called bounds remapping allows assigning a pointer with a different size or dimension to an existing array and I have tested this feature both with the nvidia and gcc compilers so don't expect any issue here.

To sum up, we need to improve the allocator module a little so that we can specify the shape of the array we need. I think it will be a simple fix to the `get_block` method in the allocator module. This may require us to convert `get_block` from a function to a subroutine. Until this is fixed we can only run simulations where (nx_local == ny_local == nz_local).

I think it also makes sense to allocate a 1D array with right size (nx*ny*nz) when `create_block` is executed because then Fortran compilers should be more comfortable assuming that our pointers point to a continuous piece of memory without any strides.

PR #29, opened 2024-01-24
Add transposer kernels for reordering field data.

semi-h:
We need to reorder field data as we switch between x, y, and z orientations.

First commit adds transposing operation from x to y.

Select type stuff we have to do here is a bit annoying, and it is mainly due to the extention of the base field type in CUDA backend to support device arrays in Fortran. Any suggestion here is welcome @JamieJQuinn.

### DIFF
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -3,7 +3,7 @@ module m_cuda_backend
 
    use m_allocator, only: allocator_t, field_t
    use m_base_backend, only: base_backend_t
-   use m_common, only: dp, globs_t
+   use m_common, only: dp, globs_t, TRP_X2Y, TRP_X2Z, TRP_Y2X, TRP_Y2Z, TRP_Z2Y
    use m_tdsops, only: dirps_t, tdsops_t
 
    use m_cuda_allocator, only: cuda_allocator_t, cuda_field_t
@@ -432,24 +432,24 @@ module m_cuda_backend
       select type(u_i); type is (cuda_field_t); u_i_d => u_i%data_d; end select
 
       select case (direction)
-      case (12) ! x2y
+      case (TRP_X2Y) ! x2y
          blocks = dim3(self%nx_loc/SZ, self%nz_loc, self%ny_loc/SZ)
          threads = dim3(SZ, SZ, 1)
          call trans_x2y_k<<<blocks, threads>>>(u_o_d, u_i_d, self%nz_loc)
-      case (13) ! x2z
+      case (TRP_X2Z) ! x2z
          blocks = dim3(self%nx_loc, self%ny_loc/SZ, 1)
          threads = dim3(SZ, 1, 1)
          call trans_x2z_k<<<blocks, threads>>>(u_o_d, u_i_d, self%nz_loc)
-      case (21) ! y2x
+      case (TRP_Y2X) ! y2x
          blocks = dim3(self%nx_loc/SZ, self%ny_loc/SZ, self%nz_loc)
          threads = dim3(SZ, SZ, 1)
          call trans_y2x_k<<<blocks, threads>>>(u_o_d, u_i_d, self%nz_loc)
-      case (23) ! y2z
+      case (TRP_Y2Z) ! y2z
          blocks = dim3(self%nx_loc/SZ, self%ny_loc/SZ, self%nz_loc)
          threads = dim3(SZ, SZ, 1)
          call trans_y2z_k<<<blocks, threads>>>(u_o_d, u_i_d, &
                                                self%nx_loc, self%nz_loc)
-      case (32) ! z2y
+      case (TRP_Z2Y) ! z2y
          blocks = dim3(self%nx_loc/SZ, self%ny_loc/SZ, self%nz_loc)
          threads = dim3(SZ, SZ, 1)
 
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -415,21 +415,21 @@ module m_cuda_backend
 
    end subroutine tds_solve_dist
 
-   subroutine trans_x2y_cuda(self, u_y, v_y, w_y, u, v, w)
+   subroutine trans_x2y_cuda(self, u_y, u)
       implicit none
 
       class(cuda_backend_t) :: self
-      class(field_t), intent(inout) :: u_y, v_y, w_y
-      class(field_t), intent(in) :: u, v, w
+      class(field_t), intent(inout) :: u_y
+      class(field_t), intent(in) :: u
 
    end subroutine trans_x2y_cuda
 
-   subroutine trans_x2z_cuda(self, u_z, v_z, w_z, u, v, w)
+   subroutine trans_x2z_cuda(self, u_z, u)
       implicit none
 
       class(cuda_backend_t) :: self
-      class(field_t), intent(inout) :: u_z, v_z, w_z
-      class(field_t), intent(in) :: u, v, w
+      class(field_t), intent(inout) :: u_z
+      class(field_t), intent(in) :: u
 
    end subroutine trans_x2z_cuda
 
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -1,4 +1,5 @@
 module m_cuda_backend
+   use iso_fortran_env, only: stderr => error_unit
    use cudafor
 
    use m_allocator, only: allocator_t, field_t
@@ -452,12 +453,10 @@ module m_cuda_backend
       case (RDR_Z2Y) ! z2y
          blocks = dim3(self%nx_loc/SZ, self%ny_loc/SZ, self%nz_loc)
          threads = dim3(SZ, SZ, 1)
-
          call reorder_z2y<<<blocks, threads>>>(u_o_d, u_i_d, &
                                                self%nx_loc, self%nz_loc)
       case default
-         print *, 'Transpose direction is undefined.'
-         stop
+         error stop 'Reorder direction is undefined.'
       end select
 
    end subroutine reorder_cuda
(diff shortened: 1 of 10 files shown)


src/cuda/backend.f90
```diff
@@ -415,21 +421,43 @@ subroutine tds_solve_dist(self, du, u, dirps, tdsops, blocks, threads)
 
    end subroutine tds_solve_dist
 
-   subroutine trans_x2y_cuda(self, u_y, v_y, w_y, u, v, w)
+   subroutine trans_x2y_cuda(self, u_y, u)
```
Nanoseb, line 424: Is code duplication needed here for performance reasons? Because I think it would be clearer, easier to maintain and less error prone if it was a single subroutine with something like:
```Fortran
trans_a2b_cuda(dir_a, dir_b, u_b, u)
integer, intent(in) :: dir_a
integer, intent(in) :: dir_b
...
select case (dir_a)
   case (dir_x)
     blocks = ...

end select
 ...
```

With `dir_x`, `dir_y` and `dir_z` parameters defined somewhere. That's just one approach, but we can have something else, like providing the content of blocks or threads as argument or in a separate function. I just feel like there is very few changes between the different functions and that doesn't really require duplication.

src/cuda/kernels_trans.f90
```diff
@@ -0,0 +1,119 @@
+module m_cuda_kernels_trans
+   use cudafor
+
+   use m_common, only: dp
+   use m_cuda_common, only: SZ
+
+contains
+
+   attributes(global) subroutine trans_x2y_k(u_y, u_x, nz)
```
Nanoseb, line 9: Similarly, I feel like these can be refactored into a single subroutine

Nanoseb (commented):
Sorry I can't really comment on the core of the code, but just had a couple of remarks about code duplication. Feel free to ignore them though. 

One last thing, that's the kind of feature for which an unit test make a lot of sense and will be fairly easy. Just have some data, transpose it several times and back to its original position and test if it is still the same as before.

src/cuda/backend.f90
semi-h: At this level we can figure out a better way to handle transpose operation and it shouldn't have any effect on performance.
What I have in mind is having something like `transpose(u_, u, TR_XTOY)`, where we define all possible transposition directions as integer parameters somewhere. Then the backends can check this parameter and call the right function.

src/cuda/kernels_trans.f90
semi-h: I'm not sure if we can merge all these kernels into one while making sure the performance is good. This is specific to CUDA backend tough, on the OpenMP backend I think it is possible. What you have in CUDA kernels is like the bit of code you have inside some number of nested loops, and any conditional checks over there can make performance bad.

Nanoseb: yes ok, that's the part I wasn't sure about.

src/cuda/backend.f90
```diff
@@ -415,21 +421,43 @@ subroutine tds_solve_dist(self, du, u, dirps, tdsops, blocks, threads)
...
       class(cuda_backend_t) :: self
-      class(field_t), intent(inout) :: u_y, v_y, w_y
-      class(field_t), intent(in) :: u, v, w
+      class(field_t), intent(inout) :: u_y
+      class(field_t), intent(in) :: u
+
+      real(dp), device, pointer, dimension(:, :, :) :: u_d, u_y_d
+      type(dim3) :: blocks, threads
+
+      select type(u); type is (cuda_field_t); u_d => u%data_d; end select
```
pbartholomew08, line 434: WRT comment that "select type stuff is annoying" - I agree. One possible way to make it slightly better would be
```
subroutine set_data_pointer(u, u_d)
  class(field_t), intent(in) :: u
  real(dp), device, pointer, dimension(:, :, :), intent(out) :: u_d
  select type(u)
  type is (cuda_field_t)
    u_d => u%data_d
  end select
end subroutine
```
and then the code in the main subroutine should be
```
call set_data_pointer(u, u_d)
call set_data_pointer(u_y, u_y_d)
```
which is perhaps a little better.

The other option would be to rewrite this as a wrapper subroutine accepting `class(field_t)` and then it calls the actual subroutine with `type(cuda_field_t)` arguments. The wrapper subroutine would need to do the select type dance (and should probably use `class default` for some error checking), but the actual implementation would be free if this complication which may or may not be an improvement.

semi-h:
Refactored all the transpose functions into one as we agreed, but kept the cuda kernels as they are due to performance concerns.

Also, renamed transpose and trans into reorder. The name tranpose stuck from the original Xcompact3D implementation, but the operation we're carrying out is not any more a transpose I believe. It is always local and just reorders data into x, y, or z directions so that we can use our 1D tridiagonal solver. What do you think about this? I added all the renaming stuff in a single commit and easily convert back if people disagree.

src/cuda/backend.f90
semi-h: I think we need something like this for sure, but was planning to deal with it later on when fixing the allocator.  The issue is mentioned here #24. And I think we can pass the size we need into this `set_data_pointer` function and handle the bounds remapping there.

pbartholomew08:
> Refactored all the transpose functions into one as we agreed, but kept the cuda kernels as they are due to performance concerns.
> 
> Also, renamed transpose and trans into reorder. The name tranpose stuck from the original Xcompact3D implementation, but the operation we're carrying out is not any more a transpose I believe. It is always local and just reorders data into x, y, or z directions so that we can use our 1D tridiagonal solver. What do you think about this? I added all the renaming stuff in a single commit and easily convert back if people disagree.

reorder seems sensible to me

src/cuda/backend.f90
pbartholomew08: I agree, better not to clutter the change.

semi-h:
Added a simple test, please let me know if there is anything you want to add. I plan to merge tomorrow.

Nanoseb:
Looking good to me. I think in the future we could move `checkperf` into a separate test module because it is used in other tests too. But this is outside the scope of this PR.

src/cuda/kernels_reorder.f90
```diff
@@ -0,0 +1,119 @@
...
+
+      real(dp), device, intent(out), dimension(:, :, :) :: u_z
+      real(dp), device, intent(in), dimension(:, :, :) :: u_x
+      integer, value, intent(in) :: nz
+
+      integer :: i, j, b_i, b_j, nx!, nz
+
+      i = threadIdx%x; b_i = blockIdx%x; b_j = blockIdx%y
+      nx = gridDim%x
+
```
JamieJQuinn, line 43: Can you add a comment on why x2z doesn't use the shared tile technique?

JamieJQuinn (changes requested):
Minor changes except possible logging system but that should go in its own PR anyway. Whole reordering processes seems much cleaner now.

JamieJQuinn:
Can someone also comment here the reasoning behind refactoring the individual reorder functions (e.g. `trans_x2y`) into the switch-style `reorder(u_y, u, RDR_X2Y)`?

To me, the calling code remains semantically identical so there's no reduction in complexity there, and there has to be a separate kernel for every direction anyway, so we haven't reduced complexity on the backend. I feel like I'm missing the advantage of introducing this new switch statement.

src/cuda/backend.f90
```diff
@@ -415,50 +417,50 @@ subroutine tds_solve_dist(self, du, u, dirps, tdsops, blocks, threads)
...
+         call reorder_y2z<<<blocks, threads>>>(u_o_d, u_i_d, &
+                                               self%nx_loc, self%nz_loc)
+      case (RDR_Z2Y) ! z2y
+         blocks = dim3(self%nx_loc/SZ, self%ny_loc/SZ, self%nz_loc)
+         threads = dim3(SZ, SZ, 1)
+
+         call reorder_z2y<<<blocks, threads>>>(u_o_d, u_i_d, &
+                                               self%nx_loc, self%nz_loc)
+      case default
+         print *, 'Transpose direction is undefined.'
```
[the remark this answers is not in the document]
slaizet: yes please for a more appropriate output stream!

Nanoseb:
> Can someone also comment here the reasoning behind refactoring the individual reorder functions (e.g. `trans_x2y`) into the switch-style `reorder(u_y, u, RDR_X2Y)`?


To me `reorder(... RDR_X2Y)` feels cleaner because you have a single subroutine to do the work, you don't have to setup 6 different ones, all having the same set of input and output definitions that is error prone. When reading the code it felt simpler that way to be more concise.

src/cuda/backend.f90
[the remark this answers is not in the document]
semi-h: We make use of this in the tests already. Changed the print here accordingly.

Could you give more details about a global logging system you have in mind? I haven't really used something like that. Maybe we can start an issue on this for further discussions.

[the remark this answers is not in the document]
Nanoseb: see https://github.com/xcompact3d/x3d2/issues/30

semi-h:
> Can someone also comment here the reasoning behind refactoring the individual reorder functions (e.g. `trans_x2y`) into the switch-style `reorder(u_y, u, RDR_X2Y)`?

It helped us to remove some of the backend procedures and we ended up having fewer lines.
Also, we need separate kernels in the CUDA backend but in the OpenMP backend we can actually have a single subrotuine doing all the reorderings based on the switch parameter.

tests/cuda/test_cuda_reorder.f90
```diff
@@ -0,0 +1,228 @@
+program test_cuda_reorder
```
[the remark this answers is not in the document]
semi-h: I think I'll keep it as is for now, but I'll be happy to have a chat later about the best practices regarding testing. I'm tempted to test a single functionality only and ideally isolating it from the rest, and reorder_cuda would require instantiating the backend and the allocator at least, and also setting the class variables correctly so that the reorder subroutines are run currectly. The way we set these class variables are still not finalised, and currently done in the main program. Plan is to move them into a dedicated function in the backend, which reads the inputs from a config file.

PR #31, opened 2024-02-08
Cleanup Github Actions and fix fprettify

JamieJQuinn:
This PR fixes a number of issues:

- fprettify will now ignore the entire file `src/cuda/exec_dist.f90` to avoid it messing with the CUDA chevrons
- Github Actions no longer reports fprettify failing as a true failure; instead it posts a warning, however there's now no obvious indication that a run has failed.
- Updated various github actions to latest versions and pinned their Python version.
- Removed check for Conventional Commits.

Note: the fprettify chevron bugfix has been submitted so the above hack to just remove the file entirely will be no longer needed once the [corresponding PR](https://github.com/pseewald/fprettify/pull/162) has been merged and fprettify updated.


Nanoseb:
Looks good to me. Though I'd also remove mentions of the "conventional commit" thing in the documentation. I think it is mainly here: https://github.com/xcompact3d/x3d2/blob/main/docs/source/developer/contributing_guidelines.rst

slaizet (approved):

Issue #30, opened 2024-02-06
Should there be a global logging system for better error reporting?


JamieJQuinn:
Currently errors are reported throughout the codebase with regular print statements:
```
print *, "Error: something happened"
```

While this is generally fine, a more sophisticated system that looks something like...
```
logger.warn("This is just a warning")
> WARNING: This is just a warning

logger.error("This is a catastrophic error")
> ERROR: This is a catastrophic error

logger.info("Just reporting regular information")
> INFO: Just reporting regular information
```
would allow some nice benefits:

- flexible, global redirection of different types of errors to files & other output streams
- uniform hierarchy of messages (e.g. error, warn, info, debug, etc)

pbartholomew08:
Something to maybe consider is whether to print regardless of rank or only on root, or atleast having the option to do so. Though this might run the risk of deadlocks, maybe a case only for the `info` class of messages.

ia267:
A part of this we should also include MPI/CUDA error handling:

Scope:

- Add reusable `check_mpi`, `check_cuda`, and `abort_mpi_job` helpers.
- Report rank, operation and underlying error string.
- Audit existing MPI return codes.
- Replace appropriate multi-rank `error stop` calls with common job-wide termination.

PR #32, opened 2024-02-09
Poisson solver

### LINKED ISSUE #55: Implement FFT Poisson solver (CUDA)
- [x] Single GPU (cuFFT)
- [ ] Multi GPU (cuFFTMp)

semi-h:
Poisson solver is the final part we need for running simulations with x3d2. The plan is to have an FFT based direct solver and use it whenever we can, and also have a CG based iterative solver. So I came up with this structure to have both methdos and decide which one to run at run time.

Few important considerations so that the structure makes more sense:
- FFT based solver is implemented in a dedicated class/module
  - This will make it easy to switch to 2DECOMP&FFT later on
  - Current pencil arrangement is already compatible with 2DECOMP&FFT
- FFT based poisson solver class is instantiated by the backend.
  - Different backends will use different libraries for running the FFT, I think this makes everything easy.
  - Alternative would be instantiating the poisson_fft class in solver, but I think it would make things uglier.
- A matrix free CG based solver requires to evaluate a Laplacian operation.
  -  I think it is a good idea to implement a laplacian subroutine in the solver class. We might require this in the solver later on just like gradient and divergence operations.
  - So a CG solver will basically call a laplacian subroutine in solver class, and few other stuff provided by backends. This is why it doesn't require its own class like the fft based solver.
  - @pbartholomew08, I actually focused on a standard CG method, can you comment on a preconditioned CG method and how similar or different it is compared to a standard CG implementation?
  
  
Closes #55

### DIFF
--- a/src/common.f90
+++ b/src/common.f90
@@ -7,6 +7,8 @@ module m_common
    integer, parameter :: RDR_X2Y = 12, RDR_X2Z = 13, RDR_Y2X = 21, &
                          RDR_Y2Z = 23, RDR_Z2X = 31, RDR_Z2Y = 32
 
+   integer, parameter :: POISSON_SOLVER_FFT = 0, POISSON_SOLVER_CG = 1
+
    type :: globs_t
       integer :: nx, ny, nz
       integer :: nx_loc, ny_loc, nz_loc
@@ -15,7 +17,7 @@ module m_common
       real(dp) :: dx, dy, dz
       integer :: nproc_x = 1, nproc_y = 1, nproc_z = 1
       character(len=20) :: BC_x_s, BC_x_e, BC_y_s, BC_y_e, BC_z_s, BC_z_e
-      logical :: use_fft
+      integer :: poisson_solver_type
    end type globs_t
 
 contains
--- a/src/common.f90
+++ b/src/common.f90
@@ -15,6 +15,7 @@ module m_common
       real(dp) :: dx, dy, dz
       integer :: nproc_x = 1, nproc_y = 1, nproc_z = 1
       character(len=20) :: BC_x_s, BC_x_e, BC_y_s, BC_y_e, BC_z_s, BC_z_e
+      logical :: use_fft
    end type globs_t
 
 contains
--- a/src/common.f90
+++ b/src/common.f90
@@ -15,6 +15,7 @@ module m_common
       integer :: n_groups_x, n_groups_y, n_groups_z
       real(dp) :: Lx, Ly, Lz
       real(dp) :: dx, dy, dz
+      real(dp) :: nu, dt
       integer :: nproc_x = 1, nproc_y = 1, nproc_z = 1
       character(len=20) :: BC_x_s, BC_x_e, BC_y_s, BC_y_e, BC_z_s, BC_z_e
       integer :: poisson_solver_type
--- a/src/common.f90
+++ b/src/common.f90
@@ -16,6 +16,7 @@ module m_common
       real(dp) :: Lx, Ly, Lz
       real(dp) :: dx, dy, dz
       real(dp) :: nu, dt
+      integer :: n_iters, n_output
       integer :: nproc_x = 1, nproc_y = 1, nproc_z = 1
       character(len=20) :: BC_x_s, BC_x_e, BC_y_s, BC_y_e, BC_z_s, BC_z_e
       integer :: poisson_solver_type
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -47,6 +47,7 @@ module m_cuda_backend
       procedure :: scalar_product => scalar_product_cuda
       procedure :: set_field => set_field_cuda
       procedure :: get_field => get_field_cuda
+      procedure :: init_poisson_fft => init_cuda_poisson_fft
       procedure :: transeq_cuda_dist
       procedure :: transeq_cuda_thom
       procedure :: tds_solve_dist
@@ -114,15 +115,6 @@ module m_cuda_backend
       allocate(backend%d2u_send_e_dev(SZ, 1, n_block))
       allocate(backend%d2u_recv_s_dev(SZ, 1, n_block))
       allocate(backend%d2u_recv_e_dev(SZ, 1, n_block))
-
-      if (globs%use_fft) then
-         allocate(cuda_poisson_fft_t :: backend%poisson_fft)
-
-         select type (poisson_fft => backend%poisson_fft)
-         type is (cuda_poisson_fft_t)
-            poisson_fft = cuda_poisson_fft_t(xdirps, ydirps, zdirps)
-         end select
-      end if
 
    end function init
 
@@ -612,6 +604,21 @@ module m_cuda_backend
       select type(f); type is (cuda_field_t); arr = f%data_d; end select
 
    end subroutine get_field_cuda
+
+   subroutine init_cuda_poisson_fft(self, xdirps, ydirps, zdirps)
+      implicit none
+
+      class(cuda_backend_t) :: self
+      type(dirps_t), intent(in) :: xdirps, ydirps, zdirps
+
+      allocate(cuda_poisson_fft_t :: self%poisson_fft)
+
+      select type (poisson_fft => self%poisson_fft)
+      type is (cuda_poisson_fft_t)
+         poisson_fft = cuda_poisson_fft_t(xdirps, ydirps, zdirps)
+      end select
+
+   end subroutine init_cuda_poisson_fft
 
 end module m_cuda_backend
 
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -59,12 +59,11 @@ module m_cuda_backend
 
  contains
 
-   function init(globs, allocator, xdirps, ydirps, zdirps) result(backend)
+   function init(globs, allocator) result(backend)
       implicit none
 
       class(globs_t) :: globs
       class(allocator_t), target, intent(inout) :: allocator
-      class(dirps_t), intent(in) :: xdirps, ydirps, zdirps
       type(cuda_backend_t) :: backend
 
       type(cuda_poisson_fft_t) :: cuda_poisson_fft
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -7,11 +7,13 @@ module m_cuda_backend
    use m_base_backend, only: base_backend_t
    use m_common, only: dp, globs_t, &
                        RDR_X2Y, RDR_X2Z, RDR_Y2X, RDR_Y2Z, RDR_Z2X, RDR_Z2Y
+   use m_poisson_fft, only: poisson_fft_t
    use m_tdsops, only: dirps_t, tdsops_t
 
    use m_cuda_allocator, only: cuda_allocator_t, cuda_field_t
    use m_cuda_common, only: SZ
    use m_cuda_exec_dist, only: exec_dist_transeq_3fused, exec_dist_tds_compact
+   use m_cuda_poisson_fft, only: cuda_poisson_fft_t
    use m_cuda_sendrecv, only: sendrecv_fields, sendrecv_3fields
    use m_cuda_tdsops, only: cuda_tdsops_t
    use m_cuda_kernels_dist, only: transeq_3fused_dist, transeq_3fused_subs
@@ -56,13 +58,15 @@ module m_cuda_backend
 
  contains
 
-   function init(globs, allocator) result(backend)
+   function init(globs, allocator, xdirps, ydirps, zdirps) result(backend)
       implicit none
 
       class(globs_t) :: globs
       class(allocator_t), target, intent(inout) :: allocator
+      class(dirps_t), intent(in) :: xdirps, ydirps, zdirps
       type(cuda_backend_t) :: backend
 
+      type(cuda_poisson_fft_t) :: cuda_poisson_fft
       integer :: n_halo, n_block
 
       select type(allocator)
@@ -110,6 +114,15 @@ module m_cuda_backend
       allocate(backend%d2u_send_e_dev(SZ, 1, n_block))
       allocate(backend%d2u_recv_s_dev(SZ, 1, n_block))
       allocate(backend%d2u_recv_e_dev(SZ, 1, n_block))
+
+      if (globs%use_fft) then
+         allocate(cuda_poisson_fft_t :: backend%poisson_fft)
+
+         select type (poisson_fft => backend%poisson_fft)
+         type is (cuda_poisson_fft_t)
+            poisson_fft = cuda_poisson_fft_t(xdirps, ydirps, zdirps)
+         end select
+      end if
 
    end function init
 
--- a/src/cuda/kernels/complex.f90
+++ b/src/cuda/kernels/complex.f90
@@ -1,0 +1,99 @@
+module m_cuda_complex
+   use cudafor
+
+   use m_common, only: dp
+   use m_cuda_common, only: SZ
+
+   implicit none
+
+contains
+
+   attributes(global) subroutine processfftdiv( &
+         div, waves, nx, ny, nz, ax, bx, ay, by, az, bz &
+         )
+      implicit none
+
+      ! Arguments
+      complex(dp), device, intent(inout), dimension(:, :, :) :: div
+      complex(dp), device, intent(in), dimension(:, :, :) :: waves
+      real(dp), device, intent(in), dimension(:) :: ax, bx, ay, by, az, bz
+      integer, value, intent(in) :: nx, ny, nz
+
+      ! Local variables
+      integer :: i, j, b, ix, iy, iz
+      real(dp) :: tmp_r, tmp_c, div_r, div_c
+
+      i = threadIdx%x
+      b = blockIdx%x
+
+      do j = 1, nx
+         ! normalisation
+         div_r = real(div(i, j, b), kind=dp)/(nx*ny*nz)
+         div_c = aimag(div(i, j, b))/(nx*ny*nz)
+
+         ! get the indices for x, y, z directions
+         ix = j; iy = i + (b-1)/(nz/2+1)*SZ; iz = mod(b-1, nz/2+1) + 1
+
+         ! post-process forward
+         ! post-process in z
+         tmp_r = div_r
+         tmp_c = div_c
+         div_r = tmp_r*bz(iz) + tmp_c*az(iz)
+         div_c = tmp_c*bz(iz) - tmp_r*az(iz)
+
+         ! post-process in y
+         tmp_r = div_r
+         tmp_c = div_c
+         div_r = tmp_r*by(iy) + tmp_c*ay(iy)
+         div_c = tmp_c*by(iy) - tmp_r*ay(iy)
+         if ( iy > ny/2 + 1 ) div_r = -div_r
+         if ( iy > ny/2 + 1 ) div_c = -div_c
+
+         ! post-process in x
+         tmp_r = div_r
+         tmp_c = div_c
+         div_r = tmp_r*bx(ix) + tmp_c*ax(ix)
+         div_c = tmp_c*bx(ix) - tmp_r*ax(ix)
+         if ( ix > nx/2 + 1 ) div_r = -div_r
+         if ( ix > nx/2 + 1 ) div_c = -div_c
+
+         ! Solve Poisson
+         tmp_r = real(waves(i, j, b), kind=dp)
+         tmp_c = aimag(waves(i, j, b))
+         if ((tmp_r < 1.e-16_dp) .or. (tmp_c < 1.e-16_dp)) then
+            div_r = 0._dp; div_c = 0._dp
+         else
+            div_r =-div_r/tmp_r
+            div_c =-div_c/tmp_c
+         end if
+
+         ! post-process backward
+         ! post-process in z
+         tmp_r = div_r
+         tmp_c = div_c
+         div_r = tmp_r*bz(iz) - tmp_c*az(iz)
+         div_c =-tmp_c*bz(iz) - tmp_r*az(iz)
+
+         ! post-process in y
+         tmp_r = div_r
+         tmp_c = div_c
+         div_r = tmp_r*by(iy) + tmp_c*ay(iy)
+         div_c = tmp_c*by(iy) - tmp_r*ay(iy)
+         if ( iy > ny/2 + 1 ) div_r = -div_r
+         if ( iy > ny/2 + 1 ) div_c = -div_c
+
+         ! post-process in x
+         tmp_r = div_r
+         tmp_c = div_c
+         div_r = tmp_r*bx(ix) + tmp_c*ax(ix)
+         div_c =-tmp_c*bx(ix) + tmp_r*ax(ix)
+         if ( ix > nx/2 + 1 ) div_r = -div_r
+         if ( ix > nx/2 + 1 ) div_c = -div_c
+
+         ! update the entry
+         div(i, j, b) = cmplx(div_r, div_c, kind=dp)
+      end do
+
+   end subroutine processfftdiv
+
+end module m_cuda_complex
--- a/src/cuda/kernels/complex.f90
+++ b/src/cuda/kernels/complex.f90
@@ -8,7 +8,7 @@ module m_cuda_complex
 
 contains
 
-   attributes(global) subroutine processfftdiv( &
+   attributes(global) subroutine process_spectral_div_u( &
          div, waves, nx, ny, nz, ax, bx, ay, by, az, bz &
          )
       implicit none
@@ -63,8 +63,8 @@ contains
          if ((tmp_r < 1.e-16_dp) .or. (tmp_c < 1.e-16_dp)) then
             div_r = 0._dp; div_c = 0._dp
          else
-            div_r =-div_r/tmp_r
-            div_c =-div_c/tmp_c
+            div_r = -div_r/tmp_r
+            div_c = -div_c/tmp_c
          end if
 
          ! post-process backward
@@ -72,7 +72,7 @@ contains
          tmp_r = div_r
          tmp_c = div_c
          div_r = tmp_r*bz(iz) - tmp_c*az(iz)
-         div_c =-tmp_c*bz(iz) - tmp_r*az(iz)
+         div_c = -tmp_c*bz(iz) - tmp_r*az(iz)
 
          ! post-process in y
          tmp_r = div_r
@@ -86,7 +86,7 @@ contains
          tmp_r = div_r
          tmp_c = div_c
          div_r = tmp_r*bx(ix) + tmp_c*ax(ix)
-         div_c =-tmp_c*bx(ix) + tmp_r*ax(ix)
+         div_c = -tmp_c*bx(ix) + tmp_r*ax(ix)
          if (ix > nx/2 + 1) div_r = -div_r
          if (ix > nx/2 + 1) div_c = -div_c
 
@@ -94,7 +94,7 @@ contains
          div(i, j, b) = cmplx(div_r, div_c, kind=dp)
       end do
 
-   end subroutine processfftdiv
+   end subroutine process_spectral_div_u
 
    attributes(global) subroutine reorder_cmplx_x2y_T(u_y, u_x, nz)
       implicit none
--- a/src/cuda/kernels/complex.f90
+++ b/src/cuda/kernels/complex.f90
@@ -208,4 +208,88 @@ contains
 
    end subroutine reorder_cmplx_z2y_T
 
+   attributes(global) subroutine reshapeDSF(uout, uin)
+      implicit none
+
+      real(dp), device, intent(out), dimension(:, :, :) :: uout
+      real(dp), device, intent(in), dimension(:, :, :) :: uin
+
+      real(dp), shared :: tile(SZ + 1, SZ)
+
+      integer :: i, j, b_i, b
+
+      i = threadIdx%x; j = threadIdx%y
+      b_i = blockIdx%x; b = blockIdx%y
+
+      tile(i, j) = uin(i, j + (b_i - 1)*SZ, b)
+
+      call syncthreads()
+
+      uout(i + (b_i - 1)*SZ, j, b) = tile(j, i)
+
+   end subroutine reshapeDSF
+
+   attributes(global) subroutine reshapeDSB(uout, uin)
+      implicit none
+
+      real(dp), device, intent(out), dimension(:, :, :) :: uout
+      real(dp), device, intent(in), dimension(:, :, :) :: uin
+
+      real(dp), shared :: tile(SZ + 1, SZ)
+
+      integer :: i, j, b_i, b
+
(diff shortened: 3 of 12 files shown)


pbartholomew08:
To get good performance (at scale) you need a preconditioner. The preconditioner is typically based on an actual matrix, e.g. a low order discretisation. I'm working on an implementation of this here https://github.com/3decomp/poissbox/pull/20, once I've got the preconditioner sorted I'll port it over to x3d2.

I don't think we want to implement CG ourselves though!

pbartholomew08:
>     * A matrix free CG based solver requires to evaluate a Laplacian operation.
>       
>       * I think it is a good idea to implement a laplacian subroutine in the solver class. We might require this in the solver later on just like gradient and divergence operations.

For consistency of the discretisation the Laplacian operator should be built by combining the div and grad operators.

src/poisson_fft.f90
```diff
@@ -0,0 +1,114 @@
...
+      allocate (self%az(self%nx), self%bz(self%nx))
+
+      allocate (self%waves(self%nx, self%ny, self%nz))
+
+      ! waves_set requires some of the preprocessed tdsops variables.
+      call self%waves_set(xdirps, ydirps, zdirps)
+
+   end subroutine base_init
+
+   subroutine waves_set(self, xdirps, ydirps, zdirps)
```
JamieJQuinn, line 72: Needs a high-level description; no idea what `waves_set` means.

src/cuda/backend.f90
```diff
@@ -51,13 +53,15 @@ module m_cuda_backend
 
  contains
 
-   function init(globs, allocator) result(backend)
+   function init(globs, allocator, xdirps, ydirps, zdirps) result(backend)
```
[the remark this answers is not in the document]
JamieJQuinn: Also need a description somewhere of what xdirps, etc are.

src/common.f90
```diff
@@ -15,6 +15,7 @@ module m_common
       real(dp) :: dx, dy, dz
       integer :: nproc_x = 1, nproc_y = 1, nproc_z = 1
       character(len=20) :: BC_x_s, BC_x_e, BC_y_s, BC_y_e, BC_z_s, BC_z_e
+      logical :: use_fft
```
JamieJQuinn, line 18: If we plan to have more than just two Poisson solvers, should this be an option instead? e.g. 

```fortran
integer, parameter :: POISSON_SOLVER_FFT = 0, POISSON_SOLVER_CG = 1, POISSON_SOLVER_PCG = 2, ...
...
integer :: poisson_solver_type = POISSON_SOLVER_FFT
select case (poisson_solver_type)
...
```

In fact, might be a good place for an enum (f2003 wow how modern):

```fortran
enum, bind(c)
   enumerator POISSON_SOLVER_FFT, POISSON_SOLVER_CG, POISSON_SOLVER_PCG
endenum
```

JamieJQuinn (changes requested):

JamieJQuinn:
The overall structure looks good. I like the simple calling code in the solve. Seems unlikely that the CG solver won't require some backend-specific stuff (so might need its own class) but the generic procedure can handle that case well, and can extend to other options if needed.

semi-h:
> For consistency of the discretisation the Laplacian operator should be built by combining the div and grad operators.

This is very interesting, I wasn't expecting this at all. But I think the gradient and divergence subroutines in solver is not the exact ones you need, right? Because they switch between velocity grid and pressure grid. I think the iterative solver will require both gradient and divergence starting from the pressure grid and ending in the pressure grid. If so we'll need a new set of tdsops to support this operation and add two new grad and div subroutines later on.

semi-h:
> The overall structure looks good. I like the simple calling code in the solve. Seems unlikely that the CG solver won't require some backend-specific stuff (so might need its own class) but the generic procedure can handle that case well, and can extend to other options if needed.

Actually we can access the backend stuff from the solver, all methods in solver do it already. For example a naive CG implementation would only require a laplacian and vector squares I guess.
And technically we could add the FFT related stuff into the backends and not require seperate class as well, but I didn't do it for two reasons.
- FFT based solver requires lots of parameters, settings, complex arrays..., it would make the backends complicated. Because it comes with bunch of complicated stuff that only relates to an FFT based approach, it made sense to me to separate it into its own class and keep the complexity away from backend. (all this is not in the PR yet)
- If we switch to a third-party library (2DECOMP&FFT), then we obviously can't call things from our backend, and all those stuff would have to be removed in case of a switch, and also having a separate class now should help later on when switching I think.

But if we're doing the CG stuff via PETSc, then its very likely that we'll need a class to keep PETSc related complexity away from backends too.

src/poisson_fft.f90
semi-h: This subroutine is not finalised yet, I'll add a description later when its finished. I realised at this stage that it needs some stuff in dirps that we didn't bother storing at tdsops class level.

pbartholomew08:
> > For consistency of the discretisation the Laplacian operator should be built by combining the div and grad operators.
> 
> This is very interesting, I wasn't expecting this at all. But I think the gradient and divergence subroutines in solver is not the exact ones you need, right? Because they switch between velocity grid and pressure grid. I think the iterative solver will require both gradient and divergence starting from the pressure grid and ending in the pressure grid. If so we'll need a new set of tdsops to support this operation and add two new grad and div subroutines later on.

The pattern of execution is `p (p grid) -> grad p (v grid) -> lapl p (p grid)`, because then this is consistent with the use of `grad(p)` in the momentum equations and the evaluation of `div(u)`. I'll try just evaluating the Laplacian directly using 2nd derivatives, maybe this will work better.

src/common.f90
pbartholomew08: alternatively (and especially if we need setup) we could do this through the types, e.g.
```
class(poisson_solver), allocatable :: poiss

! This should be an initialisation subroutine...
allocate(poisson_fft :: poiss) ! poisson_cg, etc.
poiss%init() ! Do solver-specific initialisation

call solve(poiss) ! This will type-resolve to the specific Poisson solver
```

pbartholomew08: I see you've mentioned this possibility https://github.com/xcompact3d/x3d2/pull/32#issuecomment-1935747550

src/cuda/backend.f90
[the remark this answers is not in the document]
semi-h: #46

semi-h:
Just to let you know when reviewing, all the reorder and reshape functions in complex.f90 will be removed when we switch to cuFFTMp.
Also, the plans in the init function will be adopted to cuFFTMp, and the fft_forward and fft_backward functions will be simplified down to single line calls from cuFFTMp.

Only thing I want to add before we merge is a periodic call to curl and divergence functions to report enstrophy and divergence of velocity field as the simulation goes on. We need a simple inner product function in the backends to finalise that.

src/common.f90
JamieJQuinn: @semi-h Thoughts on this?

JamieJQuinn:
> Just to let you know when reviewing, all the reorder and reshape functions in complex.f90 will be removed when we switch to cuFFTMp.

Can you mark these with a comment like "TODO remove after cuFFTMp" and create an issue for it?

src/common.f90
semi-h: I think both your and Paul's suggestions are good, I think I'll leave this as is so that Paul can go ahead with one of the approaches based on how he wants to implement the iterative solver. Is that okay, @pbartholomew08?

semi-h: All done.

semi-h:
A call for last round of review folks!

The input settings in the main file is arranged so that it produces the result I shared in the slack channel. Basically a TGV from 0 to 20 on a 256^3 grid.

The output subroutine is not very good but for the time being useful to see if there is anything going bad with the TGV, because it prints the enstrophy. I suppose this will evolve a lot into more useful output stuff later.

> > Just to let you know when reviewing, all the reorder and reshape functions in complex.f90 will be removed when we switch to cuFFTMp.
> 
> Can you mark these with a comment like "TODO remove after cuFFTMp" and create an issue for it?

They won't be being called once the cuFFTMp stuff is in and then like any redundant function they'll need to be removed. However we might want to compare performance of the two approaches so for a short while both the current FFT solver and cuFFTMp may coexist too. I think it'll be more clear with the cuFFTMp PR so not going ahead with TODO comments for now.

src/cuda/kernels/complex.f90
```diff
@@ -0,0 +1,295 @@
...
+
+      tile(i, j) = uin(i, j + (b_i - 1)*SZ, b)
+
+      call syncthreads()
+
+      uout(i + (b_i - 1)*SZ, j, b) = tile(j, i)
+
+   end subroutine reshapeCDSF
+
+   attributes(global) subroutine reshapeCDSB(uout, uin)
```
JamieJQuinn, line 274: CDSB?

src/cuda/poisson_fft.f90
```diff
@@ -0,0 +1,239 @@
...
+
+   implicit none
+
+   type, extends(poisson_fft_t) :: cuda_poisson_fft_t
+      !! FFT based Poisson solver
+      !! It can only handle 1D decompositions along z direction.
+      complex(dp), device, pointer, dimension(:) :: c_x_dev, c_y_dev, c_z_dev
+      complex(dp), device, allocatable, dimension(:, :, :) :: waves_dev
+
+      real(dp), device, allocatable, dimension(:) :: ax_dev, bx_dev, &
```
JamieJQuinn, line 24: What are a, b and c?

src/cuda/poisson_fft.f90
```diff
@@ -0,0 +1,239 @@
...
+      blocks = dim3(self%nz/2/SZ + 1, self%ny/SZ, self%nx)
+      threads = dim3(SZ, SZ, 1)
+      c_y_ptr(1:self%ny, 1:SZ, 1:(self%nx*(self%nz/2 + 1))/SZ) => self%c_y_dev
+      c_z_ptr(1:self%nz/2 + 1, 1:SZ, 1:self%nx*self%ny/SZ) => self%c_z_dev
+
+      call reorder_cmplx_z2y_T<<<blocks, threads>>>(c_y_ptr, c_z_ptr, &
+                                                    self%nx, self%nz/2 + 1)
+
+      ! In-place forward FFT in y
+      ierrfft = cufftExecZ2Z(self%planZ2Zy, self%c_y_dev, self%c_y_dev, &
```
JamieJQuinn, line 137: Error is captured but not checked? Do we need to?

src/poisson_fft.f90
```diff
@@ -0,0 +1,198 @@
...
+   end subroutine base_init
+
+   subroutine waves_set(self, xdirps, ydirps, zdirps, sz)
+      implicit none
+
+      class(poisson_fft_t) :: self
+      type(dirps_t), intent(in) :: xdirps, ydirps, zdirps
+      integer, intent(in) :: sz
+
+      complex(dp), allocatable, dimension(:) :: xkx, xk2, yky, yk2, zkz, zk2, &
```
JamieJQuinn, line 78: These need more descriptive names or comments describing what they are

src/poisson_fft.f90
```diff
@@ -0,0 +1,198 @@
...
+         wp = zdirps%stagder_v2p%a*2*zdirps%d*sin(0.5_dp*w) &
+              + zdirps%stagder_v2p%b*2*zdirps%d*sin(1.5_dp*w)
+         wp = wp/(1._dp + 2*zdirps%stagder_v2p%alpha*cos(w))
+
+         zkz(i) = cmplx(1._dp, 1._dp, kind=dp)*(nz*wp/zdirps%L)
+         ezs(i) = cmplx(1._dp, 1._dp, kind=dp)*(nz*w/zdirps%L)
+         zk2(i) = cmplx(1._dp, 1._dp, kind=dp)*(nz*wp/zdirps%L)**2
+      end do
+
+      print*, 'set waves array'
```
JamieJQuinn, line 157: Is this debug output?

src/poisson_fft.f90
```diff
@@ -0,0 +1,198 @@
...
+      allocate (self%az(self%nx), self%bz(self%nx))
+
+      allocate (self%waves(sz, self%nx, (self%ny*(self%nz/2 + 1))/sz))
+
+      ! waves_set requires some of the preprocessed tdsops variables.
+      call self%waves_set(xdirps, ydirps, zdirps, sz)
+
+   end subroutine base_init
+
+   subroutine waves_set(self, xdirps, ydirps, zdirps, sz)
```
JamieJQuinn, line 71: Unclear what this sub does. Needs a comment?

src/poisson_fft.f90
```diff
@@ -0,0 +1,198 @@
...
+              + zdirps%stagder_v2p%b*2*zdirps%d*sin(1.5_dp*w)
+         wp = wp/(1._dp + 2*zdirps%stagder_v2p%alpha*cos(w))
+
+         zkz(i) = cmplx(1._dp, 1._dp, kind=dp)*(nz*wp/zdirps%L)
+         ezs(i) = cmplx(1._dp, 1._dp, kind=dp)*(nz*w/zdirps%L)
+         zk2(i) = cmplx(1._dp, 1._dp, kind=dp)*(nz*wp/zdirps%L)**2
+      end do
+
+      print*, 'set waves array'
+      ! TODO: do loop ranges below are valid only for single rank runs
```
JamieJQuinn, line 158: Should this / can this be checked here and program aborted if conditions aren't met?

src/cuda/poisson_fft.f90
```diff
@@ -0,0 +1,239 @@
...
+
+      ierrfft = cufftCreate(poisson_fft%planZ2Zx)
+      ierrfft = cufftMakePlanMany(poisson_fft%planZ2Zx, 1, nx, &
+                                  nx, 1, nx, nx, 1, nx, &
+                                  CUFFT_Z2Z, ny*(nz/2 + 1), worksize)
+      ierrfft = cufftSetWorkArea(poisson_fft%planZ2Zx, poisson_fft%c_y_dev)
+
+   end function init
+
+   subroutine fft_forward_cuda(self, f)
```
JamieJQuinn, line 104: Would be good to have working tests for these before trying to implement cuFFTMP.

JamieJQuinn (changes requested):
Overall excellent. Needs comments + possible name changes for clarity & consistency. Might want tests but this PR is already large. Should be an issue if we decide not to put tests in here.

src/cuda/poisson_fft.f90
```diff
@@ -0,0 +1,239 @@
...
+   use m_tdsops, only: dirps_t
+
+   use m_cuda_allocator, only: cuda_field_t
+   use m_cuda_common, only: SZ
+   use m_cuda_complex, only: reorder_cmplx_x2y_T, reorder_cmplx_y2x_T, &
+                             reorder_cmplx_y2z_T, reorder_cmplx_z2y_T, &
+                             processfftdiv
+
+   implicit none
+
```
pbartholomew08, line 17: Might want to make the module `private` - I think, e.g., `init()` is declared in several places and (potentially) is not desired to be visible outside the module.

src/cuda/poisson_fft.f90
```diff
@@ -0,0 +1,239 @@
...
+                                                          c_z_ptr
+
+      type(dim3) :: blocks, threads
+      integer :: ierrfft
+
+      select type(f); type is (cuda_field_t); f_dev => f%data_d; end select
+
+      ! First reorder f into cartesian-like data structure
+      blocks = dim3(self%nz/SZ, self%nx*self%ny/SZ, 1)
+      threads = dim3(SZ, SZ, 1)
```
pbartholomew08, line 121: This pattern (or similar) occurs several times. Potentially worth capturing as a function/subroutine.

pbartholomew08: I might even capture each commented "block" of code as a subroutine/function - the comment implies that there is a named concept, but this is personal preference

src/cuda/poisson_fft.f90
```diff
@@ -0,0 +1,239 @@
...
+      complex(dp), device, dimension(:, :, :), pointer :: c_dev, c_x_ptr
+      type(dim3) :: blocks, threads
+
+      ! Get some complex array pointers with right shape
+      c_x_ptr(1:self%nx, 1:SZ, 1:(self%ny*(self%nz/2 + 1))/SZ) => self%c_x_dev
+      c_dev(1:SZ, 1:self%nx, 1:(self%ny*(self%nz/2 + 1))/SZ) => self%c_y_dev
+
+      ! Reshape from cartesian-like to our specialist data structure
+      blocks = dim3(self%nx/SZ, (self%ny*(self%nz/2 + 1))/SZ, 1)
+      threads = dim3(SZ, SZ, 1)
```
pbartholomew08, line 220: As previously commented for similar patterns

src/omp/poisson_fft.f90
```diff
@@ -0,0 +1,58 @@
+module m_omp_poisson_fft
+   use m_allocator, only: field_t
+   use m_common, only: dp
+   use m_poisson_fft, only: poisson_fft_t
+   use m_tdsops, only: dirps_t
+
+   use m_omp_common, only: SZ
+
+   implicit none
+
```
pbartholomew08, line 10: As for CUDA module, both have an `init` function, it might be worth making this `private` and explicitly exporting subroutines/functions/etc

src/solver.f90
```diff
@@ -37,23 +38,39 @@ module m_solver
       !! for later use.
 
       real(dp) :: dt, nu
+      integer :: n_iters, n_output
```
pbartholomew08, line 41: Are these different pr the same as the values in `src/common.f90`?

pbartholomew08 (commented):
A few minor comments for consideration. Otherwise I agree with Jamie's points.

src/poisson_fft.f90
semi-h: To be honest, this is a black box for me too, I copied this from Incompact3D and there were no comments there either. 
@slaizet is probably the best person to add a description and few comments here if he has time and happy to do this, otherwise someone else who fully understands whats going on here, definitely not me. And this can be done in a separate PR not to delay this one any further. 
For now I'm only adding a description to reference the paper where its discussed in detail. 

https://github.com/xcompact3d/Incompact3d/blob/2da7699558806e70574a6980a75193a80f5734c9/src/poisson.f90#L1553

slaizet: I think best would be to seat together in the same room one afternoon and add as many comments as required (when at least one of us does not understand something basically). If we are all together, it should be quite quick and productive :-) We can discuss about this during our next weekly meeting on when best to add these comments in the new framework.

src/solver.f90
semi-h: They're the same, but we need them here as well because `globs_t` is not a member of the solver class yet. Input parameters right now handled very poorly, we'll need to work on them when we're adding the capability to read input parameters from an input file in a later PR.

pbartholomew08: ah I see

src/omp/poisson_fft.f90
semi-h: I thought using the `only` keyword strictly when `use`ing a module and explicitly listing things we want to import is safe, but I see what you mean with the `init` function.

Would it work if we declare `private :: init` and continue using `only` whenever we import stuff? With this we can avoid explicitly stating the public ones.

pbartholomew08: I'm not sure, I think you are probably right about using `only`. Making things explicitly `private` as you suggest would also work I think, it depends whether you want the module to be self-contained (default `private`) or only a few things should be inaccessible (default `public`)

pbartholomew08:
> > > For consistency of the discretisation the Laplacian operator should be built by combining the div and grad operators.
> > 
> > 
> > This is very interesting, I wasn't expecting this at all. But I think the gradient and divergence subroutines in solver is not the exact ones you need, right? Because they switch between velocity grid and pressure grid. I think the iterative solver will require both gradient and divergence starting from the pressure grid and ending in the pressure grid. If so we'll need a new set of tdsops to support this operation and add two new grad and div subroutines later on.
> 
> The pattern of execution is `p (p grid) -> grad p (v grid) -> lapl p (p grid)`, because then this is consistent with the use of `grad(p)` in the momentum equations and the evaluation of `div(u)`. I'll try just evaluating the Laplacian directly using 2nd derivatives, maybe this will work better.

Adding for reference (I think we've discussed this in meetings) the direct evaluation of the Laplacian using high-order methods for 2nd derivatives on the pressure grid seems to work well (as a standalone solver).

src/poisson_fft.f90
semi-h: I don't prefer aborting the program when we encounter a setting we don't support yet, because I routinely check the performance with under such settings. For example, there are bits of the code where multi-rank is not supported (or bits that are not implemented), so running the solver would definitely give a wrong result, but it is sometimes perfectly representative in terms of performance. This is helping me catching performance related stuff early on and avoiding potential suprises when we run a feature first time after supporting it.

I avoid TODOs as much as I can but here the point is very subtle so I wanted to highlight it at least.

src/cuda/poisson_fft.f90
semi-h: I haven't really encountered an issue with this so I think okay for the moment. But I checked 2DECOMP&FFT and the error is always check there, so I think there is value and in general its good practice anyways. I'll implement the error checks with the cuFFTMp stuff, there'll be a lot less such function calls then and will be easier.

src/cuda/kernels/complex.f90
```diff
@@ -0,0 +1,295 @@
...
+         if (ix > nx/2 + 1) div_r = -div_r
+         if (ix > nx/2 + 1) div_c = -div_c
+
+         ! update the entry
+         div(i, j, b) = cmplx(div_r, div_c, kind=dp)
+      end do
+
+   end subroutine processfftdiv
+
+   attributes(global) subroutine reorder_cmplx_x2y_T(u_y, u_x, nz)
```
Nanoseb, line 99: Would there be a way to not rewrite entirely the reorder functions just because we are dealing with complex here?
Also, It would be good to test those as well. Once again, they are very easy to get slightly wrong.

src/poisson_fft.f90
Nanoseb: In this case, it seems fairly limiting, and I'd argue it would require more than a "TODO" comment. If we aren't aborting, having it at least print to the output would be good. 
Once we reach 'production', I think we should reassess this, as I agree that aborting would be better.

src/solver.f90
```diff
@@ -100,13 +139,20 @@ function init(backend, time_integrator, xdirps, ydirps, zdirps, globs) &
...
-      call allocate_tdsops(solver%xdirps, nx, dx, solver%backend)
-      call allocate_tdsops(solver%ydirps, ny, dy, solver%backend)
-      call allocate_tdsops(solver%zdirps, nz, dz, solver%backend)
+      call allocate_tdsops(solver%xdirps, nx, globs%dx, solver%backend)
+      call allocate_tdsops(solver%ydirps, ny, globs%dy, solver%backend)
+      call allocate_tdsops(solver%zdirps, nz, globs%dz, solver%backend)
+
+      select case (globs%poisson_solver_type)
+      case (POISSON_SOLVER_FFT)
+         print*, 'Poisson solver: FFT'
```
Nanoseb, line 149: These prints should only be done by rank 0. Though maybe we can wait for the logging module (#30) to clean this up.

Nanoseb (commented):
I agree with Paul and Jamie's comments, just added a couple minor ones.

src/poisson_fft.f90
semi-h: In the production code yes of course, aborting in case of incompatible settings is definitely the solution. We can't expect users to know what we support and set things accordingly.

We're now printing that waves array is correct only for a single rank run.

semi-h:
Last commit fixed most issues raised.

There are a few comments about the reshape and complex_reorder functions, I admit that their naming scheme is terrible and there is a lot to do make them better, but there is a very good chance that cuFFTMp will work fine and they'll all be removed from the codebase entirely very soon. Because we expect their lifespan to be very short, I don't want to invest time to think about a proper naming scheme, adding comments, change the way we set threads/blocks for calling them, or testing them and willing to merge the PR as is. If we decide to keep them longer for any reason I'll be happy to implement all the suggestions.

src/cuda/poisson_fft.f90
```diff
@@ -0,0 +1,239 @@
...
+   use m_cuda_complex, only: reorder_cmplx_x2y_T, reorder_cmplx_y2x_T, &
+                             reorder_cmplx_y2z_T, reorder_cmplx_z2y_T, &
+                             processfftdiv
+
+   implicit none
+
+   type, extends(poisson_fft_t) :: cuda_poisson_fft_t
+      !! FFT based Poisson solver
+      !! It can only handle 1D decompositions along z direction.
+      complex(dp), device, pointer, dimension(:) :: c_x_dev, c_y_dev, c_z_dev
```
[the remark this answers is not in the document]
JamieJQuinn: Sorry, I meant the naming seems inconsistent, one is letter_[xyz]_dev and the other letter[xyz]_dev. I don't mind which it is but seems strange to have both.

src/cuda/kernels/complex.f90
JamieJQuinn: Testing functions we're likely to throw away seems a waste.. but throwing away the functions also seems potentially wasteful :shrug: 

@semi-h Why do the reorder fns need to be thrown away when cuFFTMP comes in?

src/poisson_fft.f90
JamieJQuinn: Raised in #72

src/poisson_fft.f90
JamieJQuinn: Addressed by #72

src/cuda/kernels/complex.f90
```diff
@@ -0,0 +1,295 @@
...
+      call syncthreads()
+
+      ! copy into output array from shared
+      if ( j + (b_z - 1)*SZ <= nz ) &
+         u_y(i + (b_y - 1)*SZ, mod(b_x - 1, SZ) + 1, &
+             j + (b_z - 1)*SZ + ((b_x - 1)/SZ)*nz) = tile(j, i)
+
+   end subroutine reorder_cmplx_z2y_T
+
+   attributes(global) subroutine reshapeDSF(uout, uin)
```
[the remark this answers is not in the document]
JamieJQuinn: Addressed by #73

src/cuda/kernels/complex.f90
```diff
@@ -0,0 +1,295 @@
...
+
+      tile(i, j) = uin(i + (b_i - 1)*SZ, j, b)
+
+      call syncthreads()
+
+      uout(i, j + (b_i - 1)*SZ, b) = tile(j, i)
+
+   end subroutine reshapeDSB
+
+   attributes(global) subroutine reshapeCDSF(uout, uin)
```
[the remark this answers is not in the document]
JamieJQuinn: Addressed by #73

src/cuda/kernels/complex.f90
JamieJQuinn: Addressed by #73

src/cuda/poisson_fft.f90
JamieJQuinn: @semi-h implement in this PR or create an issue?

src/solver.f90
JamieJQuinn: Explicitly added cleaning up existing print statements to #30

src/cuda/poisson_fft.f90
JamieJQuinn: Can be addressed as part of #73

src/cuda/poisson_fft.f90
JamieJQuinn: Can be addressed as part of #73

src/cuda/kernels/complex.f90
semi-h: We just need to transform the divergence of velocity into spectral space (3D forward FFT) so that we can postprocess it to obtain pressure field in spectral space. Then we apply an inverse FFT in 3D and bring the pressure field from spectral space back to physical space. Currently we implement our own 3D FFT transforms, both forward and inverse. And this requires a series of 1D FFT calls and those reshapes/reorders in between.

cuFFTMp will allow us to carry out the 3D forward or inverse FFTs with a single call to the library.

src/cuda/poisson_fft.f90
semi-h: Actually, cuFFTMp provides 3D forward and inverse FFT transforms, and our functions here will only carry out the calls to cuFFTMp library so probably there is no value testing them.

src/cuda/poisson_fft.f90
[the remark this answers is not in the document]
semi-h: Ah I see, `c_x_dev` follows our convention and indicates the orientation of the array, but `ax` and others are relevant to `waves_set` and `postprocess..`. We can find a more descriptive name for them with #72

src/poisson_fft.f90
JamieJQuinn: @semi-h ?

src/solver.f90
```diff
@@ -489,20 +535,86 @@ subroutine curl(self, o_i_hat, o_j_hat, o_k_hat, u, v, w)
...
+
+   end subroutine poisson_fft
+
+   subroutine poisson_cg(self, pressure, div_u)
+      implicit none
+
+      class(solver_t) :: self
+      class(field_t), intent(inout) :: pressure
+      class(field_t), intent(in) :: div_u
+
```
JamieJQuinn, line 563:
```suggestion
print*, "CG Poisson solver not yet implemented."
```

src/cuda/poisson_fft.f90
JamieJQuinn: For the forward/backward FFTs sure, no need to test individually. But we're currently not unit testing the Poisson solver at all. Since we have a working implementation, I think it would be easiest to implement a correct test of this working solver, then apply the same test to cuFFTMP.

If you don't already have one kicking around, the analytical test case I usually use for Poisson solve is solving $\nabla^2 p = -\sin(\pi x) \sin(\pi y)$ and comparing to analytical solution $p_{analytical}(x,y) = \sin (\pi x) \sin(\pi y)/(2\pi^2)$.

JamieJQuinn (changes requested):
This is nearly done! Looks like there are a few minor things still needing addressed, and a proper test of the Poisson solver implemented.

semi-h:
Its a good idea to have a unit test for the Poisson solver, but I don't want to implement it here as the PR is already large. Opened an issue to deal with it later on #74.

src/solver.f90
semi-h: Ah, I committed this already but had to take it back. Printing here would result in unnecessary output every iteration when doing a performance test. Now we clarify CG is not implemented up when we initiate the poisson solver

JamieJQuinn (approved):
OK, I'm happy with the TGV showing the Poisson solver works at this stage.

PR #34, opened 2024-02-12
Make allocator more flexible with domain sizes

semi-h:
The problem is described in #24. I wanted to have a draft PR here so that we can plan ahead how to make this transition in both backends.

In the CUDA backend we always type check and assing a pointer to the `%data_d` component of `field_t`. Fortran allows a bounds remapping with pointer assignments and we can use this to have a 3D pointer pointing a 1D `data_d` array. My plan is to move all the type selection bits into a subroutine in CUDA backend like @pbartholomew08 suggested in a previous PR, and sort the bounds remapping there.

With the OpenMP backend the current way allocator works is very handy, but obviously we need to support different nx, ny, and nz and currently it is not possible. I think currently the `%data` component in `field_t` is passed directly into the kernels and it doesn't allow fixing the dimensions of `data` array when switching form x, y, and z orientations. My suggestion is again using bounds remapping and pointing to the 1D `data` array with the right dimensions based on the x, y, z orientations we're at. What do you think, @Nanoseb?

Alternative to this bounds remaping strategy is using assumed shape arrays in kernels but I think it might be more complicated especially with different velocity grid dimensions and pressure grid dimensions. I think padding is a good strategy to deal with this and assumed shape arrays would't mix well with with padding I believe.

### DIFF
--- a/src/allocator.f90
+++ b/src/allocator.f90
@@ -38,9 +38,9 @@ module m_allocator
       !> incremented each time a new block is allocated.
       integer :: next_id = 0
       !> Padded dimensions for x, y, and z oriented fields
-      integer :: xdims(3), ydims(3), zdims(3)
+      integer :: xdims_padded(3), ydims_padded(3), zdims_padded(3)
       !> Padded dimensions for natural Cartesian ordering
-      integer :: cdims(3)
+      integer :: cdims_padded(3)
       !> The pointer to the first block on the list.  Non associated if
       !> the list is empty
       ! TODO: Rename first to head
@@ -119,10 +119,10 @@ contains
       allocator%ngrid = nx_padded*ny_padded*nz_padded
       allocator%sz = sz
 
-      allocator%xdims = [sz, nx_padded, ny_padded*nz_padded/sz]
-      allocator%ydims = [sz, ny_padded, nx_padded*nz_padded/sz]
-      allocator%zdims = [sz, nz_padded, nx_padded*ny_padded/sz]
-      allocator%cdims = [nx_padded, ny_padded, nz_padded]
+      allocator%xdims_padded = [sz, nx_padded, ny_padded*nz_padded/sz]
+      allocator%ydims_padded = [sz, ny_padded, nx_padded*nz_padded/sz]
+      allocator%zdims_padded = [sz, nz_padded, nx_padded*ny_padded/sz]
+      allocator%cdims_padded = [nx_padded, ny_padded, nz_padded]
    end function allocator_init
 
    function create_block(self, next) result(ptr)
@@ -169,13 +169,13 @@ contains
       ! Set dims based on direction
       select case(dir)
       case (DIR_X)
-         dims = self%xdims
+         dims = self%xdims_padded
       case (DIR_Y)
-         dims = self%ydims
+         dims = self%ydims_padded
       case (DIR_Z)
-         dims = self%zdims
+         dims = self%zdims_padded
       case (DIR_C)
-         dims = self%cdims
+         dims = self%cdims_padded
       case default
          error stop 'Undefined direction, allocator cannot provide a shape.'
       end select
--- a/src/allocator.f90
+++ b/src/allocator.f90
@@ -64,6 +64,7 @@ module m_allocator
       class(field_t), pointer :: next
       real(dp), pointer, private :: p_data(:)
       real(dp), pointer, contiguous :: data(:, :, :)
+      integer :: dir
       integer :: refcount = 0
       integer :: id !! An integer identifying the memory block.
    contains
@@ -165,6 +166,9 @@ contains
       else
          direction = DIR_X
       end if
+
+      ! Store direction info in the field type.
+      handle%dir = direction
 
       ! Set dims based on direction
       select case(direction)
--- a/src/allocator.f90
+++ b/src/allocator.f90
@@ -48,13 +48,13 @@ module m_allocator
    end type allocator_t
 
    type :: field_t
-     !! Memory block type holding both a 3D data field and a pointer
+     !! Memory block type holding both a data field and a pointer
      !! to the next block.  The `field_t` type also holds a integer
      !! `refcount` that counts the number of references to this
      !! field.  User code is currently responsible for incrementing
      !! the reference count.
       class(field_t), pointer :: next
-      real(dp), allocatable :: data(:, :, :)
+      real(dp), allocatable :: data(:)
       integer :: refcount = 0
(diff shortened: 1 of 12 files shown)


Nanoseb:
Yes, I think that can be neat. How I see it implemented is to have all the bound remapping handled by the `allocator%get_block` function. When ever we request an array we call it with an argument like "XYZ", "YZX" or "ZXY" for example and it will return a 3D array (pointing to the 1D one) that has the right dimensions. 

We should be able to call `u => allocator%get_block(XYZ)`. I will have to look into the details, but this may imply that `field_t` will have a 1D array as well as a 3D pointer to it.

pbartholomew08:
Another option, which could have some safety benefits for other areas of code would be to encode this in types, something like
```Fortran
type data_block
  real(dp), dimension(:), allocatable :: data
end type

type extends(data_block) :: x_block
end type
```

The user code would then look like
```Fortran
type(x_block) :: u_x
type(y_block) :: u_y
! ...

! Allocator allocates appropriate sizes based on type
call allocator%get_block(u_x)
call allocator%get_block(u_y)
```

This would be a wide-ranging change. But it would also allow us to reject invalid rotations at compile time
```Fortran
call rotate_xy(u_x, u_z) ! Can't compile due to types
```

semi-h:
We extend the base allocator in CUDA backend so that we can have a device array member. I don't think we can do extends you recommend and have x directional cuda field and base field under the same class. And then it will be hard to use allocator in solver class.

Maybe we can have a brand new type that has base field as a member, then extend this so that there are one for each direction
```
type :: dir_field_t
  class(field_t) :: f
end type

type extends(dir_field_t) :: xdir_field_t
end type
```

But then we'll have a select type for figuring out the direction, then another for resolving down to exact field type and this will make things complicated I think.

semi-h:
I think I figured out a nice way to fix this problem. A summary of changes:

We used to have a `data` and `data_d` component in `field_t` directly defined as an allocatable array.
Now this is only a pointer, points to a `private` `p_data` member, which is a pointer itself. 
- I would make `p_data` an allocatable array, but to point and allocatable array it has to be a `target`, and a member in a derived type can't be a `target`, so its a pointer instead.

Allocated memory is addessed by this private data member, and then to play with different shapes we use `data` pointer. This makes it easy to bring this fix into the codebase, because we already used to refer the field arrays with the `data` member in `field_t`. Also added a `private` attribute to the `p_data`, so that it can't be accessed outside the class. We should always indicate direction when requesting a block (`get_block`), and then `allocator/field` jointly responsible from passing us back a shape we want.

We can discuss this in our meeting today, so please have a look if you have time.

The only remaning part is actually setting the padding in the `allocator_init`, and then allowing users to pass an optional orientation argument to the `get_block` function in allocator. Bounds remapping will be carried out based on the requested orientation, and the rest of the codebase should work fine with this.

slaizet:
Discussion to have during our next meeting: 
we can decide or not to constrain the mesh resolution. Basically to have nx,ny and nz as multiple of SZ. It will definitely not work if SZ is typically 256/512 but could work if SZ is 8/16/32/64/128. Note that we have similar (however more flexible) constraints for the FFTs (nx, ny and nz need to be a combination of power of 2 & prime numbers).

semi-h:
I think we covered most of the technical parts in the meeting, if there is any point not clear I'll be happy to explain.

I just removed the padding in the z-direction as the current pencil group ordering strategy doesn't require any padding in z.

And it looks like there are a lot of changes but most of it is just `get_block` passing a direction preference.

A review will be helpful if you have time, please let me know if you want to review but can't do today so we can wait, but otherwise I think its in good shape and can be merged at the end of the day.

src/allocator.f90
```diff
@@ -69,16 +83,47 @@ module m_allocator
...
+      ! Current reorder functions do not require a padding in z-direction.
+      nz_padded = nz
+
+      allocator%ngrid = nx_padded*ny_padded*nz_padded
+      allocator%sz = sz
+
+      allocator%xdims = [sz, nx_padded, ny_padded*nz_padded/sz]
+      allocator%ydims = [sz, ny_padded, nx_padded*nz_padded/sz]
+      allocator%zdims = [sz, nz_padded, nx_padded*ny_padded/sz]
+      allocator%cdims = [nx_padded, ny_padded, nz_padded]
```
Nanoseb, line 125: I would mention in the names `xdims`/`ydims`/`zdims` that these are the padded dimensions. Also related to #77, at some point we should maybe think about where these are available, and how they are accessed.

src/tdsops.f90
```diff
@@ -46,7 +46,7 @@ module m_tdsops
                                       der2nd, der2nd_sym, &
                                       stagder_v2p, stagder_p2v, &
                                       interpl_v2p, interpl_p2v
-      integer :: nrank, nproc, pnext, pprev, n, n_blocks
+      integer :: nrank, nproc, pnext, pprev, n, n_blocks, dir
```
Nanoseb, line 49: I think `dir` can be removed again here.

Nanoseb (commented):
Beside these minor comments, this looks good to me. I think it should be merged before #68 and I will adapt it there to make it work (specially the tests).

src/allocator.f90
```diff
@@ -115,6 +162,33 @@ function get_block(self) result(handle)
       handle => self%first
       self%first => self%first%next ! 2nd block becomes head block
       handle%next => null() ! Detach ex-head block from the block list
+
+      ! If no direction is specified assume DIR_X
+      if (present(dir)) then
+         direction = dir
+      else
+         direction = DIR_X
+      end if
```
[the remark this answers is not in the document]
semi-h: Removed the default direction. Now we're required to pass a direction each time we request a block. Default was only being used in the time integrator anyways, all the rest was passing a direction. Hopefully now its more obvious when working with time integrator that we're always in `DIR_X` when doing a time integration. (some tests also required passing the DIR, just a side node)

src/tdsops.f90
semi-h: Actually we make use of this. When we're in transeq_component, we don't really know what direction we work at, but we need to request a temporary block. `dir` emmber of the `dirps_t` helps us for requesting the right direction.
Also, now that we have a `dir` info in `dirps_t`, we can do a sanity check in `tds_solve` for example. Last commit added a simple sanity check, which actually helped me fixing a minor bug!

Nanoseb: ah great, I didn't check in details but just assumed it wasn't needed because that was the only change in this file.

PR #37, opened 2024-02-14
Reduce the memory usage

semi-h:
I can't believe I missed this! 

In transeq at the very end we do
`du_x = du_x + du_y + du_z`

However due to the restictions on running GPU kernels with specific thread and block dimensions we carry out this operation in 2 separate calls as
`du_x = du_x + du_y`
`du_x = du_x + du_z`

And this gives an oppurtunity to move the y2x sum up just below `transeq_y` call. Then we release `du_y, dv_y, dw_y` right after adding these into _x counterparts.

This basic fix allowed **reducing the memory usage from 18 scalar fields down to 15**, without affecting the performance at all for the CUDA backend. (15GiB for a $512^3$ simulation). The total figure excludes Poisson solvers memory requirement which is not yet in the codebase. 

Any further reductions in memory usage after this point would result in an increase in the runtime of the simulation. For example we might be able to reduce it down to 12, which shouldn't be that hard, but it would require some extra reordering operations and its better not to work on that at this stage I think.

Assuming that FFT based Poisson solver will requre ~4 scalar field equivalent memory, we should be able to fit a $1024^3$ simulation on a typical 4xA100 node!

Now we have separate `sum_yintox`, `sum_zintox`, and `vecadd` subroutines in backends, all similar at some level. All these can be combined into a single subroutine like we did with reorder subroutines, and that's what I'll do next. I'll create an issue to discuss this further, and not include this next step in the current PR. I'm happy to merge this one as soon as someone approves.

### DIFF
--- a/src/backend.f90
+++ b/src/backend.f90
@@ -30,7 +30,6 @@ module m_base_backend
       procedure(transeq_ders), deferred :: transeq_z
       procedure(tds_solve), deferred :: tds_solve
       procedure(reorder), deferred :: reorder
-      procedure(sum_yzintox), deferred :: sum_yzintox
       procedure(sum_intox), deferred :: sum_yintox
       procedure(sum_intox), deferred :: sum_zintox
       procedure(vecadd), deferred :: vecadd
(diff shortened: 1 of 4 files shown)


src/backend.f90
```diff
@@ -95,7 +96,7 @@ end subroutine reorder
    end interface
 
    abstract interface
-      subroutine sum_yzintox(self, u, u_y, u_z)
+      subroutine sum_intox(self, u, u_)
```
JamieJQuinn, line 99: I presume this is the start of the refactor mentioned in #38 ?

semi-h: Yes, once we decide how to do it I'll implement in a separate PR. I just had to separate y and z here so that memory reduction can be implemented.

JamieJQuinn (approved):

PR #45, opened 2024-02-20
Add curl operator in solver

semi-h:
I added two new vector operators

- ~~divergence_v2v: required for checking the divergence of velocity field~~
- curl_v2v: required for enstrophy

Also renamed original divergence -> divergence_v2p and gradient -> gradient_p2v to clarify that these are switching from vertex grid to pressure grid.

Had a quick chat with @slaizet and he pointed out that we actually check the divergence of velocity field at the pressure grid and divergence of velocity at velocity grid wouldn't be zero. So removing divergence_v2v as we don't see any use case for it.

### DIFF
--- a/src/solver.f90
+++ b/src/solver.f90
@@ -45,7 +45,7 @@ module m_solver
    contains
       procedure :: transeq
       procedure :: divergence_v2p
-      procedure :: gradient
+      procedure :: gradient_p2v
       procedure :: run
    end type solver_t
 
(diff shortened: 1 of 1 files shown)


semi-h:
I dropped `_v2v` from the name of the curl subroutine. I realised that we have `transeq`, `der1st`, `der2nd` and so on all start on velocity grid and end on velocity grid. I think it is okay to omit `_v2v` everywhere, because its the primary grid and making it default makes sense. If we switch from `p2v`, `v2p`, or work on pressure grid directly (`p2p`), it is important to highlight this in the name.

PR #49, opened 2024-02-21
Run tests in parallel on CI

### LINKED ISSUE #20: Run tests in parallel in github actions
At the moment ctest only runs in serial.

~~See https://discourse.cmake.org/t/ctest-and-mpi-parallel-googletests/5557 to have them run with MPI.~~

Nanoseb:
closes #20

### DIFF
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -47,5 +47,5 @@ foreach(testfile IN LISTS TESTSRC)
   find_package(OpenMP REQUIRED)
   target_link_libraries(${test_name} PRIVATE OpenMP::OpenMP_Fortran)
 
-  add_test(NAME ${test_name} COMMAND sh -c "mpirun -np 4 ${test_name}")
+  add_test(NAME ${test_name} COMMAND sh -c "mpirun --oversubscribe -np 4 ${test_name}")
 endforeach()
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,5 +1,5 @@
 
-set(CMAKE_CTEST_ARGUMENTS "--output-on-failure")
+set(CMAKE_CTEST_ARGUMENTS "--output-on-failure" CACHE STRING "comma separated list of arguments to pass to ctest")
 
 set(TESTSRC
   test_allocator.f90
--- a/tests/CMakeLists.txt
+++ b/tests/CMakeLists.txt
@@ -1,3 +1,6 @@
+
+set(CMAKE_CTEST_ARGUMENTS "--output-on-failure")
+
 set(TESTSRC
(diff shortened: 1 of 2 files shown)


Nanoseb:
I had to use `--oversubscribe` because the CI runner doesn't seem to have more than 1 core. Though it shouldn't matter for the logic of the tests themselves.

tests/CMakeLists.txt
```diff
@@ -43,5 +47,5 @@ foreach(testfile IN LISTS TESTSRC)
   find_package(OpenMP REQUIRED)
   target_link_libraries(${test_name} PRIVATE OpenMP::OpenMP_Fortran)
 
-  add_test(NAME ${test_name} COMMAND sh -c "mpirun -np 1 ${test_name}")
+  add_test(NAME ${test_name} COMMAND sh -c "mpirun --oversubscribe -np ${CMAKE_CTEST_NPROCS} ${test_name}")
```
semi-h, line 50: Allocator and reorder tests are the ones that doesn't require a parallel run. Both of them don't do any communication and even in a full simulation with multiple ranks they're always local to their own rank. Running these on multiple ranks is still a valid test but if we can avoid this easily it is better I guess.

Nanoseb: that's a good point, though I can't think of a clean way of doing it yet.

pbartholomew08: You could use a function to build the test definition, I'll see if I can dig out an example where I've done this

pbartholomew08: Something like the following should work
```
function(define_test test_name np)
  add_test(NAME ${test_name} COMMAND sh -c "mpirun --oversubscribe -np ${np} ...")
endfunction()

# Make a single process test (assumes there is a test_foo target)
define_test(test_foo 1)

# Make a parallel test
define_test(test_bar ${CMAKE_CTEST_NPROCS})
```

tests/CMakeLists.txt
pbartholomew08, line 1: I think we are missing the CUDA link options with this file.

However, if we make the link option `PUBLIC` on the library definition then I think it should be inherited by anything which links against it.

pbartholomew08 (commented):
Minor comment, otherwise looks good to me.

tests/CMakeLists.txt
Nanoseb: not sure what you mean by "making the link option public on the lib definition" so re-added the -cuda flag here.

pbartholomew08: Hmm, not sure why but removing those from the test definitions failed to compile. I think best to keep them and everything else looks good to me

pbartholomew08 (approved):

PR #65, opened 2024-02-26
Bug fixes

semi-h:
Fixed some minor bugs in the solver level.

### DIFF
--- a/src/common.f90
+++ b/src/common.f90
@@ -5,7 +5,7 @@ module m_common
    real(dp), parameter :: pi = 4*atan(1.0_dp)
 
    integer, parameter :: RDR_X2Y = 12, RDR_X2Z = 13, RDR_Y2X = 21, &
-                         RDR_Y2Z = 23, RDR_Z2Y = 32
+                         RDR_Y2Z = 23, RDR_Z2X = 31, RDR_Z2Y = 32
 
    type :: globs_t
       integer :: nx, ny, nz
--- a/src/solver.f90
+++ b/src/solver.f90
@@ -137,6 +137,10 @@ contains
    end subroutine
 
    subroutine transeq(self, du, dv, dw, u, v, w)
+      !! Skew-symmetric form of convection-diffusion terms in the
+      !! incompressible Navier-Stokes momemtum equations, excluding
+      !! pressure terms.
+      !! Inputs from velocity grid and outputs to velocity grid.
       implicit none
 
       class(solver_t) :: self
@@ -218,6 +222,8 @@ contains
    end subroutine transeq
 
    subroutine divergence_v2p(self, div_u, u, v, w)
+      !! Divergence of a vector field (u, v, w).
+      !! Inputs from velocity grid and outputs to pressure grid.
       implicit none
 
       class(solver_t) :: self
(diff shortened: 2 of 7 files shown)


src/solver.f90
```diff
@@ -397,56 +398,83 @@ subroutine curl(self, o_x, o_y, o_z, u, v, w)
       class(field_t), intent(inout) :: o_x, o_y, o_z !! omega_x/_y/_z
```
semi-h, line 398: Confusing thing here is we normaly use `_x/y/z` to indicate the orientation of the array, but curl gives us back a field with three component by definition, and `o_x/y/z` here indicate three components of this vector. If it sounds better to you I can rename as `_i/j/k` so that it indicates the individual components of a vector field.

src/common.f90
```diff
@@ -5,7 +5,7 @@ module m_common
    real(dp), parameter :: pi = 4*atan(1.0_dp)
 
    integer, parameter :: RDR_X2Y = 12, RDR_X2Z = 13, RDR_Y2X = 21, &
-                         RDR_Y2Z = 23, RDR_Z2Y = 32
+                         RDR_Y2Z = 23, RDR_Z2X = 31, RDR_Z2Y = 32
```
pbartholomew08, line 8: This implies Z->X is now required, does the reordering issue need updating?

src/solver.f90
slaizet: _Xcomp / _Ycomp / _Zcomp?

pbartholomew08: I think go for the `_i/j/k` or even a combination of both to indicate the index and the orientation, e.g.
```
oi_x
oj_x
ok_x
```
or whichever orientation is appropriate

semi-h: How about `_xhat, _yhat, _zhat`? `comp` sounds like the shorthand for compact

slaizet: Xpart / Ypart / Zpart?

semi-h: I tend to omit `_x` for x oriented arrays in most places because its the default. Its like we omit `_v2v` if the operator works on velocity grid.

src/common.f90
semi-h: Done

semi-h:
I just renamed with `_i_hat`, `_j_hat`, and `_k_hat` as its a common vector notation and least confusing option considering we use `_x/y/z` for array orientations. Let me know if you need any other change otherwise it is ready to be merged one someone approves.

src/solver.f90
```diff
@@ -390,63 +391,90 @@ subroutine gradient_p2v(self, dpdx, dpdy, dpdz, pressure)
 
    end subroutine gradient_p2v
 
-   subroutine curl(self, o_x, o_y, o_z, u, v, w)
+   subroutine curl(self, o_i_hat, o_j_hat, o_k_hat, u, v, w)
```
JamieJQuinn, line 394: Can you add a comment here explaining the notation?

JamieJQuinn (changes requested):
Looks great. Can you just add some comments to the `curl` sub?

JamieJQuinn (approved):

PR #67, opened 2024-03-01
Add support for scalar product operator

semi-h:
We need a scalar product operator for calculating the enstrophy in the domain. The CUDA kernel can be implemented in a more performant way, there are lots of examples out there, but the current implementation is not bad. We only need to run this once in maybe 100/1000 iterations for post-processing so not worried too much.

### DIFF
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -41,6 +41,7 @@ module m_cuda_backend
       procedure :: sum_yintox => sum_yintox_cuda
       procedure :: sum_zintox => sum_zintox_cuda
       procedure :: vecadd => vecadd_cuda
+      procedure :: scalar_product => scalar_product_cuda
       procedure :: set_fields => set_fields_cuda
       procedure :: get_fields => get_fields_cuda
       procedure :: transeq_cuda_dist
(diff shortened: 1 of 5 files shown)


semi-h:
We'll probably have padding in our data structure when the grid size not a multiple of SZ and then if we use a library to calculate the dot product it'll be hard to eliminate the padded entries.

src/cuda/backend.f90
```diff
@@ -529,6 +531,35 @@ subroutine vecadd_cuda(self, a, x, b, y)
...
+      sum_d = 0._dp
+
+      n = size(x_d, dim = 2)
+      blocks = dim3(size(x_d, dim = 3), 1, 1)
+      threads = dim3(SZ, 1, 1)
+      call scalar_product<<<blocks, threads>>>(sum_d, x_d, y_d, n)
+
+      s = sum_d
+
+      call MPI_Allreduce(MPI_IN_PLACE, s, 1, MPI_DOUBLE_PRECISION, MPI_SUM, &
```
JamieJQuinn, line 558: Since we're thinking about mixed/reduced precision, should the precision passed to MPI derive from the actual data type? Not sure what the typical way to do this is in MPI codes. @pbartholomew08, @Nanoseb?

src/omp/backend.f90
```diff
@@ -329,31 +340,27 @@ subroutine copy_into_buffers(u_send_s, u_send_e, u, n, n_blocks)
 
    end subroutine copy_into_buffers
 
-   subroutine set_fields_omp(self, u, v, w, u_in, v_in, w_in)
+   subroutine set_field_omp(self, f, arr)
```
JamieJQuinn, line 343: Unclear argument names. Perhaps `field` and `data`?

src/cuda/kernels/reorder.f90
```diff
@@ -199,6 +199,27 @@ attributes(global) subroutine axpby(n, alpha, x, beta, y)
 
    end subroutine axpby
 
+   attributes(global) subroutine scalar_product(s, x, y, n)
```
JamieJQuinn, line 202: Caller needs to understand this is pencil sum to set correct n_threads/blocks. This should be documented here.

JamieJQuinn (changes requested):
Looks solid, just needs some documentation.

PR #68, opened 2024-03-01
Reorder for OMP backend

### LINKED ISSUE #42: Reordering operations
Need reordering steps from:

- x -> y
  - [x] cuda
  - [ ] omp
- x -> z
  - [x] cuda
  - [ ] omp
- y -> x
  - [x] cuda
  - [ ] omp
- y -> z
  - [x] cuda
  - [ ] omp
- z -> x
  - [ ] cuda
  - [ ] omp
- z -> y
  - [x] cuda
  - [ ] omp

~~Note: z -> x is not required as of 19/02/2024.~~

Curl operator requires z -> x, 27/02/2024.

### LINKED ISSUE #64: Implement local to global and global to local index mapping
With the backend-dependent layout we should have a method(s) to obtain the local index in (i,j,k) and from that the global index (or vice-versa). This will enable users who need the Cartesian coordinates to get these without duplicate code.

Nanoseb:
closes #42 
closes #64


Nanoseb:
This is ready for review. I am not 100% happy with having to send the block sizes around like that, but I have opened a separate issue to deal with it #77. Also I am not 100% sure about having this `select case` inside tight loops. I would expect compilers to notice the check is always constant and not bother checking it. But if it becomes a problem I will have to write separate functions for each direction (which I tried to avoid here).

pbartholomew08 (commented):
I've not checked the maths as not a good idea at end of the week, a few code/style comments

src/ordering.f90
```diff
@@ -0,0 +1,93 @@
...
+            dir_k = int(nx_loc/SZ)*(k-1) + ceiling(1.0*i/SZ) 
+         case (dir_Z)
+            dir_i = mod(i-1, SZ) + 1
+            dir_j = k
+            dir_k = int(nx_loc/SZ)*(j-1) + ceiling(1.0*i/SZ) 
+      end select
+
+   end subroutine get_index_dir
+
+   ! Swap between two sets of directional indices following reorder_dir
```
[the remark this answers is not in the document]
Nanoseb: It is directional to directional. But indeed the comment isn't clear

src/ordering.f90
```diff
@@ -0,0 +1,93 @@
...
+      integer, intent(out) :: i, j, k
+      integer, intent(in) :: dir_i, dir_j, dir_k
+      integer, intent(in) :: dir
+      integer, intent(in) :: SZ, nx_loc, ny_loc, nz_loc
+      
+      select case (dir)
+         case (dir_X)
+            i = dir_j
+            j = mod(dir_k -1, ny_loc/SZ)*SZ + dir_i 
+            k = ceiling(1.0*dir_k/(ny_loc/SZ))
```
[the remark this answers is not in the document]
Nanoseb: It is both `ceiling` needs it, but also we want to be sure we are not using integer divisions.

Nanoseb:
Also note that the tests written here are meant to also run on the `cuda` backend but I haven't tried them because I don't have the hardware at hand.

tests/CMakeLists.txt
```diff
@@ -8,9 +8,10 @@ function(define_test testfile np backend)
   add_executable(${test_name} ${testfile})
 
   if(${backend} STREQUAL cuda)
-    target_compile_options(${test_name} PRIVATE "-cuda")
+    target_compile_options(${test_name} PRIVATE "-cpp -cuda")
```
semi-h, line 11: What is the cpp flag is for?

tests/test_reordering.f90
```diff
@@ -0,0 +1,180 @@
+program test_reorder
```
semi-h, line 1: Having a high level test for the reorders is great, I think we can remove the low level tests for reorders on CUDA once we're sure this works for CUDA too. I think it is not yet ready to run the CUDA kernels though, mainly because of the `data_d` component in `cuda_field_t`. I see that you don't have the hardware to test so I'll be happy to make those minor changes later on and get this working with CUDA kernels as well.

tests/CMakeLists.txt
```diff
@@ -21,12 +22,14 @@ function(define_test testfile np backend)
...
 
 define_test(test_allocator.f90 1 omp)
+define_test(test_reordering.f90 1 omp)
 define_test(omp/test_omp_tridiag.f90 ${CMAKE_CTEST_NPROCS} omp)
 define_test(omp/test_omp_transeq.f90 ${CMAKE_CTEST_NPROCS} omp)
 define_test(omp/test_omp_dist_transeq.f90 ${CMAKE_CTEST_NPROCS} omp)
 
 if(${CMAKE_Fortran_COMPILER_ID} STREQUAL "PGI" OR
     ${CMAKE_Fortran_COMPILER_ID} STREQUAL "NVHPC")
+  define_test(test_reordering.f90 1 cuda)
```
semi-h, line 32: Please see my comment on the test program. I think defining the test like this with the current setup only repeats the test with the CPU kernels. We need to pass a `-DCUDA` flag to the compiler so that the correct classes are instantiated. But because the test program is not ready as of now passing the `-DCUDA` flag will result in errors. Once the test program is able to test the CUDA backend we can have this line back so that with NVIDIA compiler we test both CPU and GPU kernels.

src/omp/backend.f90
```diff
@@ -291,6 +293,40 @@ subroutine reorder_omp(self, u_, u, direction)
...
+            ndir_loc = self%zdirps%n
+            ndir_groups = self%zdirps%n_blocks
+      end select
+
+      do k=1, ndir_groups
+         do j=1, ndir_loc
+            do i=1, SZ
+               call get_index_reordering(out_i, out_j, out_k, i, j, k, direction, &
+                                         SZ, self%xdirps%n, self%ydirps%n, self%zdirps%n)
+               u_%data(out_i, out_j, out_k) = u%data(i,j,k)
```
semi-h, line 326: We needed the mappings between Cartesian and application storage so its great we have them. And using them is the quickest way to have reorders ready. My only concern is the performance of the data copies here. A `$omp parallel do` at the top loop will definitely make things better (maybe with a `collapse(2)` statement as well), and probably is the most we can do without writing specific functions. I think we'll need to check their performance at some point and decide what to do.

src/ordering.f90
```diff
@@ -0,0 +1,98 @@
...
+                       RDR_X2Y, RDR_X2Z, RDR_Y2X, RDR_Y2Z, RDR_Z2X, RDR_Z2Y
+
+   implicit none
+contains
+   !!
+   !! "Application storage" stores spatial data with a directionality for better cache locality
+   !!  This set of functions converts indices from this application storage (_dir) to cartesian indices (_ijk)
+   !! 
+
+   ! Get cartesian index from application storage directional one
```
semi-h, line 13: Moving this comment just below the subroutine decleration and starting with double `!!` triggers ford to create documentation automatically for us. See https://xcompact3d.github.io/x3d2/developer/style.html#in-code-docs

src/ordering.f90
```diff
@@ -0,0 +1,98 @@
...
+      integer, intent(out) :: i, j, k                   ! cartesian indices
+      integer, intent(in) :: dir_i, dir_j, dir_k        ! application storage indices
+      integer, intent(in) :: dir                        ! direction of the applicatino storage indices
+      integer, intent(in) :: SZ, nx_loc, ny_loc, nz_loc ! dimensions of the block
+      
+      select case (dir)
+         case (dir_X)
+            i = dir_j
+            j = mod(dir_k - 1, ny_loc/SZ)*SZ + dir_i 
+            k = ceiling(real(dir_k)/(ny_loc/SZ))
```
semi-h, line 24: I was thinking if there is any way to avoid switching to reals. Because both the inputs and outputs are integers, it can be safer to keep everything as integers in the intermediate operations as well. For example maybe this can work here `1 + (dir_k - 1)/(ny_loc/SZ)`

tests/test_reordering.f90
```diff
@@ -0,0 +1,180 @@
...
+   use m_omp_backend, only: omp_backend_t
+#endif
+
+    implicit none
+
+    logical :: allpass = .true.
+    class(field_t), pointer :: u_x, u_y, u_z, u_x_original
+
+    integer :: nrank, nproc
+    integer :: ierr, i, j, k, i_, j_, k_, i__, j__, k__
```
pbartholomew08, line 29: Could we use something more descriptive than `i_/i__` to represent the different index types, or atleast do
```
integer :: i_, j_, k_       ! Computed directional indices
integer :: i__, j__, k__ ! Computed Cartesian indices
```

tests/test_reordering.f90
```diff
@@ -0,0 +1,180 @@
...
+                end if
+                call get_index_dir(i_, j_, k_, i, j, k, dir_Z, SZ, xdirps%n, xdirps%n, xdirps%n)
+                call get_index_ijk(i__, j__, k__, i_, j_, k_, dir_Z, SZ, xdirps%n, xdirps%n, xdirps%n)
+                if (i /= i__ .or. j /= j__ .or. k/= k__) then
+                    print *, "Error in dir Z"
+                    allpass = .false.
+                end if
+            end do
+        end do
+    end do
```
pbartholomew08, line 127: Extract as a utility subroutine so you can easily pass `xdirps/ydirps/zdirs` and test across all 3, or does this not make sense to do?

pbartholomew08 (commented):
All looks good. As @semi-h has mentioned extending the tests to cover CUDA I will leave approving for now.

The tests I think are sensible in that they check the reordering is reversible. However to ensure that they are indeed working, I would suggest
```
1) reorder a field from Cartesian into Directional/vectorised layout
2) use the Fortran intrinsic to reshape this into a 1D layout
3) check that neighbouring values in the Cartesian field are separated by SZ steps in the vectorised layout
```

tests/test_reordering.f90
Nanoseb: ah yes, sure. Feel free to edit this PR to make it work there

src/ordering.f90
Nanoseb: Good point, I will look into it

Nanoseb:
> All looks good. As @semi-h has mentioned extending the tests to cover CUDA I will leave approving for now.
> 
> The tests I think are sensible in that they check the reordering is reversible. However to ensure that they are indeed working, I would suggest
> 
> ```
> 1) reorder a field from Cartesian into Directional/vectorised layout
> 2) use the Fortran intrinsic to reshape this into a 1D layout
> 3) check that neighbouring values in the Cartesian field are separated by SZ steps in the vectorised layout
> ```

Actually, thinking about it, it will work for the particular data layout we are using now, but won't when we switch to not having `SZ` cells on a single line (which we are planning to do to reduce padding). So not sure it is worth it to create a test for it now. 

Reversibility is the easiest to tests these things, but indeed it is flawed if you make the same mistake twice. Something I could do is have more convoluted ways to go back to the original field, that way it is less prone to errors cancelling each other out. I also can test if all the indices are 'touched' by the mapping once and only once by each function. I believe all of this should cover most edge cases that could arise.

tests/test_reordering.f90
Nanoseb: Yeah, definitely

src/omp/backend.f90
Nanoseb: Agreed, for the `parallel do` I forgot it. Though not sure what gain `collapse(2)` could bring as the loop is perfectly balanced in terms of work in each iteration and the thread count should be small compared to the number of iterations. But it doesn't hurt in putting it there.

Nanoseb: added the `parallel do`, and I also created an issue to track this #78

tests/test_reordering.f90
semi-h: I think it is all sorted now. For now I haven't removed the low level CUDA tests in cuda folder because they include performance tests as well. Once we have a mechanism for checking performance in the high level code low level reorder tests in cuda folder can be removed completely. Checking the performance is very helpful for me to maintain a good performance, so I prefer not losing this ability.

tests/test_reordering.f90
```diff
@@ -0,0 +1,225 @@
...
+    end subroutine
+
+    subroutine check_reorder(allpass, a, b, message)
+        logical, intent(inout) :: allpass
+        class(field_t), intent(in) :: a, b
+        character(len=*), intent(in) :: message
+        real(dp) :: tol = 1d-8
+
+#ifdef CUDA
+        select type (a); type is (cuda_field_t); temp_1 = a%data_d; end select
```
Nanoseb, line 208: @semi-h Just a thought not related to this issue but would it make sense to have temp_1 stored inside the a structure. We could call something like a%copy_to_host(), that would run this line and we should be able to use a%data just like on the omp backend?
Same we could have a%copy_to_device() to do the opposite.
We would still need the ifdef but it will simplify/clarify code quite a bit

tests/test_reordering.f90
semi-h: The only thing is in the test program there is a mix of 3/4 spacing, it would be nice to have it consistent with the main codebase. Apart from this I think the PR is in good shape, please merge after taking a look at the changes I pushed.

Nanoseb: Yes, sure. Once we have a dedicated place to put performance tests we can clean up the duplicate unit tests.

tests/test_reordering.f90
semi-h: For doing that I think we would need to allocate the `%data` component, which currently we keep as a private member to discourage people accessing directly this component. Maybe we can use the `get_field` in the backends for making this cleaner, but I think I'll be making some changes in `get_field` with the cartesian / specialist data structure mapping so it might be better leaving this after `get/set_field` functions are more stable.

pbartholomew08 (approved):

pbartholomew08:
> > All looks good. As @semi-h has mentioned extending the tests to cover CUDA I will leave approving for now.
> > The tests I think are sensible in that they check the reordering is reversible. However to ensure that they are indeed working, I would suggest
> > ```
> > 1) reorder a field from Cartesian into Directional/vectorised layout
> > 2) use the Fortran intrinsic to reshape this into a 1D layout
> > 3) check that neighbouring values in the Cartesian field are separated by SZ steps in the vectorised layout
> > ```
> 
> Actually, thinking about it, it will work for the particular data layout we are using now, but won't when we switch to not having `SZ` cells on a single line (which we are planning to do to reduce padding). So not sure it is worth it to create a test for it now.
> 
> Reversibility is the easiest to tests these things, but indeed it is flawed if you make the same mistake twice. Something I could do is have more convoluted ways to go back to the original field, that way it is less prone to errors cancelling each other out. I also can test if all the indices are 'touched' by the mapping once and only once by each function. I believe all of this should cover most edge cases that could arise.

OK, I thought the data structure was quite frozen, maybe we should stick with the reordering as is for now - it is a good test as you say. Future commits can always extend the tests. Ideally the test we want shouldn't rely directly on the data structure but this is easier said than done

PR #76, opened 2024-03-14
Add reorder kernels for switching from/to Cartesian ordering

semi-h:
The idea is we interact with outside world always with Cartesian data order.

Backends require a special data structure, but they are capable of converting the specialist data structure into Cartesian and vice versa, so we can and should always input from and output to Cartesian ordered arrays.
We plan to carry out this operation with the `get_field`/`set_field` subroutines in both backends, but as of now they can't carry out the conversion to and from Cartesian yet. Becase in order to switch to and from Cartesian we need to be able to request a Cartesian shape from the allocator, (#34), and the plan is to merge the PR enables this on Monday after finalising some important decisions.

Although this PR doesn't enables the actual functionality we need, it provides all the necessary kernels. I think the best would be merging this before #34, and making the final minor changes in `get_field`/`set_field` in #34.

### DIFF
--- a/src/backend.f90
+++ b/src/backend.f90
@@ -206,12 +206,14 @@ module m_base_backend
 contains
 
    subroutine get_field_data(self, data, f, dir)
+   !! Extract data from field `f` optionally reordering into `dir` orientation.
+   !! To output in same orientation as `f`, use `call ...%get_field_data(data, f, f%dir)`
       implicit none
 
       class(base_backend_t) :: self
-      real(dp), dimension(:, :, :), intent(out) :: data
-      class(field_t), intent(in) :: f
-      integer, optional, intent(in) :: dir
+      real(dp), dimension(:, :, :), intent(out) :: data !! Output array
+      class(field_t), intent(in) :: f !! Field
+      integer, optional, intent(in) :: dir !! Desired orientation of output array (defaults to Cartesian)
 
       class(field_t), pointer :: f_temp
       integer :: direction, rdr_dir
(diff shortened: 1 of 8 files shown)


semi-h:
I think the PR in its final shape, and ready for a review. `get_field_data` and `set_field_data` functions are now implemented at base backend level, and capable of passing back a 3D array with any directionality we can optionally provide. All use cases in the codebase as of now input and output a Cartesian 3D array, so I made Cartesian ordering default if no directionality is specified. However, one can easily allocate a x-directional Fortran array and request `get_field_data` to output into x-directional 3D array. @JamieJQuinn, please let me know if the PR provides the functionality you need for the IO.

src/backend.f90
```diff
@@ -202,4 +203,70 @@ subroutine init_poisson_fft(self, xdirps, ydirps, zdirps)
...
+         call self%copy_f_to_data(data, f)
+      end if
+
+   end subroutine get_field_data
+
+   subroutine set_field_data(self, f, data, dir)
+      implicit none
+
+      class(base_backend_t) :: self
+      class(field_t), intent(inout) :: f
```
JamieJQuinn, line 244:
```suggestion
      class(field_t), intent(inout) :: f !! Field
```

src/backend.f90
```diff
@@ -202,4 +203,70 @@ subroutine init_poisson_fft(self, xdirps, ydirps, zdirps)
...
+      real(dp), dimension(:, :, :), intent(in) :: data
```
JamieJQuinn, line 245:
```suggestion
      real(dp), dimension(:, :, :), intent(in) :: data !! Input array
```

src/backend.f90
```diff
@@ -202,4 +203,70 @@ subroutine init_poisson_fft(self, xdirps, ydirps, zdirps)
...
+      integer, optional, intent(in) :: dir
```
JamieJQuinn, line 246:
```suggestion
      integer, optional, intent(in) :: dir !! Orientation of input array (defaults to Cartesian)
```

JamieJQuinn (approved):
Added some documentation but otherwise looks good! Cartesian remap should enable continued development on IO.

semi-h:
> Added some documentation but otherwise looks good! Cartesian remap should enable continued development on IO.

Thanks a lot!

PR #83, opened 2024-03-29
Use 3D FFT transforms directly and drop the current 1D decomposition capable strategy

### LINKED ISSUE #73: Possible cleanup needed in cuda complex reorder kernels
Currently, the subroutines in `src/cuda/kernels/complex.f90` are considered temporary and tied to the current implementation of the FFT Poisson solver using cuFFT. When this is reimplemented in cuFFTMP, these functions are likely to be removed.

*Only* if we decide to keep the existing reordering subroutines specific to cuFFT should we tidy those subroutines up and document properly. This issue is just a reminder and can be closed if the relevant subroutines are removed as planned.

semi-h:
We're moving away from doing our own FFTs using our specialist data structure to carry out 3D FFT transforms. The strategy CUDA backend implements prior to this PR was capable of being extended to 1D decomposition (already done earlier for the Thomas based approach), and as it was limited to 1D it was possible to take advantage of this for some optimisations. However, the distributed tridiagonal solver we have necessitates each slab/pencil to have certain amount of thickness, which limits a 1D decomposition strategy for going above 2k^3 ~ 4k^3 domain sizes.

This important shift in strategy was already mentioned in #32, and also #73.

The changes implemented up to so far includes switching from specialist data structure to cartesian data structure so that distributed FFT libraries can be used easily. 

~~And currently, there is a bug in CUDA backend in `poisson_fft` that prevents running with different `nx`, `ny`, `nz` sizes. Possibly due to 3D FFT transform halving a dimension that is different than the current implementation assumes. Should be fixed soon as its not a big deal.~~ Fixed.

Closes #73.


src/cuda/kernels/complex.f90
JamieJQuinn, line 1: Probably a good time to start documenting this function. At the very least I think it needs a description of arguments and a reference for the calculation being carried out. I presume this will need @slaizet's input?

JamieJQuinn: Also, could this file be renamed to something more descriptive? Suggestions:

- `spectral_transform.f90`
- `spectral_processing.f90`
- `poisson_solver.f90`

src/cuda/poisson_fft.f90
```diff
@@ -71,39 +69,21 @@ function init(xdirps, ydirps, zdirps) result(poisson_fft)
...
-    ierrfft = cufftSetWorkArea(poisson_fft%planZ2Zx, poisson_fft%c_y_dev)
+    allocate (poisson_fft%c_w_dev(nx/2 + 1, ny, nz))
+    allocate (poisson_fft%fft_worksize((nx/2 + 1)*ny*nz))
+
+    ! 3D plans
+    ierr = cufftCreate(poisson_fft%plan3D_fw)
+    ierr = cufftMakePlan3D(poisson_fft%plan3D_fw, nz, ny, nx, CUFFT_D2Z, &
+                           worksize)
+    ierr = cufftSetWorkArea(poisson_fft%plan3D_fw, poisson_fft%fft_worksize)
+    if (ierr /= 0) error stop 'Forward 3D FFT plan generation failed'
```
JamieJQuinn, line 80:
```suggestion
    if (ierr /= 0) then
      write(stderr, *), "CuFFT Error Code: ", ierr
      error stop 'Forward 3D FFT plan generation failed'
    end if
```

Just to add more debug info when necessary.

src/cuda/poisson_fft.f90
```diff
@@ -71,39 +69,21 @@ function init(xdirps, ydirps, zdirps) result(poisson_fft)
...
+
+    ierr = cufftCreate(poisson_fft%plan3D_bw)
+    ierr = cufftMakePlan3D(poisson_fft%plan3D_bw, nz, ny, nx, CUFFT_Z2D, &
+                           worksize)
+    ierr = cufftSetWorkArea(poisson_fft%plan3D_bw, poisson_fft%fft_worksize)
+    if (ierr /= 0) error stop 'Backward 3D FFT plan generation failed'
```
JamieJQuinn, line 86:
```suggestion
    if (ierr /= 0) then
      write(stderr, *), "CuFFT Error Code: ", ierr
      error stop 'Backward 3D FFT plan generation failed'
    end if
```

src/cuda/poisson_fft.f90
```diff
@@ -114,47 +94,21 @@ subroutine fft_forward_cuda(self, f)
...
+    ! Using f_dev directly in cufft call causes a segfault
+    ! Pointer switches below fixes the problem
+    f_c_ptr = c_loc(f_dev)
+    call c_f_pointer(f_c_ptr, f_ptr)
 
-    ! In-place forward FFT in x
-    ierrfft = cufftExecZ2Z(self%planZ2Zx, self%c_x_dev, self%c_x_dev, &
-                           CUFFT_FORWARD)
+    ierr = cufftExecD2Z(self%plan3D_fw, f_ptr, self%c_w_dev)
+    if (ierr /= 0) error stop 'Forward 3D FFT execution failed'
```
JamieJQuinn, line 111:
```suggestion
    if (ierr /= 0) then
      write(stderr, *), "CuFFT Error Code: ", ierr
      error stop 'Forward 3D FFT execution failed'
    end if
```

src/cuda/poisson_fft.f90
```diff
@@ -165,47 +119,21 @@ subroutine fft_backward_cuda(self, f)
...
-
-    ! Backward FFT transform in z from complex to real
-    ierrfft = cufftExecZ2D(self%planZ2Dz, self%c_z_dev, self%f_tmp)
-
-    ! Finally reorder f back into our specialist data structure
-    blocks = dim3(self%nz/SZ, self%nx*self%ny/SZ, 1)
-    threads = dim3(SZ, SZ, 1)
-    call reshapeDSB<<<blocks, threads>>>(f_dev, self%f_tmp) !&
+    ierr = cufftExecZ2D(self%plan3D_bw, self%c_w_dev, f_ptr)
+    if (ierr /= 0) error stop 'Backward 3D FFT execution failed'
```
JamieJQuinn, line 136:
```suggestion
    if (ierr /= 0) then
      write(stderr, *), "CuFFT Error Code: ", ierr
      error stop 'Backward 3D FFT execution failed'
    end if
```

src/cuda/poisson_fft.f90
```diff
@@ -214,32 +142,23 @@ subroutine fft_postprocess_cuda(self)
...
+    complex(dp), device, dimension(:, :, :), pointer :: c_dev
     type(dim3) :: blocks, threads
+    integer :: tsize
 
-    ! Get some complex array pointers with right shape
-    c_x_ptr(1:self%nx, 1:SZ, 1:(self%ny*(self%nz/2 + 1))/SZ) => self%c_x_dev
-    c_dev(1:SZ, 1:self%nx, 1:(self%ny*(self%nz/2 + 1))/SZ) => self%c_y_dev
+    ! tsize is different than SZ, because here we work on a 3D Cartesian
+    ! data structure, and free to specify any suitable thread/block size.
+    tsize = 16
```
JamieJQuinn, line 151: If this might be tuned later, can it be stored more publicly? E.g. as a parameter in `common.f90`. I'm worried this parameter will get forgotten later.

JamieJQuinn (changes requested):
Looks grand! I added a few suggestions for more informative error handling, and some organisational requests but I think this is ready to merge after those are done.

src/cuda/kernels/complex.f90
semi-h: Added a description for a few arguments, and a reference too if someone wants more details. I need @slaizet's help with the remaining ones.

With the complex reorder functions gone, in this module we'll only carry out postprocessing in spectral space, so renaming is a good idea. I went with `spectral_processing.f90`.

src/cuda/poisson_fft.f90
semi-h: At the moment the only use case for this variable is here in the postprocessing, and for this specific kernel. GPU kernels are generally written considering the block and thread sizes, so I think its a good idea to define it here and use right after. For example, if we need to process a 3D Cartesian array its likely that we'll require a different thread size. The variable here works well because the kernel we have requires 3 separate 1D arrays with global size to carry out the operation, probably we won't need a similar operation anywhere else. If we observe use of this elsewhere then maybe we can move somewhere else later on.

PR #84, opened 2024-04-05
Global-local grid indexing

semi-h:
Implements support for inidivudal ranks in a 3D decomposition setting to figure out their location in the global grid.

### DIFF
--- a/src/tdsops.f90
+++ b/src/tdsops.f90
@@ -44,7 +44,7 @@ module m_tdsops
     !! in each coordinate direction.
     class(tdsops_t), allocatable :: der1st, der1st_sym, der2nd, der2nd_sym, &
       stagder_v2p, stagder_p2v, interpl_v2p, interpl_p2v
-    integer :: nrank, nproc_dir, pnext, pprev, n, n_blocks, dir
+    integer :: nrank_dir, nproc_dir, pnext, pprev, n, n_blocks, dir
     integer :: n_offset = 0 ! offset w.r.t. global domain at the starting point
     real(dp) :: L, d
   end type dirps_t
--- a/src/tdsops.f90
+++ b/src/tdsops.f90
@@ -38,15 +38,14 @@ module m_tdsops
   end interface tdsops_t
 
   type :: dirps_t
-      !! Directional tridiagonal solver container.
-      !!
-      !! This class contains the preprocessed tridiagonal solvers for operating
-      !! in each coordinate direction.
-    class(tdsops_t), allocatable :: der1st, der1st_sym, &
-      der2nd, der2nd_sym, &
-      stagder_v2p, stagder_p2v, &
-      interpl_v2p, interpl_p2v
+    !! Directional tridiagonal solver container.
+    !!
+    !! This class contains the preprocessed tridiagonal solvers for operating
+    !! in each coordinate direction.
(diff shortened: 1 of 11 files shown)


src/tdsops.f90
```diff
@@ -38,15 +38,14 @@ module m_tdsops
   end interface tdsops_t
 
   type :: dirps_t
```
semi-h, line 40: We may want to move `dirps_t` to `domain.f90` file instead, what do you think?

src/tdsops.f90
```diff
@@ -38,15 +38,14 @@ module m_tdsops
...
-      der2nd, der2nd_sym, &
-      stagder_v2p, stagder_p2v, &
-      interpl_v2p, interpl_p2v
+    !! Directional tridiagonal solver container.
+    !!
+    !! This class contains the preprocessed tridiagonal solvers for operating
+    !! in each coordinate direction.
+    class(tdsops_t), allocatable :: der1st, der1st_sym, der2nd, der2nd_sym, &
+      stagder_v2p, stagder_p2v, interpl_v2p, interpl_p2v
     integer :: nrank, nproc, pnext, pprev, n, n_blocks, dir
```
semi-h, line 47: The `nrank` and `nproc` here meant to be the directional `nrank` and directional `nproc`, but I see that it might be confusing. Maybe its better renaming them `nrank_dir`, `nproc_dir`?

tests/omp/test_omp_transeq.f90
```diff
@@ -50,15 +51,8 @@ program test_omp_transeq
...
-  zdirps%nproc = nproc
-
-  call set_pprev_pnext( &
-    xdirps%pprev, xdirps%pnext, &
-    ydirps%pprev, ydirps%pnext, &
-    zdirps%pprev, zdirps%pnext, &
-    xdirps%nproc, ydirps%nproc, zdirps%nproc, nrank &
-    )
+  ydirps%nproc = 1
+  zdirps%nproc = 1
```
semi-h, line 55: For example, I see that the global `nproc` obtained from MPI is assigned to all local `nproc`s here, but actually `nproc = xdirps%nproc*ydirps%nproc*zdirps%nproc`.

semi-h (commented):
Just wanted to point out a few places so that everyone can take a quick look

src/tdsops.f90
pbartholomew08: I think the suggested change is good

tests/omp/test_omp_transeq.f90
pbartholomew08: Would this be better done by `init_dirps` and `init_globs`? I've outlined these in the PETSc branch, based on this code, but we could cherry pick and ensure there's a single correct implementation

semi-h: Yes, I think its better. Right now we repeat this in the tests but should move all this to an initialisation function ideally. #85 is probably a better place to work this out though, its already simplifying relevant bits.

semi-h:
@JamieJQuinn, `domain_decomposition` subroutine now figures out the `n_offset` parameter and it is stored in `dirps_t`. Can be accessed via `x/y/zdirps%n_offset`. Would this work for you in the io?

src/tdsops.f90
semi-h: I think better to leave this for now because #85 will make some changes here.

PR #85, opened 2024-04-09
Move dimensions to allocator

### LINKED ISSUE #77: Have SZ, nx_loc, n_groups etc. accessible "globally"
At the moment all the variables that define block sizes are defined in various places and sometimes duplicated over multiple structures. For example, `backend_t` has `nx_loc`, `ny_loc`, `ny_loc`, but so has `backend_t%xdirps%n` etc. 
`SZ` is accessed in `m_omp_common` or `m_cuda_common` depending on the backend, meaning we can't access it outside the backend functions (unless being passed as argument from there etc.). 

It would be good to have a single consistent way of accessing all of these variable. It will also reduce the number of arguments needed to be passed around.

Nanoseb:
closes #77

### DIFF
--- a/tests/omp/test_omp_dist_transeq.f90
+++ b/tests/omp/test_omp_dist_transeq.f90
@@ -115,7 +115,7 @@ program test_transeq
     d2u_send_s, d2u_send_e, d2u_recv_s, d2u_recv_e, &
     u, u_recv_s, u_recv_e, &
     v, v_recv_s, v_recv_e, &
-    der1st, der1st, der2nd, nu, nproc, pprev, pnext, n_block &
+    der1st, der1st, der2nd, nu, nproc, pprev, pnext, n_block, sz &
     )
 
   ! check error
--- a/tests/omp/test_omp_transeq.f90
+++ b/tests/omp/test_omp_transeq.f90
@@ -60,23 +60,15 @@ program test_omp_transeq
     xdirps%nproc, ydirps%nproc, zdirps%nproc, nrank &
     )
 
-  xdirps%n = globs%nx_loc
-  ydirps%n = globs%ny_loc
-  zdirps%n = globs%nz_loc
-
-  xdirps%n_blocks = globs%n_groups_x
-  ydirps%n_blocks = globs%n_groups_y
-  zdirps%n_blocks = globs%n_groups_z
-
   xdirps%dir = DIR_X
   ydirps%dir = DIR_Y
   zdirps%dir = DIR_Z
 
-  omp_allocator = allocator_t(xdirps%n, ydirps%n, zdirps%n, SZ)
+  omp_allocator = allocator_t(globs%nx_loc, globs%ny_loc, globs%nz_loc, SZ)
   allocator => omp_allocator
   print *, 'OpenMP allocator instantiated'
 
-  omp_backend = omp_backend_t(globs, allocator)
+  omp_backend = omp_backend_t(allocator)
   backend => omp_backend
   print *, 'OpenMP backend instantiated'
 
@@ -84,7 +76,7 @@ program test_omp_transeq
 
   n_glob = globs%nx
   n = n_glob/nproc
-  n_block = xdirps%n_blocks
+  n_block = globs%n_groups_x
 
   nu = 1._dp
   omp_backend%nu = nu
--- a/tests/omp/test_omp_tridiag.f90
+++ b/tests/omp/test_omp_tridiag.f90
@@ -299,7 +299,7 @@ contains
 
       call exec_dist_tds_compact(du, u, u_recv_s, u_recv_e, &
                                  send_s, send_e, recv_s, recv_e, &
-                                 tdsops, nproc, pprev, pnext, n_block)
+                                 tdsops, nproc, pprev, pnext, n_block, sz)
 
     end do
   end subroutine run_kernel
--- a/tests/test_reordering.f90
+++ b/tests/test_reordering.f90
@@ -32,6 +32,7 @@ program test_reorder
   real(dp), allocatable, dimension(:, :, :) :: u_array, temp_1, temp_2
 
   integer :: dims(3)
+  integer :: cdims(3)
 
   integer :: nrank, nproc
   integer :: ierr, i, j, k
@@ -72,25 +73,13 @@ program test_reorder
   globs%ny_loc = globs%ny/nproc
   globs%nz_loc = globs%nz/nproc
 
-  globs%n_groups_x = globs%ny_loc*globs%nz_loc/SZ
-  globs%n_groups_y = globs%nx_loc*globs%nz_loc/SZ
-  globs%n_groups_z = globs%nx_loc*globs%ny_loc/SZ
-
-  xdirps%n = globs%nx_loc
-  ydirps%n = globs%ny_loc
-  zdirps%n = globs%nz_loc
-
-  xdirps%n_blocks = globs%n_groups_x
-  ydirps%n_blocks = globs%n_groups_y
-  zdirps%n_blocks = globs%n_groups_z
-
 #ifdef CUDA
   cuda_allocator = cuda_allocator_t(globs%nx_loc, globs%ny_loc, globs%nz_loc, &
                                     SZ)
   allocator => cuda_allocator
   print *, 'CUDA allocator instantiated'
 
-  cuda_backend = cuda_backend_t(globs, allocator)
+  cuda_backend = cuda_backend_t(allocator)
   backend => cuda_backend
   print *, 'CUDA backend instantiated'
 #else
@@ -98,7 +87,7 @@ program test_reorder
   allocator => omp_allocator
   print *, 'OpenMP allocator instantiated'
 
-  omp_backend = omp_backend_t(globs, allocator)
+  omp_backend = omp_backend_t(allocator)
   backend => omp_backend
   print *, 'OpenMP backend instantiated'
 #endif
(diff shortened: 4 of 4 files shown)


semi-h:
The main change here is introducing new member variables and removing some of them, involving `dirps_t`, `field_t`, and `allocator_t` classes. Some of these are great and I completely agree, but there are a few I have some concerns. 
- `cdims` in `allocator_t` :heavy_check_mark:
  - A new member stores the actual sizes of the subdomain, I think very helpful. It also helped getting rid of `globs` in certain places which is great.
- `n_groups_x/y/z` in `allocator_t` :heavy_check_mark:
  - We manage the padding in the allocator and the number of groups in the domain directly depends on this, its a good idea to store this info in the allocator.
- Add `n` in `field_t` and remove it from `dirps_t`
  - I think this is the most important one, mainly due to the difference between velocity `n` and pressure `n`. I'll leave a separate review on this so that I can point out specific examples and suggest an alternative solution. I think going ahead with this will make managing the difference between velocity/pressure grid harder later on.
- Add `sz` in `field_t`
  - If we want to avoid using the `SZ` from `m_omp_common`, then I think we can make `sz` a member of the `backend_t`. Because once the backend is instantiated, `SZ` is fixed and it'll never change. Having this as a member in `field_t` is a bit misleading in my opinion, it implies as if fields might have a different `SZ` value, but its never the case. There are a few locations where passing and using `sz` as a variable instead of using it as a parameter will have performance issues too, I'll comment on these separately.
- Add `n_groups` in `field_t` and remove the corresponding member from `dirps_t`
  - As far as I can see this only replaced accessing `n_groups` from `dirps_t` with `field_t`. Also, it didn't really allowed us to avoid passing `dirps_t` in any of the subroutines we access this member, so I'm not sure if this simplifies anything. In my view its better to store `n_groups` in `dirps_t` because it doesn't really depend on the field instance, field instance is just a scratch space after all, but it completely depends on the direction we're working at a given time.

There is also a new member `n_padded` in `field_t` but its not used anywhere yet, so not commenting on it.

src/omp/exec_dist.f90
```diff
@@ -2,7 +2,6 @@ module m_omp_exec_dist
   use mpi
 
   use m_common, only: dp
-  use m_omp_common, only: SZ
```
semi-h, line 5: Not being able to access `SZ` as a parameter will probably make performance worse, because compiler won't be certain whether the `simd` loops can be vectorised or not at compile time.

src/omp/exec_dist.f90
```diff
@@ -127,15 +128,15 @@ subroutine exec_dist_transeq_compact( &
       ! Handle dud by locally generating u*v
       do j = 1, n
         !$omp simd
-        do i = 1, SZ
+        do i = 1, sz
```
semi-h, line 131: For example here we basically tell compiler to do its best to vectorise the loop here. But if the loop here goes from 1 to 7 lets say, its impossible for compiler to vectorise this. And if `sz` here is a variable not a parameter, compiler can't be sure at compile time that this is a good integer for vectorisation. In the best case scenario it'll generate multiple versions of assembly, and will have to decide which one to execute only at runtime, which is not ideal. It may alltogether refuse to generate vectorised assembly, and this will depend on the compiler. So here I recommend using `SZ` as a parameter directly from `m_omp_common`.

src/omp/exec_dist.f90
```diff
@@ -93,22 +93,23 @@ subroutine exec_dist_transeq_compact( &
...
     n_halo = 4
     n = tdsops_d2u%n
-    n_data = SZ*n_block
+    n_data = sz*n_groups
 
-    allocate (ud(SZ, n))
-    allocate (ud_recv_e(SZ, n_halo))
-    allocate (ud_recv_s(SZ, n_halo))
+    allocate (ud(sz, n))
+    allocate (ud_recv_e(sz, n_halo))
```
semi-h, line 108: Another minor example, if we have both `sz` and `n_halo` as a parameter, it'll probably be better for performance as the size of the allocation can be determined at the compile time. `n_halo` is not a parameter yet, but we can actually make it, its always a constant value in backends and there are places we hardcoded things based on `n_halo=4`.

src/omp/exec_dist.f90
```diff
@@ -177,7 +178,7 @@ subroutine exec_dist_transeq_compact( &
 
       do j = 1, n
         !$omp simd
-        do i = 1, SZ
+        do i = 1, sz
```
semi-h, line 181: Just another example where a parameter `SZ` will make performance better.

semi-h (commented):
Here I focused entirely on `SZ`.

src/omp/backend.f90
```diff
@@ -393,23 +362,21 @@ real(dp) function scalar_product_omp(self, x, y) result(s)
...
-    do k = 1, n_blocks
+    do k = 1, u%n_groups
       do j = 1, n_halo
         !$omp simd
-        do i = 1, SZ
-          u_send_s(i, j, k) = u(i, j, k)
-          u_send_e(i, j, k) = u(i, n - n_halo + j, k)
+        do i = 1, u%sz
+          u_send_s(i, j, k) = u%data(i, j, k)
+          u_send_e(i, j, k) = u%data(i, u%n - n_halo + j, k)
```
semi-h, line 379: In particular here, we can't have `field_t%n`. We have to pass an `n` value into this subroutine which is determined from the corresponding `tdsops`, when `copy_into_buffers` is called in `tds_solve`. `tds_solve` always have a `tdsops`, and passing the `tdsops%n` down to `copy_into_buffers` is necessary. Otherwise we'll end up sending wrong set of slices.

src/allocator.f90
```diff
@@ -170,12 +180,28 @@ function get_block(self, dir) result(handle)
...
+      handle%n = self%cdims(3)
+      handle%SZ = self%zdims_padded(1)
+      handle%n_padded = self%zdims_padded(2)
+      handle%n_groups = self%zdims_padded(3)
     case (DIR_C)
       dims = self%cdims_padded
+      handle%n = -1
+      handle%SZ = -1
+      handle%n_padded = -1
+      handle%n_groups = -1
```
semi-h, line 204: I think this also suggests that we shouldn't store these variables in `field_t`. `field_t` is effectively a scratch space and sometimes these variables don't have a meaning. These assignments in particular can cause annoying bugs.

semi-h (commented):
I'm starting a new review specifically for discussing `field_t%n`.

I tried to avoid inferring the actual size the tridiagonal solver will make use from the domain shape. This is mainly because of the difference between velocity grid and pressure grid.

I'll try to give an example:
We call tds_solve in various settings and often we pass a velocity grid data and get back pressure grid data such as
https://github.com/xcompact3d/x3d2/blob/ce2efbc240e21f4ac5e63b0d1e4a3a7f7d1d6f31/src/solver.f90#L284-L285

In this case, its very important to use the correct `n` in the `tds_solve`.

https://github.com/xcompact3d/x3d2/blob/ce2efbc240e21f4ac5e63b0d1e4a3a7f7d1d6f31/src/cuda/backend.f90#L369
The `tdsops` class we have is the key here. It has got an `n` member where the size of the tridiagonal system is defined. Its always the correct value, either `n` of velocity or `n` of pressure because it is literally the class that defines the operation. It is determined based on the operation is a `_p2v` or `_v2p` or so on. We pass this down to `exec_dist` via `tdsops_dev` in #L409 below.
https://github.com/xcompact3d/x3d2/blob/ce2efbc240e21f4ac5e63b0d1e4a3a7f7d1d6f31/src/cuda/backend.f90#L404-L411
And evantually we call the kernel with `tdsops%n` below in #L43.
https://github.com/xcompact3d/x3d2/blob/ce2efbc240e21f4ac5e63b0d1e4a3a7f7d1d6f31/src/cuda/exec_dist.f90#L40-L44

##  Halo exchange
The actual `n` of the field is also important in halo exchanges. Because if we can't determine the actual `n`, whether its pressure or velocity, we can't send the right last 4 slices to the neighbouring ranks.
https://github.com/xcompact3d/x3d2/blob/ce2efbc240e21f4ac5e63b0d1e4a3a7f7d1d6f31/src/cuda/backend.f90#L395-L396
That's why we need to pass an `n` into the `copy_into_buffers`, so that the last 4 layers are accessed correctly, based on its pressure or velocity data. See below.
https://github.com/xcompact3d/x3d2/blob/ce2efbc240e21f4ac5e63b0d1e4a3a7f7d1d6f31/src/omp/backend.f90#L412
Because of this we need to pass an `n` value to the `copy_into_buffers`, and can't assume its `field_t%n`.

I think having a `field_t%n` is not a good idea because of all these. It'll be misleading and the `n` in a given direction can be stored in `dirps_t%n`. When its safe to do so we can use it, or when necessary we'll have to access `tdsops_t%n`.

src/omp/backend.f90
```diff
@@ -309,43 +305,16 @@ subroutine reorder_omp(self, u_, u, direction)
...
-      error stop 'unsuported reordering'
-    end select
-
     !$omp parallel do private(out_i, out_j, out_k) collapse(2)
-    do k = 1, ndir_groups
-      do j = 1, ndir_loc
-        do i = 1, SZ
+    do k = 1, u%n_groups
+      do j = 1, u%n
+        do i = 1, u%sz
```
semi-h, line 314: To simplify the `reorder_omp` how about introducing a variable `n_groups(3)` in the `allocator_t` class and;
```
do k = 1, self%allocator%n_groups(u%dir)
  do j = 1, self%allocator%cdims(u%dir)
    do i = 1, self%sz (which is from backend%sz or SZ directly)
```
It doesn't require us passing anything extra and removes all the select case stuff above.

JamieJQuinn:
Seems like there are a few different definitions of `n` used here:

1. Global n{x,y,z} - entire simulation domain across all nodes
2. Local n{x,y,z} - local simulation domain unique to one node
3. Padded local n{x,y,z} - actual size of array used to store local domain
4. Pressure/velocity n{x,y,z} - either n or n-1 depending on staggeredness of variable

So questions to quell my quonfusion:

1. What is the value intended to be stored in `field_t%n` and `field_t%n_padded` @Nanoseb? The actual size of the field, including staggeredness?
2. What should `tdsops%n` contain @semi-h? It looks like it's *always* local, pressure grid `n{x,y,z}`. So we can't naively derive this from `field_t%n` because `field_t%n` can either be pressure or velocity size? Would having `field_t%n` help with error handling, e.g. if someone passes in the wrong staggeredness of array?
3. Unsure why `field_t%n` wouldn't contain the right `n` for the halo exchange @semi-h. Can you elaborate?

FWIW it feels like having this information about field within field is sensible but I agree with @semi-h  that it can't be misleading. I wonder if there's a better naming scheme we can use to distinguish between the above kinds of `nx` we have floating around.

Nanoseb:
I think there are two different questions here, where are these variable stored and how they are called. 
The main purpose of this PR was to put everything in the fields. Most of the times when we act on fields we want to loop through them, so having all the sizes we would need at hand was the main drive for this work. That's the reason why, even if it leads to some duplication of quantities (SZ will be the same for every field for example), I still think it makes sense to have it accessible via the field object.

Concerning the naming conventions, I agree this needs to be clarified/improved. One of the reason it isn't really clear is because `get_block` doesn't distinguish between velocity (n) and pressure (n+1) fields. Something we could do is, just like we do with the direction, to ask for either a cell based block or a vertex based one when requesting `get_block`. Requesting either will set the right value in `n`.
That way, we can always assume that `n` is the size we need either in a pressure or velocity array.

As we mentioned in the related issue, the plan was to remove the `_loc` from `n` and add `_global` whenever we talk about global sizes. So any `n` here is local.

PR #88, opened 2024-05-13
Add Adams-Bashforth time-integration schemes

CFD-Xing:
This PR implements Adams-Bashforth integration schemes with provision for the start-up phase by successively increasing the time-integration order.

The Dahlquist equation is used to check Adams-Bashforth integration schemes from order 1 to 4. New tests have been added in the following file
`tests/omp/test_omp_adamsbashforth.f90`

### DIFF
--- a/src/omp/backend.f90
+++ b/src/omp/backend.f90
@@ -383,6 +383,8 @@ contains
     real(dp), intent(in) :: b
     class(field_t), intent(inout) :: y
 
+    y%data = a*x%data + b*y%data ! fixme
+
   end subroutine vecadd_omp
 
   real(dp) function scalar_product_omp(self, x, y) result(s)
--- a/src/time_integrator.f90
+++ b/src/time_integrator.f90
@@ -89,14 +89,13 @@ contains
     self%deriv(2)%ptr => dv
     self%deriv(3)%ptr => dw
 
-    order = min(self%istep, self%order)
-    call self%adams_bashforth(order, dt)
+    call self%adams_bashforth(dt)
 
     ! increment step counter
     self%istep = self%istep + 1
   end subroutine step
 
-  subroutine adams_bashforth(self, order, dt)
+  subroutine adams_bashforth(self, dt)
     class(time_intg_t), intent(inout) :: self
     integer, intent(in) :: order
     real(dp), intent(in) :: dt
@@ -104,6 +103,7 @@ contains
     integer :: i, j
     class(field_t), pointer :: ptr
 
+    order = min(self%istep, self%order)
     do i = 1, self%nvars
       ! update solution
       call self%backend%vecadd(self%coeffs(1, order)*dt, &
@@ -116,23 +116,23 @@ contains
       end do
 
       ! rotate pointers
-      if (self%istep == order .and. self%order > order) then
+      if (self%order > order) then
         ! for startup
-        if (self%istep == 1) then
-          ! update olds(1) with new derivative
-          call self%backend%vecadd(1.0_dp, self%deriv(i)%ptr, 0._dp, &
-                                   self%olds(i, 1)%ptr)
-        else
+        if (self%istep > 1) then
           call rotate(self%olds(i, :), order)
         end if
-      else if (self%order > 2) then
+      else 
         ! after startup
-        call rotate(self%olds(i, :), order - 1)
+        if (self%order > 2) then
+          call rotate(self%olds(i, :), order - 1)
+        end if
       end if
 
       ! update olds(1) with new derivative
-      call self%backend%vecadd(1.0_dp, self%deriv(i)%ptr, 0._dp, &
-                               self%olds(i, 1)%ptr)
+      if (self%order > 1) then
+        call self%backend%vecadd(1.0_dp, self%deriv(i)%ptr, 0._dp, &
+                                 self%olds(i, 1)%ptr)
+      end if
     end do
 
   end subroutine adams_bashforth
--- a/src/time_integrator.f90
+++ b/src/time_integrator.f90
@@ -26,12 +26,13 @@ module m_time_integrator
 
 contains
 
-  function init(backend, allocator, nvars)
+  function init(backend, allocator, order, nvars)
     implicit none
 
     type(time_intg_t) :: init
     class(base_backend_t), pointer :: backend
     class(allocator_t), pointer :: allocator
+    integer, intent(in), optional :: order
     integer, intent(in), optional :: nvars
 
     integer :: i, j
@@ -46,6 +47,12 @@ contains
 
     init%backend => backend
     init%allocator => allocator
+
+    if (present(order)) then
+      init%order = order
+    else
+      init%order = 1
+    end if
 
     if (present(nvars)) then
       init%nvars = nvars
@@ -101,7 +108,6 @@ contains
 
     integer :: i, j
     integer :: order
-    class(field_t), pointer :: ptr
 
     order = min(self%istep, self%order)
     do i = 1, self%nvars
--- a/src/time_integrator.f90
+++ b/src/time_integrator.f90
@@ -8,7 +8,7 @@ module m_time_integrator
   private adams_bashforth
 
   type :: time_intg_t
-    integer :: istep, nsteps, nsubsteps, order, nvars, nolds
+    integer :: istep, order, nvars, nolds
     real(dp) :: coeffs(4, 4)
     type(flist_t), allocatable :: olds(:, :)
     type(flist_t), allocatable :: curr(:)
--- a/src/time_integrator.f90
+++ b/src/time_integrator.f90
@@ -152,7 +152,7 @@ contains
     ! rotate pointer
     ptr => sol(n)%ptr
     do i = n, 2, -1
-      sol(n)%ptr => sol(n - 1)%ptr
+      sol(i)%ptr => sol(i - 1)%ptr
     end do
     sol(1)%ptr => ptr
 
--- a/src/time_integrator.f90
+++ b/src/time_integrator.f90
@@ -134,9 +134,9 @@ contains
       call self%backend%vecadd(-0.5_dp*dt, self%olds(i, 1)%ptr, 1._dp, &
                                self%curr(i)%ptr)
 
-      ! for startup
+      ! rotate pointers
       if (self%istep == 1 .and. self%order > 2) then
-        ! rotate pointers
+        ! for startup
         call rotate(self%olds(i, :), 2)
       end if
 
@@ -163,13 +163,12 @@ contains
       call self%backend%vecadd(5._dp/12._dp*dt, self%olds(i, 2)%ptr, &
                                1._dp, self%curr(i)%ptr)
 
-      ! for startup
+      ! rotate pointers
       if (self%istep == 2 .and. self%order > 3) then
-        ! rotate pointers
+        ! for startup
         call rotate(self%olds(i, :), 3)
-        ! after startup
       else
-        ! rotate pointers
+        ! after startup
         call rotate(self%olds(i, :), 2)
       end if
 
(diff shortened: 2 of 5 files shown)


src/time_integrator.f90
```diff
@@ -8,10 +8,16 @@ module m_time_integrator
   type :: time_intg_t
     integer :: istep, nsteps, nsubsteps, order, nvars, nolds
     type(flist_t), allocatable :: olds(:, :)
+    type(flist_t), allocatable :: curr(:)
+    type(flist_t), allocatable :: deriv(:)
```
CFD-Xing, line 12: Using flist_t to generalize time-integration schemes to an arbitrary number of variables

src/time_integrator.f90
```diff
@@ -8,10 +8,16 @@ module m_time_integrator
...
     class(base_backend_t), pointer :: backend
     class(allocator_t), pointer :: allocator
   contains
     procedure :: step
+    procedure :: adams_bashforth_1st
+    procedure :: adams_bashforth_2nd
+    procedure :: adams_bashforth_3rd
+    procedure :: adams_bashforth_4th
```
CFD-Xing, line 20: Implement Adams-Bashforth methods of order 1 to 4.

src/time_integrator.f90
```diff
@@ -35,11 +41,13 @@ function constructor(backend, allocator, nvars)
...
       constructor%nvars = nvars
-    else
+    else 
       constructor%nvars = 3
     end if
 
-    constructor%nolds = 0
+    constructor%istep = 0
+    constructor%order = 1
+    constructor%nolds = constructor%order - 1
```
CFD-Xing, line 50: Initialize relevant parameters during construction

src/time_integrator.f90
```diff
@@ -55,23 +63,175 @@ end function constructor
...
+      end if
+
+    ! Fourth-order
+    else if (self%order .eq. 4) then
+      if (self%istep .eq. 0) then
+        call self%adams_bashforth_1st(dt)
+      else if (self%istep .eq. 1) then
+        call self%adams_bashforth_2nd(dt)
+      else if (self%istep .eq. 2) then
+        call self%adams_bashforth_3rd(dt)
```
CFD-Xing, line 111: Using lower-order approaches as startup, currently assumes constant time step dt

src/time_integrator.f90
```diff
@@ -55,23 +63,175 @@ end function constructor
...
+        self%olds(i,3)%ptr => self%olds(i,2)%ptr
+        self%olds(i,2)%ptr => self%olds(i,1)%ptr
+        self%olds(i,1)%ptr => ptr
+      ! after startup
+      else
+        ! roll pointers
+        ptr => self%olds(i,3)%ptr
+        self%olds(i,3)%ptr => self%olds(i,2)%ptr
+        self%olds(i,2)%ptr => self%olds(i,1)%ptr
+        self%olds(i,1)%ptr => ptr
```
CFD-Xing, line 228: Simply roll pointers to avoid unnecessary repeated copy. Only update the last step residuals.

src/time_integrator.f90
```diff
@@ -55,23 +63,175 @@ end function constructor
...
+      call self%backend%vecadd(3._dp / 8._dp * dt, self%olds(i,3)%ptr, 1._dp, self%curr(i)%ptr)
+      
+      ! for startup
+      if (self%istep .eq. 3 .and. self%order .gt. 4) then
+        ! roll pointers
+        ptr => self%olds(i,4)%ptr
+        self%olds(i,4)%ptr => self%olds(i,3)%ptr
+        self%olds(i,3)%ptr => self%olds(i,2)%ptr
+        self%olds(i,2)%ptr => self%olds(i,1)%ptr
+        self%olds(i,1)%ptr => ptr
```
CFD-Xing, line 221: This part of the code is used as start-up for an higher-order time-integration approach

src/time_integrator.f90
```diff
@@ -55,23 +63,175 @@ end function constructor
...
-    type(flist_t) :: vels(:), olds(:)
-    real :: coeffs(:)
+  subroutine adams_bashforth_1st(self, dt)
+    class(time_intg_t), intent(inout) :: self
+    real(dp), intent(in) :: dt
+
+    integer :: i
+
+    do i = 1, self%nvars
+      call self%backend%vecadd(dt, self%deriv(1)%ptr, 1._dp, self%curr(1)%ptr)
```
semi-h, line 130: I think we should use `i` here instead of `1` when accessing entries of `deriv` and `curr`.

src/time_integrator.f90
semi-h: Is it possible to move rolling pointers into a function so that we call this function each time we need to roll? I think it'll look more tidy like that.

semi-h (commented):
Looks great! You can test this simply by running the main program on GPUs and it will run a TGV case on a 256^3 mesh with Re=1600 from $t=0$ to $t=20t_c$. You'll get the enstrophy distribution printed on the terminal. If you want to use a different time step or a larger mesh you'll need to edit the corresponding parameters in the main program. Note that multi-GPU support is not yet in the mainline so you'll have to run on a single GPU for the time being.

There are a few style mismatches between the codebase and here, but you can use fprettify to sort them out for you. See https://xcompact3d.github.io/x3d2/developer/style.html.

@slaizet, do we want Runge-Kutta time steppers as well or happy with Adams-Bashforth for now?

src/time_integrator.f90
```diff
@@ -55,23 +63,175 @@ end function constructor
...
+    self%curr(1)%ptr => u
+    self%curr(2)%ptr => v
+    self%curr(3)%ptr => w
+
+    ! assign pointer to variables
+    self%deriv(1)%ptr => du
+    self%deriv(2)%ptr => dv
+    self%deriv(3)%ptr => dw
+
+    ! First-order
```
Nanoseb, line 82: This feels quite verbose and repetitive, can we not use something like:
```Fortran
order = min(self%istep +1, self%order)
select case (order)
  case (1)
    call self%adams_bashforth_1st(dt)
  case (2)  
    call self%adams_bashforth_2nd(dt)
    ...
  
```

Or even have the order be an argument of the `self%adams_bashforth()` function and have the logic there.

CFD-Xing:
> Looking good. It would be good to have tests for each time advancement scheme (maybe using a simple equation with an analytical solution, to check that they indeed converge to the expected order). RK schemes (at least 3rd order and possibly 4th order) would be good to have, possibly in another PR.

Implementing RK schemes is going to be more challenging and will require more significant changes in order to work with a segregated solver. The velocity update, Poisson solver, and velocity correction need to be solved in stages, this will require changes to both `solver.f90` and `time_integration.f90`. Basically, the whole time-integration process need to be wrapped into a function/subroutine.

semi-h:
@Nanoseb also mentioned earlier but it looks like there is a pattern across `adams_bashforth` 1 to 4 and this can be handled with a single subroutine if we pass the order parameter. The coefficients belong to different orders can be stored in an array (or array of pointers) and then a loop over them can sort the `vecadd` lines, and then in the part the pointers are rolled the if check is based on the order variable anyways so it looks quite straightforward to implement.

Also, Fortran indexing is 1 based so starting `istep` from 1 could be a better idea, any thoughts on this?

pbartholomew08:
> Also, Fortran indexing is 1 based so starting `istep` from 1 could be a better idea, any thoughts on this?

I could go either way - we aren't indexing into an array with `istep` (e.g. a record of all timesteps) and time starts at t=0, so in my opinion `istep=0` makes sense (also gives you `t = istep * dt`). However in Xcompact3d `istep` starts at 1 so from familiarity perspective it might be better to retain this.

slaizet:
I would prefer to start at 1 please :-)

src/time_integrator.f90
```diff
@@ -39,9 +51,13 @@ function constructor(backend, allocator, nvars)
       constructor%nvars = 3
     end if
 
-    constructor%nolds = 0
+    constructor%istep = 1
+    constructor%order = 1
+    constructor%nolds = constructor%order - 1
```
CFD-Xing, line 56: Any objection to rename `constructor` by `init`. This would seem to be more consistent with the rest of the code.

Nanoseb:
see failed test. Note that you can run the locally with `make test`.

src/omp/backend.f90
```diff
@@ -383,6 +383,8 @@ subroutine vecadd_omp(self, a, x, b, y)
     real(dp), intent(in) :: b
     class(field_t), intent(inout) :: y
 
+    y%data = a*x%data + b*y%data ! fixme
```
CFD-Xing, line 386: This is a temporary implementation necessary for the `omp` test

tests/omp/test_omp_adamsbashforth.f90
```diff
@@ -0,0 +1,168 @@
...
+
+      ! compute l2 norm for a given step size
+      allocate (err(nstep))
+      u%data(1, 1, 1) = 1.0_dp
+
+      ! startup
+      istartup = k - 1
+      do i = 1, istartup
+        du%data(1, 1, 1) = dahlquist_rhs(u%data(1, 1, 1))
+        call time_integrator%step(u, v, w, du, dv, dw, dt)
```
CFD-Xing, line 85: Currently, `v`, `w`, `dv`, and `dw` are placeholder variables, only `u` and `du` are actually used for the test. I think that the `step` API should be modified to accept a flist insead, so that it can be used for an arbitrary number of variables.

CFD-Xing:
enstrophy

This is the enstrophy figure obtained with this PR

src/omp/backend.f90
Nanoseb: maybe we should create an issue about it then

tests/omp/test_omp_adamsbashforth.f90
Nanoseb: Same, maybe create an issue about it. It will be lost if it stays in the PR only

src/time_integrator.f90
```diff
@@ -5,73 +5,158 @@ module m_time_integrator
...
+      else
+        ! after startup
+        if (self%order > 2) then
+          call rotate(self%olds(i, :), order - 1)
+        end if
+      end if
+
+      ! update olds(1) with new derivative
+      if (self%order > 1) then
+        call self%backend%vecadd(1.0_dp, self%deriv(i)%ptr, 0._dp, &
```
semi-h, line 140: We can potentially do a pointer switch here too I think. Not very important but saves a few `vecadd`s. If you decide to give this a go please run AB3 again with the TGV to make sure it doesn't break anything, and also check out the memory usage too. A mistake in pointer switch might lead to constant allocations at each time step.

CFD-Xing: I think I will do that in the next PR about Runge-Kutta. I am now thinking to add a swap routine.

CFD-Xing: One difference is the `self%deriv` pointers point to the input parameters `du, dv, dw`. And in `solver.f90` there is this function call to
`self%backend%allocator%release_block(du)`

Is this going to cause problem?

With `self%olds`, everything is internal to the time integrator type.

semi-h: Its completely fine provided that `du` points to a field. So if you do
```
ptr => olds(?)
olds(?) => du
du => ptr
```
I think it should work. Probably requires `du` to be `intent(inout)`.

semi-h: Ah, you work on them via the `flist_t` `deriv`, so it can be more tricky than the code snipped above.

CFD-Xing: Yeah, `flist_t olds` and  `flist_t deriv`  are going to be swaped, but `du` still point to the same memory location, which will be released just after...

PR #94, opened 2024-05-17
Add Runge-Kutta time-integration schemes 

CFD-Xing:
This PR implements Runge-Kutta schemes from order 1 to 4. 

The default time-stepping approach is set to AB3.

All schemes (AB1 to AB4 and RK1 to RK4) have been validated for the Dahlquist equation.

For the TGV case, this PR has been tested with RK3 with the following result

enstrophy

TODO: 
- Validate order of convergence for Navier-Stokes for RK1 to RK4 and AB1 to AB4.
- Generalize the time stepping API for arbitrary numbers of variable (currently only working for u, v, and w)


src/time_integrator.f90
```diff
@@ -5,73 +5,270 @@ module m_time_integrator
...
+    init%rk_a(:, 3, 3) = [0.0_dp, 0._dp, 0._dp]
+    init%rk_b(:, 3) = &
+      [2._dp/9.0_dp, 1._dp/3._dp, 4._dp/9._dp, 0._dp]
+
+    ! rk4
+    init%rk_a(:, 1, 4) = [0.5_dp, 0._dp, 0._dp]
+    init%rk_a(:, 2, 4) = [0._dp, 0.5_dp, 0._dp]
+    init%rk_a(:, 3, 4) = [0._dp, 0._dp, 1._dp]
+    init%rk_b(:, 4) = &
+      [1._dp/6._dp, 1._dp/3._dp, 1._dp/3._dp, 1._dp/6._dp]
```
CFD-Xing, line 68: Currently, general Butcher coefficient storage is used for the explicit RK implementation. If we restrict the RK schemes from order 1 to 4, we can simplify this by only storing the diagonal of `rk_a`.

src/time_integrator.f90
```diff
@@ -5,73 +5,270 @@ module m_time_integrator
...
+        ! update stage solution
+        if (self%istage > 1) then
+          call self%backend%vecadd(1.0_dp, self%olds(i, 1)%ptr, 0._dp, &
+                                   self%curr(i)%ptr)
+        end if
+        do j = 1, self%istage
+          call self%backend%vecadd(self%rk_a(j, self%istage, self%nstage)*dt, &
+                                   self%olds(i, j + 1)%ptr, &
+                                   1._dp, self%curr(i)%ptr)
+        end do
```
CFD-Xing, line 206: This loop can be eliminated if we decide to only consider "diagonal" explicit RK. For the current implementation, we are simply multiplying arrays by zero for `j < self%istage`.

src/time_integrator.f90
CFD-Xing: Also, several alternatives exist for RK2, RK3, and RK4. We should agree on which version(s) of RK we want to support (SSP, Low storage, etc.) for each order.

semi-h:
@CFD-Xing can you do a rebase please? It'll help for reviewing the RK related stuff.

src/time_integrator.f90
```diff
@@ -26,33 +30,101 @@ module m_time_integrator
 
 contains
 
-  function init(backend, allocator, order, nvars)
+  subroutine finalize(self)
```
CFD-Xing, line 33: Add a destructor here

tests/test_time_integrator.f90
```diff
@@ -185,13 +187,15 @@ program test_omp_adamsbashforth
...
     print *, 'order', order
-    if (abs(order - real(k, dp)) > 0.1_dp) then
+    if (abs(order - real(time_integrator%order, dp)) > 0.25_dp) then
       allpass = .false.
       write (stderr, '(a)') 'Check order... failed'
     else
       write (stderr, '(a)') 'Check order... passed'
     end if
 
+    call time_integrator%finalize
```
CFD-Xing, line 197: Deallocate time_integrator memory for each scheme

pbartholomew08: would this not be helpful as a code comment? after the PR is merged it will essentially "disappear" otherwise (and likewise for above comment).

tests/test_time_integrator.f90
```diff
@@ -174,7 +176,7 @@ program test_omp_adamsbashforth
       ! compute l2 norms
       norm(j) = norm2(err)
       norm(j) = sqrt(norm(j)*norm(j)/real(nstep, dp))
-      print *, err(nstep)
+      print *, norm(j)
```
CFD-Xing, line 179: Fix an error here, does not affect convergence results.

src/time_integrator.f90
```diff
@@ -5,18 +5,23 @@ module m_time_integrator
...
-  private adams_bashforth
+  private adams_bashforth, runge_kutta
 
   type :: time_intg_t
-    integer :: istep, order, nvars, nolds
+    integer :: method, istep, istage, order, nstep, nstage, nvars, nolds
     real(dp) :: coeffs(4, 4)
+    real(dp) :: rk_b(4, 4)
+    real(dp) :: rk_a(3, 3, 4)
+    character(len=3) :: sname
```
CFD-Xing, line 15: Add a member variable to store the method name, useful to keep track of the time-integration approach used, specially for time-convergence study

src/time_integrator.f90
```diff
@@ -26,33 +31,103 @@ module m_time_integrator
...
+    ! deallocate memory
+    deallocate (self%olds)
+    deallocate (self%curr)
+    deallocate (self%deriv)
+
+    print *, self%sname, ' time integrator deallocated'
+
+  end subroutine finalize
+
+  function init(backend, allocator, method, nvars)
```
semi-h, line 58: I think `method` argument which is an integer at the moment can be expanded into two arguments such that one argument indicates whether we want `runge-kutta` or `adams-bashforth`, and then another argument to indicate the integration order. The integrator can throw an error and stop the program if an unsupported method/order pair is requested.

PR #95, opened 2024-05-17
Add mesh object

### LINKED ISSUE #91: Naming and storage convention for grid information
**What**

Currently the code stores information about the grid dimensions in various locations, e.g. SZ in `common.f90`, local length of _something_ is stored in `xdirps%n`, number of blocks `xdirps%nblocks`. Padded versions of these are stored in `allocator%xdims_padded(3)`. A field stores its direction/orientation in `field_t%dir` but no size information. Accessing this all together at the point of use is difficult and requires bringing multiple objects together.

**Proposed solution**

Fields should store their local and padded size, and their grid type (vertex, cell, face, edge centred) which ensures that the sizes are correct for their use.

```
type :: field_t
  ...
  integer :: type
  integer :: n ! "length" of field
  integer :: ngroups ! Number of vector groups in field
  ! And padded versions ...
  ...
  contains
  ...
  ...
end type
```

The `get_block` subroutine will also take the field "type" in addition to the direction. This will allow it to set the local and padded dimensions according to the field's grid type and orientation. Suggested types are:
- `VERT` (velocity grid)
- `CELL` (pressure grid)
- `X_EDGE` x-oriented edge (interpolation / derivative to/from velocity/pressure along x)
- `Z_FACE` z-normal oriented face (interpolation / derivative to/from velocity/pressure along y)
- `Y_EDGE`, `Z_EDGE` - reserved but unused
- `X_FACE`, `Y_FACE` - reserved but unused
- `NULL` - no specified grid location - is this necessary???

A field's size will depend on where in the tridiagonal system its block is, boundary blocks will have different sizes from internal blocks, however after creation the field doesn't need to know its location, only its size.
The allocator will take `tdsops_t` (X/Y/Z) as arguments on initialisation, this will allow it to determine where in the grid it is, and use this information at field creation time.

**Summary**

- Fields should know their sizes, both local (true data) and padded (multiple of vector length)
- Allocator will set the field sizes based on their orientation and grid type (for use intent) and their location in the domain
- Access to field dimensions will be via getter/setter functions/subroutines. This will allow us to make the dimension fields `private`, preventing direct access to them and addressing concerns about them becoming invalid. In the `NULL` field case only the setter will allow changing these, otherwise it is an invalid call and causes an `error stop`.

Nanoseb:
This PR implements the ideas discussed in #91 

The current progress is the following:

- [x] mesh module
- [x] OMP code updated
- [x] CUDA code updated
- [x] tests updated

The code function `mesh%get_n` isn't finalised yet because I am a bit unsure of the actual result it should return depending on the periodicity of the BC and the `data_loc` (`VERT`, `CELL`, etc.)
@pbartholomew08 or @semi-h could you have a look at it?

closes #91


src/mesh.f90
```diff
@@ -0,0 +1,323 @@
...
+        n = n
+      case(Y_FACE)
+        n = n
+      case(Z_FACE)
+        n = n
+      case(X_EDGE)
+      case(Y_EDGE)
+      case(Z_EDGE)
+    end select
+  end function
```
Nanoseb, line 300: This is the core function that needs to be checked/fixed

src/tdsops.f90
```diff
@@ -169,7 +178,7 @@ subroutine deriv_1st(self, delta, scheme, bc_start, bc_end, sym)
 
     self%dist_sa(:) = alpha; self%dist_sc(:) = alpha
 
-    n = self%n
+    n = self%mesh%get_n(self%dir, VERT)
```
semi-h, line 181: Can't really assume its `VERT` here, this will lead to wrong results. Depends on the type of the operator. `p2v`/`v2p` etc. I think its best resolving this before an instance of `tdsops_t` is created.

src/tdsops.f90
```diff
@@ -84,8 +87,14 @@ function tdsops_init(n, delta, operation, scheme, n_halo, from_to, &
...
     integer :: n_stencil
+    integer :: n
+    real(dp) :: delta
+
+    tdsops%mesh = mesh
+    tdsops%dir = dir
+    n = tdsops%mesh%get_n(dir, data_loc)
+    delta = tdsops%mesh%geo%d(dir)
 
-    tdsops%n = n
```
semi-h, line 88: A very important variable. I actually think we need a `n_in` and `n_out`. This way it shouldn't be misleading as if its the `n` in `dirps` or somewhere else. Happy to explain this in detail in our meeting on Monday.

src/xcompact.f90
```diff
@@ -56,81 +60,52 @@ program xcompact
...
+  
+  ! Global domain dimensions
+  L_global = [2*pi, 2*pi, 2*pi]
 
-  ! read ns from the input file
-  globs%nx = 256; globs%ny = 256; globs%nz = 256
+  ! Domain decomposition in each direction
+  nproc_dir = [1, 1, 1]
+
+  mesh = mesh_t(dims_global, nproc_dir, L_global, SZ)
```
semi-h, line 72: This made the prep steps so easy, looks great now!

semi-h (commented):
I think `mesh` class made things a lot better. I left a lot of comments, mostly minor or easy to fix stuff. The most important point is the `n` variable in `tds_solve`. The suggested changes in the PR will probably cause some issues with tridiagonal solves, but the fix is easy and doesn't require any big structural changes.

src/mesh.f90
```diff
@@ -0,0 +1,325 @@
...
+    ! Local mesh dimensions
+    nx = mesh%dims_global(1)/mesh%par%nproc_dir(1)
+    ny = mesh%dims_global(2)/mesh%par%nproc_dir(2)
+    nz = mesh%dims_global(3)/mesh%par%nproc_dir(3)
+
+    ! Apply padding based on sz
+    nx_padded = nx - 1 + mod(-(nx - 1), sz) + sz
+    ny_padded = ny - 1 + mod(-(ny - 1), sz) + sz
+    ! Current reorder functions do not require a padding in z-direction.
+    nz_padded = nz
```
semi-h, line 110: Padding is actually related to the allocator more than anything else but now its getting sorted inside the mesh object. It can be better if it is handled by the allocator instead.

src/allocator.f90
```diff
@@ -33,17 +35,14 @@ module m_allocator
...
     !> incremented each time a new block is allocated.
     integer :: next_id = 0
-    !> Padded dimensions for x, y, and z oriented fields
-    integer :: xdims_padded(3), ydims_padded(3), zdims_padded(3)
-    !> Padded dimensions for natural Cartesian ordering
-    integer :: cdims_padded(3)
     !> The pointer to the first block on the list.  Non associated if
     !> the list is empty
     ! TODO: Rename first to head
+    type(mesh_t) :: mesh
```
semi-h, line 45: `mesh` is a member in allocator, however only the `get_padded_dims` method is being used inside. Maybe we can still pass the local domain size to the allocator and let it sort out the padding, and capture this info and store a copy in `mesh`. This way we won't need to have a copy of the mesh object in the allocator.
One use case of this in the future: If we have two separate backends (CUDA+OpenMP) working at the same time, we may want to allow them to work with their preferred SZ. Then `mesh%sz` won't be as global as we currently think it is. So it might be better to pass a `sz` when instantiating the allocator object and have it sort out the padding, and teach the padding to the `mesh` object instead.

src/mesh.f90
```diff
@@ -0,0 +1,325 @@
...
+
+  ! Stores parallel domain related information
+  type :: parallel_t
+    integer :: nrank ! local rank ID
+    integer :: nproc ! total number of ranks/proc participating in the domain decomposition
+    integer, dimension(3) :: nrank_dir ! local rank ID in each direction
+    integer, dimension(3) :: nproc_dir ! total number of proc in each direction
+    integer, dimension(3) :: n_offset  ! number of cells offset in each direction due to domain decomposition
+    integer, dimension(3) :: pnext ! rank ID of the previous rank in each direction
+    integer, dimension(3) :: pprev ! rank ID of the next rank in each direction
```
semi-h, line 27: There are few members here each with lenght 3. One alternative is having a struct with members `nrank_dir`, `nproc_dir`, etc. and have this new struct in a lenght 3 array. 
We can also discuss about naming the spatial dimensions. Mesh object uses lenght 3 arrays but we access the `dirps` objects with `x/y/zdirps`. If people prefer numbering we can change the naming for `dirps` objects too, and maybe have them in an array with size 3 instead.

semi-h (commented):
Just a few more points we might want to discuss in our meeting today

src/mesh.f90
```diff
@@ -0,0 +1,325 @@
...
+    class(mesh_t), intent(in) :: self
+    class(field_t), intent(in) :: phi
+    integer, intent(in) :: data_loc
+    integer, dimension(3) :: dims
+
+    dims = self%get_dims(phi%dir, data_loc)
+
+  end function
+
+  pure function get_dims_dir(self, dir, data_loc) result(dims)
```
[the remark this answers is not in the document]
Nanoseb: ah I actually didn't realise we would ever need `DIR_C` not on `VERT`. But that's easy enough to change. With how DIR_X/Y/Z and _C work similarly, I think I'd rather have a single function to handle the 4 of them.

src/mesh.f90
```diff
@@ -0,0 +1,325 @@
...
+        n = n
+      case(Z_FACE)
+        n = n
+      case(X_EDGE)
+      case(Y_EDGE)
+      case(Z_EDGE)
+    end select
+  end function
+
+  pure function get_location(self, i, j, k) result(xloc)
```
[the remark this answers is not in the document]
Nanoseb: yes, will do

src/omp/backend.f90
```diff
@@ -66,46 +67,47 @@ function init(globs, allocator) result(backend)
...
-    allocate (backend%dud_send_s(SZ, 1, n_block))
-    allocate (backend%dud_send_e(SZ, 1, n_block))
-    allocate (backend%dud_recv_s(SZ, 1, n_block))
-    allocate (backend%dud_recv_e(SZ, 1, n_block))
-    allocate (backend%d2u_send_s(SZ, 1, n_block))
-    allocate (backend%d2u_send_e(SZ, 1, n_block))
-    allocate (backend%d2u_recv_s(SZ, 1, n_block))
-    allocate (backend%d2u_recv_e(SZ, 1, n_block))
+    backend%mesh = mesh
+    n_groups = backend%mesh%get_n_groups(DIR_X)
```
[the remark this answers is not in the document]
Nanoseb: ah yes, I had fixes that in my previous PR but forgot about it here.

src/omp/backend.f90
```diff
@@ -283,24 +292,29 @@ subroutine tds_solve_dist(self, du, u, dirps, tdsops)
     class(field_t), intent(in) :: u
     type(dirps_t), intent(in) :: dirps
     class(tdsops_t), intent(in) :: tdsops
-    integer :: n_halo
+    integer :: n_halo, n, n_groups, dir
 
     ! TODO: don't hardcode n_halo
     n_halo = 4
+    dir = u%dir
+    n = self%mesh%get_n(u)
```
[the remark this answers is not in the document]
Nanoseb: yeah, that's my issue with `NONE`, by allowing it it means that we can never use `u%data_loc` due to the off chance it can be `NONE`. I feel like we should keep `NONE` only for edge cases or test etc. but not allow it in these kind of portion of the code.

I haven't added it yet, but I am planning on having the code through an error and crash when using `NONE` in not allowed places.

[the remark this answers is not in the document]
semi-h: This links back to `field%n` and why I was really concerned about it. I have the exact same concerns using `get_n` here. Having the `data_loc` in `field` optional makes it impossible to use `get_n` in some places, and I think this is a very good feature, and that's why I'm completely happy if `data_loc` is optional, but really concerned otherwise. And requiring a `data_loc` in most cases is effectively equivalent to making it a normal argument, not an optional one. Regarding having a `data_loc` in a field, I think just the opposite, and we should use `NONE` in most cases and a `VERT` or similar in rare cases. Because otherwise we're making the solver unnecessarily complicated. There are also cases where we can't specify a `data_loc`, but still need to use `tds_solve`.

Can you share your opinion on using `tdsops_t%n` here instead? A `tdsops` object is the building block of a `tds_solve` call and we rely on a lot of data it provides to us, including the `n` of the resulting field. In my opinion `get_n` is a good replacement here over `tdsops%n`.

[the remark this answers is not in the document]
Nanoseb: do you have examples where specifying the `data_loc` in the field won't work?

I am having issues understanding where/why it might be an issue. At the end of the day, a field has some data in it located on either vertices, cell centers, face etc. and that is 'constant' throughout its life. If the tds solve needs to loop or do something that is not of the size of the data it is working with, maybe that just needs to be handled by tds solve itself but won't affect `n`. 
What I mean is that `get_n` will always give you the `n` of the data it stores, but if for some reason tds_solve needs `n+1` or something else, then this needs to be handled differently. And it shouldn't be called `n`.

My issue with using `tdsops_t%n` is that it makes everything less clear, we will have two different `n` and depending on what we need to do we use one or the other. That's confusing for anyone (like myself) not very familiar with the tdsops module.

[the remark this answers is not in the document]
semi-h: First a very quick summary of `tds_solve` design logic: tridiagonal solver algorithm implementation (in distributed.f90) uses preprocessed set of arrays all sized equivalent to the size of tridiagonal system they solve, and lets call this size `n_tds` for the moment for clarification (this is stored in `tdsops%n`). Then there are loops in the implementation where we traverse the domain as well as these preprocessed arrays. These loops ranges are equivalent to the size of the tridiagonal system we're solving and I think its sensible to define the loop ranges based on `n_tds`. The meaningful number of entries in the output field is practically determined by the `n_tds`, so making sure that we have the right `n` in the output field beforehand and using this `n` in the loop ranges isn't a straightforward process. While the `tdsops` operator knows the size of the tridiagonal system it solves, determining the loop range in a different way from a different object is not a good idea in my opinion, and I think its error prone.

Also, replacing `tdsops%n` here with `get_n` necessitates us to specify the `data_loc` of every single field we request, even if they're short lived intermediate storages. Its true that a field stores data in one of these predetermined locations, but ambiguity around this is helping us with abstraction. For example we can pass two fields with `data_loc` `NONE`, and still get a `tds_solve` work correctly, and would be nice to keep this. 

One example where we can use this ambiguity and thus having a `data_loc` won't work:
Lets say we have a scalar at `CELL` and want to obtain say the x-derivative at `VERT`. We need 3 `tds_solve`s to get to the final result, but there are multiple paths we can choose. For example we can do `CELL` -> `Z_FACE` -> `X_EDGE` -> `VERT` or `CELL` -> `Z_FACE` -> `Y_EDGE` -> `VERT`. Its possible for us to carry out both these operations with a single subroutine such as `call custompath_p2v(dpdx, p, tdsopsA, tdsopsB, tdsopsC)`. Whatever path we decide to carry out is basically specified when we call this subroutine, and because they pass through different `data_loc`s, we can't specify the `data_loc` of intermediate fields inside the subroutine by hand. However, as long as we pass the correct set of `tdsopsA/B/C` operators, we're certain that the path we specify is respected and we obtain the correct result at the end.

I see that the naming in `tdsops%n` can be confusing, so like I used above maybe renaming it to `tdsops%n_tds` or anything else will distinguish it from `n` and make things more clear. Also, its use is strictly in the context of `tds_solve` functions I think, and really shouldn't be used anywhere else. I'll be happy to add documentation regarding `tds_solve` stuff if you think it will be helpful. What do you think?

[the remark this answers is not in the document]
Nanoseb: Thanks for the detailed explanation, it is a lot more clear now. So let's do that, rename `tdsops%n` to `tdsops%n_tds` and use it when we have to, but keep the possibility to have `data_loc` specified when we want `n` outside the solver.

As a general rule, I think it will still be good to define `data_loc` whenever possible, even if not used directly it will help debug things and check for consistency of sizes at times.

And indeed, having these kind of high level design choices written somewhere could be useful, specially for anyone getting started with contributing to x3d2.

Nanoseb:
Still WIP, but so far the main changes are
- remove `mesh` from `tdsops`
- add `tds_n` to `tdsops`
- remove `n` from dirps
- use `tds_n` instead of `mesh%get_n` when working with tds things
- compute `tds_n` based on the `operation`, `scheme` and `from_to` (this needs to be implemented)

src/tdsops.f90
```diff
@@ -122,6 +123,28 @@ function tdsops_init(n, delta, operation, scheme, n_halo, from_to, &
...
+    if (present(from_to)) then
+      select case(from_to)
+      case("v2p")
+        data_loc = CELL
+      case("p2v")
+        data_loc = VERT
+      end select
+    end if
+
+end function
```
Nanoseb, line 145: @semi-h would you be able to write such function?
You have a way better understanding of these things, so I think it is safer if you do it.

The idea is to get the `data_loc` useful for a tdsops, upon allocation of tdsops it then defines `tds_n` with `tds_n = mesh%get_n(dir, data_loc)`.

src/omp/backend.f90
```diff
@@ -66,57 +67,64 @@ function init(globs, allocator) result(backend)
...
+    integer :: data_loc, tds_n
+    real(dp) :: delta
 
     allocate (tdsops_t :: tdsops)
 
     select type (tdsops)
     type is (tdsops_t)
-      tdsops = tdsops_t(n, dx, operation, scheme, n_halo, from_to, &
+      data_loc = get_operation_data_loc(operation, scheme, from_to)
+      tds_n = self%mesh%get_n(dir, data_loc)
```
semi-h, line 125: If we get the dimensions in 3D mesh both for `vert` and `cell` first, then we can get the number of cells or vertices along a `dir` easily.
```
n_cells(:) = mesh%get_mesh_dims(CELL)
n_verts(:) = mesh%get_mesh_dims(VERT)
if (from_to == 'v2p') then
  tds_n = n_cells(dir)
else
  tds_n = n_verts(dir)
end if
```

or we can also use `mesh%get_n_dir`
```
if (from_to == 'v2p') then
  tds_n = mesh%get_n_dir(dir, CELL)
else
  tds_n = mesh%get_n_dir(dir, VERT)
end if
```

Nanoseb: But `from_to` is optional, when it isn't defined is it fine to have `tds_n = mesh%get_n_dir(dir, VERT)`?

semi-h: Yes, if not defined we assume its `v2v`, so we need `VERT` dimensions in that case too.

Nanoseb: ok, that makes things a lot easier than I thought

src/omp/backend.f90
```diff
@@ -168,61 +170,69 @@ subroutine transeq_omp_dist(self, du, dv, dw, u, v, w, dirps)
...
-    type(dirps_t), intent(in) :: dirps
-    integer :: n_halo
+    integer, intent(in) :: dir
+    integer :: n_halo, n, nproc_dir, pprev, pnext
+    integer :: n_groups
 
     ! TODO: don't hardcode n_halo
     n_halo = 4
+    n_groups = self%mesh%get_n_groups(dir)
+    n = self%mesh%get_n(u)
```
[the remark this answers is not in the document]
Nanoseb: actually, I am tempted to keep it here. This is a halo swap which has nothing to do with the tri diagonal system. So I think it is better to keep `get_n`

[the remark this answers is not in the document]
semi-h: I see, in `tds_solve` we have the operator so can easily use the `tds_n` but here we don't have an operator available.

The only issue is `get_n` from a field requires the field to have a defined `data_loc`. in `DIR_X` this is fine because the `data_loc` for `solver%u/v/w` are all defined in this PR. But we call this function in `DIR_Y` and `DIR_Z` as well, and then it will complain about a `NONE` `data_loc` and won't be able to get the `n`.

What about using `self%mesh%get_n(dir, VERT)` here instead? transeq_component is a specialist implementation and we know that it always works from `VERT` to `VERT`, so it will solve the problem. Not the best solution, but maybe we can try something else later in #98, what do you think?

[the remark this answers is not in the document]
Nanoseb: Just defined `data_loc` in a lot more fields in `solver.f90` so it should be fine. In preparation of #98, I only defined it for input fields so it shouldn't be redundant.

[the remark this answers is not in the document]
semi-h: Yes, this solves the problem.

Just to clarify, with #98 we'll be able to obtain `data_loc`s in all the fields in solver by specifying `data_loc` of the `solver%u`, `solver%v`, and `solver%w` only. Because in the rest we'll be able to obtain the resulting `data_loc` from the operations, either from `tds_solve` or `reorder`. For example for `u_y` in `transeq`, we `reorder` data from `u` to `u_y`, so `reorder` function can copy the `data_loc` as well as reordering and copying the data inside. So specifying the `data_loc` for `u_y` and many others will be redundant with #98.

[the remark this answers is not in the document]
Nanoseb: ah yes indeed. I will let you remove the redundant ones then if you feel that's needed.

tests/omp/test_omp_tridiag.f90
```diff
@@ -43,15 +51,30 @@ program test_omp_tridiag
...
+  n_glob = dims_global(1)
+  n = mesh%get_n(DIR_X, VERT)
+  n_groups = mesh%get_n_groups(DIR_X)
   n_iters = 1
 
-  allocate (u(SZ, n, n_block), du(SZ, n, n_block))
+  u_field => allocator%get_block(DIR_X, VERT)
+  du_field => allocator%get_block(DIR_X, VERT)
+  u = u_field%data
+  du = du_field%data
```
semi-h, line 77: The idea here was that we sort out the `pprev`, `pnext` and allocation of the fields (arrays here actually) ourselves so that the tridiagonal solver can be tested in isolation in a unit test (including the sendrecv that tridiagonal solver require). Even if there is something wrong with the `allocator` or `mesh` we expect this test to run correctly as long as the low level tridiagonal solver algorithm is implemented correctly.

src/omp/backend.f90
```diff
@@ -283,24 +299,30 @@ subroutine tds_solve_dist(self, du, u, dirps, tdsops)
...
+                         self%mesh%par%nproc_dir(dir), &
+                         self%mesh%par%pprev(dir), &
+                         self%mesh%par%pnext(dir))
 
     call exec_dist_tds_compact( &
       du%data, u%data, self%u_recv_s, self%u_recv_e, &
       self%du_send_s, self%du_send_e, self%du_recv_s, self%du_recv_e, &
-      tdsops, dirps%nproc_dir, dirps%pprev, dirps%pnext, dirps%n_blocks &
-      )
+      tdsops, self%mesh%par%nproc, self%mesh%par%pprev(dir), self%mesh%par%pnext(dir), &
```
semi-h, line 324: I think this has to be `nproc_dir` instead of `nproc`.

src/mesh.f90
```diff
@@ -0,0 +1,375 @@
...
+    ! Define number of cells and vertices in each direction
+    mesh%vert_dims = [nx, ny, nz]
+    mesh%cell_dims = mesh%vert_dims
+
+
+    do dir=1, 3
+      is_last_domain = (mesh%par%nrank_dir(dir) == mesh%par%nproc_dir(dir))
+      if (is_last_domain) then
+        if (mesh%periodic_dir(dir)) then
+          mesh%vert_dims(dir) = mesh%vert_dims(dir) + 1
```
semi-h, line 122: Vertex dims are always the dimensions we set in the simulation. Only the cell counts can be different. The number of cells in a direction is just one less than vertices in that direction if we're at the last region in the domain .and. the BC is .not. periodic. In all the other cases they're equal to each other.

src/mesh.f90
```diff
@@ -0,0 +1,375 @@
...
+    call domain_decomposition(mesh)
+
+    ! Local mesh dimensions
+    nx = mesh%dims_global(1)/mesh%par%nproc_dir(1)
+    ny = mesh%dims_global(2)/mesh%par%nproc_dir(2)
+    nz = mesh%dims_global(3)/mesh%par%nproc_dir(3)
+
+
+    ! TODO: fixme: hard code no periodic boundary (for now)
+    mesh%periodic_dir(:) = .false.
```
semi-h, line 111: if you're not planning to fix in the current PR you can set this to periodic so that the TGV case works once this is merged

src/mesh.f90
```diff
@@ -0,0 +1,375 @@
...
+    integer :: nx, ny, nz, dir
+    integer :: ierr
+    logical :: is_last_domain
+
+    allocate(mesh%geo)
+    allocate(mesh%par)
+    mesh%dims_global(:) = dims_global
+    ! Geometry
+    mesh%geo%L = L_global
+    mesh%geo%d = mesh%geo%L(:) / mesh%dims_global(:)
```
semi-h, line 95: `d` actually depends on periodicity but can be fixed later on as well. This is correct for a periodic case. If not periodic then we need to divide to `mesh%dims_global(dir)-1`, its like the global cell count in that direction.

src/mesh.f90
Nanoseb: Implemented it and set default to `.true.`

src/cuda/backend.f90
```diff
@@ -425,97 +434,102 @@ subroutine reorder_cuda(self, u_o, u_i, direction)
...
     class(field_t), pointer :: u_temp
     type(dim3) :: blocks, threads
+    integer :: nx_loc, ny_loc, nz_loc
 
     call resolve_field_t(u_o_d, u_o)
     call resolve_field_t(u_i_d, u_i)
 
+    nx_loc = self%mesh%get_n(DIR_X, VERT)
+    ny_loc = self%mesh%get_n(DIR_Y, VERT)
+    nz_loc = self%mesh%get_n(DIR_Z, VERT)
```
semi-h, line 444: Its name is actually misleading, sorry, but we need the padded dimensions here.

src/cuda/backend.f90
```diff
@@ -533,13 +547,18 @@ subroutine sum_yintox_cuda(self, u, u_y)
...
     type(dim3) :: blocks, threads
+    integer :: nx_loc, ny_loc, nz_loc
 
     call resolve_field_t(u_d, u)
     call resolve_field_t(u_y_d, u_y)
 
-    blocks = dim3(self%nx_loc/SZ, self%ny_loc/SZ, self%nz_loc)
+    nx_loc = self%mesh%get_n(DIR_X, VERT)
+    ny_loc = self%mesh%get_n(DIR_Y, VERT)
+    nz_loc = self%mesh%get_n(DIR_Z, VERT)
```
semi-h, line 557: padded dimensions here as well

src/cuda/backend.f90
```diff
@@ -552,13 +571,18 @@ subroutine sum_zintox_cuda(self, u, u_z)
...
     type(dim3) :: blocks, threads
+    integer :: nx_loc, ny_loc, nz_loc
 
     call resolve_field_t(u_d, u)
     call resolve_field_t(u_z_d, u_z)
 
-    blocks = dim3(self%nx_loc, self%ny_loc/SZ, 1)
+    nx_loc = self%mesh%get_n(DIR_X, VERT)
+    ny_loc = self%mesh%get_n(DIR_Y, VERT)
+    nz_loc = self%mesh%get_n(DIR_Z, VERT)
```
semi-h, line 581: also padded dimensions here

src/solver.f90
```diff
@@ -584,7 +590,7 @@ subroutine output(self, t, u_out)
     real(dp) :: enstrophy, div_u_max, div_u_mean
     integer :: ierr
 
-    if (self%backend%nrank == 0) print *, 'time = ', t
+    if (self%mesh%par%is_root()) print *, 'time = ', t
```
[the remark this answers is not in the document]
Nanoseb: All these into have to be in the mesh anyway, so I just thought to also create `is_root()` as it simplifies things and makes code more readable. We can think about moving the `par` part of the `mesh` somewhere else in the future, but I am personally fine with having it there.

src/mesh.f90
Nanoseb: I'd rather keep it this way as you may sometimes need to know things in multiple directions and it is consistent with the usage of other `mesh` fields.

Nanoseb:
I believe it is now ready to be reviewed/merged. 
The only thing I couldn't test was the CUDA backend, but if it compiled and the test pass it should be fine.

semi-h:
I can checkout and test the PR on CUDA to make sure. Is it okay if I push into this PR some commits to fix if there are any minor issues we didn't catch by eye on the CUDA backend?

I think we still don't have the automatic formatting set up, so can you run fprettify manually if you haven't done so?

Nanoseb:
yes, go ahead with testing with cuda (and pushing here). And good point about fpretiffy, I will run that

semi-h:
At the current stage it compiles and tests are passing, however the TGV case is failing on CUDA backend. I'll look into this and let you know when its all sorted.

Working a bit on the new mesh and allocator made me realise that they're a bit tangled to each other, for now it should be alright, but I think we need to look into separating them in a better way maybe. I can open an issue to point out the things we can try later.

src/cuda/backend.f90
```diff
@@ -228,42 +232,45 @@ subroutine transeq_cuda_dist(self, du, dv, dw, u, v, w, dirps, &
...
-    type(dirps_t), intent(in) :: dirps
-    integer :: n_halo
+    integer, intent(in) :: dir
+    integer :: n_halo, n, nproc_dir, pprev, pnext
+    integer :: n_groups
 
     ! TODO: don't hardcode n_halo
     n_halo = 4
+    n_groups = self%mesh%get_n_groups(dir)
+    n = self%mesh%get_n(dir, VERT)
```
semi-h, line 265: I had to change `get_n(u)` to `get_n(dir, VERT)` here because in the CUDA backend at this level we don't have a field `u` available, and it didn't compile. This should work all fine, but later on we can try to find a better solution.

semi-h:
It was a really weird bug...

For some reason, the `mesh%get_n` functions getting called inside the `poisson_fft` modules was giving the wrong value. In the base poisson module the values were correct, but in the backend poisson modules it was always wrong, and a random value.

I had to remove the calls to `mesh%get_n` inside the backend poisson modules, but still use the `mesh%get_n` when instantiating the base class and the dimensions. I think its safer in this way anyways, but still don't understand why the `mesh%get_n` calls inside the backend poisson modules were giving the wrong result. I think we need to keep an eye on this, as it may be a problem in some other parts of the code as well.

Any thoughts why there was such a behaviour? First I thought it was due to `mesh` member not being defined as `target`, but it wasn't the issue. We access lots of stuff by reference and never had such an issue.

src/solver.f90
```diff
@@ -93,37 +96,41 @@ function init(backend, time_integrator, host_allocator, &
     class(field_t), pointer :: u_init, v_init, w_init
 
     real(dp) :: x, y, z
-    integer :: nx, ny, nz, i, j, k
+    integer :: i, j, k
+    integer, dimension(3) :: dims
+    real(dp), dimension(3) :: xloc
 
     solver%backend => backend
+    solver%mesh = mesh
```
semi-h, line 104: Never experimented with copying user defined types in Fortran, are we sure that this is doing a 'deep copy'? Because mesh has some members which are also user defined types, are all of them copied into this new instance of the mesh class?

Also, would it be better if we point to the object we created in the main program? In the backend modules we actually point but here we copy instead, is there a reason for it?

Nanoseb: Oh yes, that's an oversight from my part, it should be a pointer:
```
solver%mesh => mesh
```
Otherwise indeed it will create a copy which we don't want. I will correct taht on monday

semi-h:
I would like to implement something quick with `get_dims` but wanted to ask your opinion first.

We use `get_dims` only in solver and allocator, and always with specifying a `dir` and a `data_loc`. I think the `dir` argument for `get_dims` is redundant. It only returns the 3D dimensions, so `dir` actually doesn't play a role for gathering this info.

The only case where it will be relevant is when we ask for `DIR_X/Y/Z` and a `data_loc`, but we never do this currently, do you have any plan for using this combination? I think this combination can be a bit misleading, because it returns (SZ, n, n_groups), where `n` is the actual domain size, but `SZ` and `n_groups` are inherently padded dimensions.

So if you're not planning to use `DIR_X/Y/Z` and `data_loc` combination, shall we change `get_dims` a bit such that it only inputs a `data_loc` and gives the relevant dimensions based on `data_loc` only (also can be overloaded with a field input)?

Or if you think the`DIR_X/Y/Z` and `data_loc` combination will be useful, how about renaming this to `get_field_dims`, where we require a `dir` and `data_loc` input, and also have `get_dims`, where we require only a `data_loc` input?

Nanoseb:
Yes, indeed. I implemented `get_dims` before seeing how it would be used, that makes sense. Though we can also add `get_field_dims` as it may still be useful at some point (and the name is more clear).

semi-h:
> Yes, indeed. I implemented `get_dims` before seeing how it would be used, that makes sense. Though we can also add `get_field_dims` as it may still be useful at some point (and the name is more clear).

Okay, renamed the existing `get_dims` to `get_field_dims` so that they can be used if there is need.

Then implemented a `get_dims` that inputs a `data_loc` and returns the local domain size.

Also implemented a `get_global_dims` because the FFT based Poisson solver requires the global dimensions. This required adding two new member variables in the mesh class; `global_vert_dims` and `global_cell_dims`. Please check the naming convention and the changes if they all make sense.

Nanoseb:
Looks good to me. A bit of a shame the logic of `get_n` is implemented twice (the set of `select case...`). But I don't see an obvious way to make it better without making it overly convoluted.
So I'd say it is good to go.

PR #99, opened 2024-06-21
Bump the Fortran standard to F2018 for GNU compilers

pbartholomew08:
When building with debugging flags on, the use of `error stop` in a `pure` procedure was rejected as an F2018 feature (currently setting F2008)


Nanoseb (approved):
lgtm

semi-h (approved):

PR #101, opened 2024-06-24
add reordering to/from DIR_C with openMP

### LINKED ISSUE #100: implement reorder to/from cartesian with openMP backend

Nanoseb:
closes #100

### DIFF
--- a/src/common.f90
+++ b/src/common.f90
@@ -37,46 +37,11 @@ contains
   pure subroutine get_dirs_from_rdr(dir_from, dir_to, rdr_dir)
     integer, intent(out) :: dir_from, dir_to
     integer, intent(in) :: rdr_dir
+    integer, dimension(2) :: dirs
     
-    select case (rdr_dir)
-    case (RDR_X2Y)
-      dir_from = DIR_X
-      dir_to = DIR_Y
-    case (RDR_X2Z)
-      dir_from = DIR_X
-      dir_to = DIR_Z
-    case (RDR_Y2X)
-      dir_from = DIR_Y
-      dir_to = DIR_X
-    case (RDR_Y2Z)
-      dir_from = DIR_Y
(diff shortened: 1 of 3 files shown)


pbartholomew08 (approved):
LGTM. To check my understanding, the two new indexing cases are simply stating "If you want the Cartesian coordinates of the Cartesian layout you already have them"?

src/common.f90
```diff
@@ -34,7 +34,53 @@ module m_common
 
 contains
 
-  integer function get_rdr_from_dirs(dir_from, dir_to) result(rdr_dir)
+  pure subroutine get_dirs_from_rdr(dir_from, dir_to, rdr_dir)
+    integer, intent(out) :: dir_from, dir_to
+    integer, intent(in) :: rdr_dir
+    
+    select case (rdr_dir)
```
semi-h, line 41: What do you think about using `findloc` intrinsic in Fortran instead of the `select case` block? Haven't really experimented with this before but looks like we can use it to obtain `dir_from` and `dir_to` indices easily with a single query using the `rdr_map` array.

https://gcc.gnu.org/onlinedocs/gcc-12.3.0/gfortran/FINDLOC.html

Nanoseb: yeah, that's what I was wondering too. Performance wise I am not sure what's the best. The other option would be two lut for `from_dir` and `to_dir` of size 31 (with lots of zeros).

semi-h: You're right, it can be bad for performance, this one is called very often inside the reorder loop. Maybe its better to explore different options later when we're investigating the performance.

Nanoseb: actually, that call wasn't needed so I removed it. It should be better now performance wise, regardless of the implementation

Nanoseb:
I believe it is ready now

semi-h (approved):

PR #103, opened 2024-06-24
Implement omp backend for basic math operators

pbartholomew08:


src/omp/backend.f90
```diff
@@ -382,16 +386,59 @@ subroutine vecadd_omp(self, a, x, b, y)
...
+        do j = 1, dims(2)
+          !$omp simd
+          do i = 1, SZ
+            s = s + x%data(i, j, k) * y%data(i, j, k)
+          end do
+          !$omp end simd
+        end do
+      end do
+      !$omp end parallel do
+    end if
```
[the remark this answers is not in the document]
Nanoseb: yes, indeed

src/omp/backend.f90
```diff
@@ -382,16 +386,59 @@ subroutine vecadd_omp(self, a, x, b, y)
...
+    dir = x%dir
+
+    s = 0.0_dp
+    if (dir == DIR_C) then
+      !$omp parallel do reduction(+:s) collapse(3)
+      do k = 1, dims(3)
+        do j = 1, dims(2)
+          do i = 1, dims(1)
+            s = s + x%data(i, j, k) * y%data(i, j, k)
+          end do
```
semi-h, line 420: If you use the `scalar_product` operator with `DIR_C` arrays and think that it could make a difference in performance, there is a simple trick you can implement to enable vectorisation here as well. 

You just need to separate the last loop into two as
```
do i_outer = 1, (dims(1) - 1)/SZ + 1
  ! pragma with reduction on s and private i
  do i_inner = 1, SZ
    i = (i_outer - 1)*SZ + i_inner
    s = s + ...
  end do
end do
! a cleanup loop
do i = 1, mod(dims(1), SZ)
  s = s + ... !final index = i + dims(1)/SZ
end do
```

pbartholomew08: Done - also reduces total amount of code.

pbartholomew08:
I've added tests (OMP only, however it should be relatively easy to adapt to CUDA) for the scalar produce and sum into x implementations (vecadd appears to be tested indirectly through the timestepping tests).

src/omp/backend.f90
```diff
@@ -379,18 +412,91 @@ subroutine vecadd_omp(self, a, x, b, y)
     class(field_t), intent(in) :: x
     real(dp), intent(in) :: b
     class(field_t), intent(inout) :: y
+    integer, dimension(3) :: dims
+    integer :: i, j, k, ii
+
+    integer :: nvec, remstart
+
+    if ((x%dir /= y%dir) .or. (x%data_loc /= y%data_loc)) then
```
semi-h, line 420: I wouldn't do a sanity check for the `data_loc` here because for example time integrator arrays don't need a `data_loc`, but the arrays they're summed over sometimes have a `data_loc`. I think its reasonable to leave this to the user. We can even use `vecadd` to obtain crude interpolations later on if we need using two arrays with different `data_loc`s.

src/omp/backend.f90
```diff
@@ -379,18 +412,91 @@ subroutine vecadd_omp(self, a, x, b, y)
...
 
     class(omp_backend_t) :: self
     class(field_t), intent(in) :: x, y
+    integer, dimension(3) :: dims
+    integer :: i, j, k, ii
+    integer :: ierr
+
+    integer :: nvec, remstart
+
+    if ((x%dir /= y%dir) .or. (x%data_loc /= y%data_loc)) then
```
semi-h, line 466: `scalar_product` requires the actual dimensions of an array, so making sure that both arrays have a `data_loc /= NONE` is important. I think mesh methods we call here will complain anyways but having a sanity check here can make it easy to debug.

src/omp/backend.f90
```diff
@@ -379,18 +412,91 @@ subroutine vecadd_omp(self, a, x, b, y)
...
+      error stop "Called scalar product with incompatible fields"
+    end if
+
+    dims = self%mesh%get_field_dims(x)
```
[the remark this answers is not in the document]
pbartholomew08: > get_field_dims in the mesh class is a bit misleading I believe. It returns the actual dimensions for a DIR_C array, but for DIR_X/Y/Z arrays its a bit different. It returns the actual number of entries along its own DIR, but the first and last dims are always SZ and the number of groups.

I'm not sure I understand the use case for `get_field_dims` then, based on
```
 pure function get_padded_dims_dir(self, dir) result(dims_padded)
  !! Getter for padded dimensions with structure in `dir` direction
```
I would have thought the former gets the non-padded dimensions, and the latter the padded dimensions?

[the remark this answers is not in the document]
semi-h: I can't think of a use case for `get_field_dims` to be honest. If one needs non-padded dimensions along the present `DIR` and also the number of groups its better to get these with two separate calls instead.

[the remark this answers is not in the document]
semi-h: The problem is we can't really have non-padded dimensions in `DIR_X/Y/Z` arrangements. Certain combinations of the first index and last index result in padding in that case.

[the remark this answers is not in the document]
pbartholomew08: Would it not be better to have it return the non-padded dimensions? Are the padded dimensions not just `size(phi%data)`? How do we get the non-padded dimensions then? I'm sorry this is unclear to me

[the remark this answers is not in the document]
semi-h: This is very tricky and not documented at all, my bad.
```
   2 4 6 8 10
   o o o o o
   x x x x x
   x x x x x
___x x x x x___
   x x x x x
   x x x x x
   x x x x x
   x x x x x
   1 3 5 7 9
```
group numbers are either below or above. Each group have 4 lines in total. `x` indicates a real entry, `o` indicates a padding. `_`is just a visual guide for separating two groups. Domain size is (nx=n, ny=7, nz=5) Lets say x-direction is towards inside the screen, y-dir is up, and z-dir is to the right. The dims of our `DIR_X` array shown here is `(SZ=4, nx, 10)`. The groups numbered 2, 4, 6, 8, 10 have some padding in them, but 1, 3, 5, 7, and 9 don't. So we can't really have a non-padded `k` index. What we can say is that only group numbers divisible by 2 include some amount of padding. To make this easy we can divide a `k` loop into two loops, so that one loop goes from 1 to 2(because we have two groups in y-direction, ny=7, SZ=4). and then the other loop goes from 1 to 5 (because nz=5).

src/omp/backend.f90
pbartholomew08: Done

src/omp/backend.f90
pbartholomew08: Done

src/omp/backend.f90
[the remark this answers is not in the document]
semi-h: padded dimension of an array is equal to `size(phi%data)`.

By the way, the easiest solution here would be calling a reorder function and reorder whatever `DIR` we have happen to have in `scalar_product` into `DIR_C`. Then the 3 dimensions and their paddings are very easy to deal with. non-padded dimensions can be obtained by `get_dims(phi%data_loc)`, and then loop bounds will make sure we stay within the non-padded entries. The only issue with this is it does a reorder unnecessarily. In the solver we use `scalar_product` only once in a while, for the enstrophy, so no problem if it doens't perform well. But I think the iterative solver uses it a bit more often, but not sure how much weight it has overall.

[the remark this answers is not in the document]
pbartholomew08: OK, I'll go with the reordering approach. The iterative solver shouldn't have to do this, we only need to provide a routine for `lapl(p)`, any scalar products in the solver are handled by PETSc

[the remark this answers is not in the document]
Nanoseb: Could we not have a function that returns the 'non padded' `SZ` value given the group index? It won't be perfect for performance because vectorisation might be harder, but that should be better than reordering and be more 'clear'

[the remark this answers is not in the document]
pbartholomew08: OK, scalar product is now implemented using reordering to DIR_C :)

pbartholomew08:
I think this branch is ready to go

src/omp/backend.f90
```diff
@@ -379,18 +412,120 @@ subroutine vecadd_omp(self, a, x, b, y)
...
+    if ((x%data_loc == NONE) .or. (y%data_loc == NONE)) then
+      error stop "You must set the field location before calling scalar product"
+    end if
+    if (x%data_loc /= y%data_loc) then
+      error stop "Called scalar product with incompatible fields"
+    end if
+
+    ! Reorient data into temporary DIR_C storage
+    x_ => self%allocator%get_block(DIR_C)
+    x_%data_loc = x%data_loc
```
Nanoseb, line 478: I'd use `get_block(DIR_C, x%data_loc)` or use the setter instead of directly setting it.

src/omp/backend.f90
```diff
@@ -379,18 +412,120 @@ subroutine vecadd_omp(self, a, x, b, y)
...
+    if (x%dir == DIR_C) then
+      call self%copy_f_to_data(x_%data, x)
+    else
+      RDR = get_rdr_from_dirs(x%dir, DIR_C)
+      call self%reorder(x_, x, RDR)
+    end if
+    y_ => self%allocator%get_block(DIR_C)
+    y_%data_loc = y%data_loc
```
Nanoseb, line 486: I'd use `get_block(DIR_C, x%data_loc)` or use the setter instead of directly setting it.

src/omp/backend.f90
```diff
@@ -379,18 +412,120 @@ subroutine vecadd_omp(self, a, x, b, y)
...
+    s = 0.0_dp
+    !$omp parallel do reduction(+:s) private(i, ii) collapse(2)
+    do k = 1, dims(3)
+      do j = 1, dims(2)
+        ! Execute inner vectorised loops
+        do ii = 1, nvec
+          !$omp simd reduction(+:s)
+          do i = 1, SZ
+            s = s + x_%data(i + (ii - 1)*SZ, j, k)* &
+                y_%data(i + (ii - 1)*sZ, j, k)
```
Nanoseb, line 508: SZ case

tests/test_sum_zintox.f90
```diff
@@ -0,0 +1,103 @@
+program test_sum_zintox
```
Nanoseb, line 1: because there is only very few differences between this test and test_sum_yintox, maybe they can be combined into a single test (will just ease maintenance if/when things change)

Nanoseb:
Just a handful of **very** minor comments, otherwise lgtm

src/omp/backend.f90
```diff
@@ -379,18 +412,120 @@ subroutine vecadd_omp(self, a, x, b, y)
...
+
+    ! Reorient data into temporary DIR_C storage
+    x_ => self%allocator%get_block(DIR_C)
+    x_%data_loc = x%data_loc
+    if (x%dir == DIR_C) then
+      call self%copy_f_to_data(x_%data, x)
+    else
+      RDR = get_rdr_from_dirs(x%dir, DIR_C)
+      call self%reorder(x_, x, RDR)
+    end if
```
semi-h, line 484: I think the whole if block can be replaced by the higher level `get_field_data(x_%data, x)` function instead. If the `x` field is not in `DIR_C`, `get_field_data` first reorders it to `DIR_C` (by default) and then copies data back to the output array.

src/omp/backend.f90
pbartholomew08: Done

src/omp/backend.f90
pbartholomew08: ah good point, done.

src/omp/backend.f90
pbartholomew08: done

src/omp/backend.f90
pbartholomew08: Done

tests/test_scalar_product.f90
```diff
@@ -0,0 +1,160 @@
...
+
+    if (nrank == 0) then
+      print *, "Testing ", test
+    end if
+    
+    a => backend%allocator%get_block(dir)
+    b => backend%allocator%get_block(dir)
+
+    call a%set_data_loc(VERT)
+    call b%set_data_loc(VERT)
```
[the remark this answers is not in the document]
pbartholomew08: Done

tests/test_sum_zintox.f90
pbartholomew08: Good point - done

pbartholomew08:
I think all resolved - I've left the larger conversation about using reordering unresolved in case there's anything there I've missed.

semi-h (approved):

PR #110, opened 2024-07-02
Introduce a namelist for basic parameters.

semi-h:
I did some tests with varying domain sizes and dt, it was a pain to compile each time. This is not exhaustive list of parameters but at least will make simple tests easy.

Any other parameter suggestions other than these?

### DIFF
--- a/src/solver.f90
+++ b/src/solver.f90
@@ -98,8 +98,8 @@ contains
     character(30) :: der1st_scheme, der2nd_scheme, &
                      interpl_scheme, stagder_scheme
     namelist /solver_params/ Re, dt, n_iters, n_output, poisson_solver_type, &
-                             time_intg, der1st_scheme, der2nd_scheme, &
-                             interpl_scheme, stagder_scheme
+      time_intg, der1st_scheme, der2nd_scheme, &
+      interpl_scheme, stagder_scheme
 
     real(dp) :: x, y, z
     integer :: i, j, k
@@ -207,7 +207,7 @@ contains
     type(dirps_t), intent(inout) :: dirps
     class(base_backend_t), intent(in) :: backend
     character(*), intent(in) :: der1st_scheme, der2nd_scheme, &
-                                 interpl_scheme, stagder_scheme
+                                interpl_scheme, stagder_scheme
 
     call backend%alloc_tdsops(dirps%der1st, dirps%dir, &
                               'first-deriv', der1st_scheme)
--- a/src/omp/backend.f90
+++ b/src/omp/backend.f90
@@ -452,7 +452,7 @@ contains
 
     use mpi
 
-    use m_common, only: NONE, get_rdr_from_dirs
+    use m_common, only: none, get_rdr_from_dirs
 
     implicit none
 
@@ -464,8 +464,8 @@ contains
     integer :: nvec, remstart
     integer :: ierr
 
-    if ((x%data_loc == NONE) .or. (y%data_loc == NONE)) then
-      error stop "You must set the field location before calling scalar product"
+    if ((x%data_loc == none) .or. (y%data_loc == none)) then
+      error stop "You must set the data_loc before calling scalar product"
     end if
     if (x%data_loc /= y%data_loc) then
       error stop "Called scalar product with incompatible fields"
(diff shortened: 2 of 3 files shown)


pbartholomew08:
do we want to specify [Px, Py, Pz] in the namelist?

semi-h:
Added nproc_dir in the namelist so that we can specify the decomposition in the input file. In case the input file specification doesn't match with the number of ranks requested with `mpirun`, we set nproc_dir so that we have a 1D domain decomposition in Z direction. Probably need to change this when we add support for 2D decomposition in the FFTs, can even come up with a way to set the decomposition in an automated way to maximise performance.

src/solver.f90
```diff
@@ -105,18 +108,30 @@ function init(backend, mesh, time_integrator, host_allocator, &
...
+    ! set defaults
+    Re = 1600._dp; dt = 0.001_dp; n_iters = 20000; n_output = 100
+    poisson_solver_type = 'FFT'
+
+    if (command_argument_count() >= 1) then
+      call get_command_argument(1, input_file)
+      open (100, file=input_file)
+      read (100, nml=solver_params)
+      close (100)
+    end if
```
semi-h, line 129: Any thoughts on getting a command argument inside a subroutine like this?

semi-h:
I plan to add a parameter for specifiying the time integrator as well. I'll also look into setting BCs from the input file.

Are we okay to have strings in the input namelist? For example, the way we specify the Poisson solver currently is;
```
&solver_params
...
poisson_solver_type = 'FFT' ! or 'CG', anything else stops the program with error
/End
```

For example, are we happy to set BCs like
```
BC_x_start = 'periodic' ! or 'neumann' or `dirichlet`
```
or do we want integers 0, 1, 2 like in Xcompact3D? Also, Xcompact3D namings for the BC are like `nclx1, nclxn`, shall we use `start/end` or stick with `1` or `n` as suffix? Or any other idea?

Nanoseb:
Just a quick comment, is only having a single BC type per BC not limiting? Aren't there cases where we would want the velocity to be Dirichlet and pressure to be Neumann for the same boundary for example?

slaizet:
To discuss (this afternoon) but we never impose BC on the pressure however we may want different BC for velocity/scalar

src/backend.f90
```diff
@@ -27,7 +27,6 @@ module m_base_backend
     real(dp) :: nu
     class(mesh_t), pointer :: mesh
     class(allocator_t), pointer :: allocator
-    class(dirps_t), pointer :: xdirps, ydirps, zdirps
```
pbartholomew08, line 30: Is this related to adding the namelist?

src/mesh.f90
```diff
@@ -77,31 +80,51 @@ module m_mesh
...
-      mesh%periodic_BC(:) = .true.
-    end if
+    BC_all(1, 1) = BC_x(1); BC_all(1, 2) = BC_x(2)
+    BC_all(2, 1) = BC_y(1); BC_all(2, 2) = BC_y(2)
+    BC_all(3, 1) = BC_z(1); BC_all(3, 2) = BC_z(2)
+    do dir = 1, 3
+      do j = 1, 2
+        select case (trim(BC_all(dir, j)))
+        case ('periodic')
+          mesh%BCs_global(dir, j) = BC_PERIODIC
```
pbartholomew08, line 109: we could add a check that both BCs are periodic

src/solver.f90
pbartholomew08: I'm not sure what you mean by the question, I think you need to get the count (as done) and then index into this to find the value you want (as done).

src/mesh.f90
semi-h: Done

src/backend.f90
semi-h: Not really, I just moved dirps instantiation inside solver and cleaned relevant things a bit. Also moved time integrator inside solver to simplify the input file namelist groups.

src/solver.f90
semi-h: I was thinking maybe passing the file name as an argument when instantiating a solver object from the main program. Because otherwise if a main program uses solver class it requires an argument, but if the file name is passed as an argument the main program can be more flexible when retrieving and passing the input file.

Nanoseb (approved):

PR #112, opened 2024-07-09
Add actuator disc model

semi-h:
This PR implements the disc model using the existing backend functions, so compatible with all the backends.

### DIFF
--- a/src/solver.f90
+++ b/src/solver.f90
@@ -11,6 +11,7 @@ module m_solver
   use m_tdsops, only: tdsops_t, dirps_t
   use m_time_integrator, only: time_intg_t
   use m_mesh, only: mesh_t
+  use m_windturb_adm, only: windturb_adm_t
 
   implicit none
 
@@ -45,6 +46,7 @@ module m_solver
     real(dp) :: dt, nu
     integer :: n_iters, n_output
     integer :: ngrid
+    logical :: wind_turbine
 
     class(field_t), pointer :: u, v, w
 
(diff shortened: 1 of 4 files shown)


src/windturb_adm.f90
```diff
@@ -0,0 +1,190 @@
...
+
+    deallocate (vel_avg)
+
+    ! no filtering for the moment
+    self%disc(:)%U_disc_filt = self%disc(:)%U_disc
+
+    ! use a hack to zeroise the force fields
+    call self%backend%vecadd(0._dp, Fx, 0._dp, Fx)
+    call self%backend%vecadd(0._dp, Fy, 0._dp, Fy)
+    call self%backend%vecadd(0._dp, Fz, 0._dp, Fz)
```
semi-h, line 173: Just use `call Fx%fill(0._dp)` so on

semi-h: this is just to zeroise the field

src/solver.f90
```diff
@@ -542,6 +560,32 @@ subroutine curl(self, o_i_hat, o_j_hat, o_k_hat, u, v, w)
...
+    Fx => self%backend%allocator%get_block(DIR_X)
+    Fy => self%backend%allocator%get_block(DIR_X)
+    Fz => self%backend%allocator%get_block(DIR_X)
+
+    call self%windturb_adm%compute_sources(Fx, Fy, Fz, u, v, w)
+
+    ! add resulting forces into the rhs of the momentum equation
+    !call self%backend%vecadd(self%windturb_adm%rho_air, Fx, 1._dp, du)
+    !call self%backend%vecadd(self%windturb_adm%rho_air, Fy, 1._dp, dv)
+    !call self%backend%vecadd(self%windturb_adm%rho_air, Fz, 1._dp, dw)
```
semi-h, line 581: uncomment here to add the contribution of the wind turbines back into the RHS

PR #114, opened 2024-07-19
Fix a bug in OpenMP distributed tridiagonal solver

semi-h:
Last week when implementing Thomas on CUDA I realised we use this trick often, but its not really suitable for the OpenMP backend. Thought all was removed when quickly checked last week, haven't noticed this one!

### DIFF
--- a/src/omp/kernels/distributed.f90
+++ b/src/omp/kernels/distributed.f90
@@ -26,7 +26,7 @@ contains
     integer :: i, j!, b
 
     real(dp) :: c_m4, c_m3, c_m2, c_m1, c_j, c_p1, c_p2, c_p3, c_p4, &
-                temp_du, alpha, last_r
+                alpha, last_r
 
     ! store bulk coeffs in the registers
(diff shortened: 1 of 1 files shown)


Nanoseb:
lgtm, but why is it not suitable? I am not sure I understand what is going wrong.
Also, if the results produced were wrong it may be a sign we need better tests.

pbartholomew08:
`temp_du` needs to be an array for SIMD access

Nanoseb:
ah yes of course

Nanoseb (approved):

pbartholomew08:
I agree though, it indicates the tests aren't catching something

Nanoseb:
This is maybe just a performance issue. The compiler may detect it can't vectorise the loop so it doesn't?

pbartholomew08:
Confirming that would require checking assembly/vectorisation reports. I think that the pragma means "vectorise this, ignoring reasons not to", but with strict debug flags enabled it highlights as a warning and or raises an error

PR #115, opened 2024-07-19
Make BC specifications more verbose in tdsops

semi-h:
Continuation of #111. Ended up passing BCs as integer parameters.


Nanoseb (approved):
Looks good to me. Having integer parameters is indeed neater than carrying strings around.

PR #118, opened 2024-08-06
OMP implementation of Thomas algorithm

pbartholomew08:
Initial pass at implementing Thomas algorithm based on the CUDA version, it doesn't yet pass the tests, perhaps a 2nd pair of eyes will spot something.

@semi-h the initial value of `du(i, jm1, b)` on line 159 of `cuda/thomas.f90` would appear to be unset to me and here the error norm of the periodic case is significantly worse. Should it be initialised outside the function or I'm misunderstanding?

### DIFF
--- a/cmake/decomp2d/downloadBuild2decomp.cmake.in
+++ b/cmake/decomp2d/downloadBuild2decomp.cmake.in
@@ -1,6 +1,6 @@
 # downloadBuild2decomp.cmake.in
 #
-cmake_minimum_required(VERSION 3.0.2)
+cmake_minimum_required(VERSION 3.10)
 
 project(downloadBuild2decomp NONE)
 
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -18,6 +18,7 @@ set(SRC
   backend/omp/common.f90
   backend/omp/kernels/distributed.f90
   backend/omp/kernels/spectral_processing.f90
+  backend/omp/kernels/thomas.f90
   backend/omp/sendrecv.f90
   backend/omp/exec_dist.f90
   backend/omp/exec_thom.f90
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -20,6 +20,7 @@ set(SRC
   backend/omp/kernels/spectral_processing.f90
   backend/omp/sendrecv.f90
   backend/omp/exec_dist.f90
+  backend/omp/exec_thom.f90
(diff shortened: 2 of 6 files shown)


src/omp/exec_thom.f90
```diff
@@ -0,0 +1,175 @@
...
+        !! Forward pass
+        do j = 1, n
+          do i = 1, 4
+            jm(i) = modulo(j - (1 + i), n) + 1
+            jp(i) = modulo(j - (n + (i - 1)), n) + 1
+          end do
+
+          !$omp simd
+          do i = 1, SZ
+            du(i, j, k) = sum(coeffs(1:4) * u(i, jm(1:4), k))
```
semi-h, line 131: The mapping between coeffs 1 to 4 and jm(1:4) seems like mismatched to indices provided via `jm` array. I think `jm` should be reversed here like `(4:1:-1)`. Also very curious if such index accesses can be vectorised. I don't trust compilers and instinctively avoid these but completely fine if we can achieve performance we're aiming.

pbartholomew08:
Issue was in the definition of the `jm` and `jp` array indices. Current solution still uses `sum()` I don't think this should inhibit the `omp simd` around the outer loop.

semi-h:
I think it would be better to move the outer omp parallel loop outside the `der_univ` functions like in the distributed algorithm implementation. Then we call these core functions designed to operate on a single group in a loop as in
https://github.com/xcompact3d/x3d2/blob/9ab64bf4896999641d78bd3121dd5e300f1ec2f6/src/omp/exec_dist.f90#L39-L40
This would be useful for the cache-blocked transport equation based on Thomas algorithm. Distributed implementation uses this idea here
https://github.com/xcompact3d/x3d2/blob/9ab64bf4896999641d78bd3121dd5e300f1ec2f6/src/omp/exec_dist.f90#L110-L124
It allows reading input arrays only once, and also writing outputs only once. Improves performance when working on multiple operations.

Have you had a chance to benchmark the performance?

pbartholomew08:
Ah, good point, I'll do that.

No, only just got this working so not benchmarked yet.

pbartholomew08:
To be resolved using @semi-h's optimised code.

semi-h:
Updated the implementation based on the version I used for benchmarking for the DistD2 paper on Marenostrum. The fused implementation for the transeq operation is not implemented here for Thomas algorithm but I can sort that out later.

PR #120, opened 2024-08-27
Add a vector_calculus class

semi-h:
This PR moves the vector calculus operations currently implemented inside solver class out to a dedicated new class. This new structure should allow accessing vector calculus operations outside the solver class. Examples include the laplacian operation iterative Poisson solver needs and curl/divergence for postprocessing/monitoring.

### DIFF
--- a/src/backend.f90
+++ b/src/backend.f90
@@ -71,11 +71,10 @@ module m_base_backend
       !! transeq equation obtains the derivatives direction by
       !! direction, and the exact algorithm used to obtain these
       !! derivatives are decided at runtime. Backend implementations
-      !! are responsible from directing calls to transeq_ders into
-      !! the correct algorithm.
+      !! are responsible from directing calls to tds_solve to the
+      !! correct algorithm.
       import :: base_backend_t
       import :: field_t
-      import :: dirps_t
       import :: tdsops_t
       implicit none
 
--- a/src/backend.f90
+++ b/src/backend.f90
@@ -67,12 +67,12 @@ module m_base_backend
   end interface
 
   abstract interface
-    subroutine tds_solve(self, du, u, dirps, tdsops)
-         !! transeq equation obtains the derivatives direction by
-         !! direction, and the exact algorithm used to obtain these
-         !! derivatives are decided at runtime. Backend implementations
-         !! are responsible from directing calls to transeq_ders into
-         !! the correct algorithm.
+    subroutine tds_solve(self, du, u, tdsops)
+      !! transeq equation obtains the derivatives direction by
+      !! direction, and the exact algorithm used to obtain these
+      !! derivatives are decided at runtime. Backend implementations
+      !! are responsible from directing calls to transeq_ders into
+      !! the correct algorithm.
       import :: base_backend_t
       import :: field_t
       import :: dirps_t
@@ -82,7 +82,6 @@ module m_base_backend
       class(base_backend_t) :: self
       class(field_t), intent(inout) :: du
       class(field_t), intent(in) :: u
-      type(dirps_t), intent(in) :: dirps
       class(tdsops_t), intent(in) :: tdsops
     end subroutine tds_solve
   end interface
--- a/src/vector_calculus.f90
+++ b/src/vector_calculus.f90
@@ -19,7 +19,7 @@ module m_vector_calculus
 
   interface vector_calculus_t
     module procedure init
-  end interface vector_calculus_t 
+  end interface vector_calculus_t
 
 contains
 
@@ -33,7 +33,8 @@ contains
 
   end function init
 
-  subroutine curl(self, o_i_hat, o_j_hat, o_k_hat, u, v, w, x_der1st, y_der1st, z_der1st)
+  subroutine curl(self, o_i_hat, o_j_hat, o_k_hat, u, v, w, &
+                  x_der1st, y_der1st, z_der1st)
     !! Curl of a vector field (u, v, w).
     !! Inputs from velocity grid and outputs to velocity grid.
     implicit none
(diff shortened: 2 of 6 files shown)


src/backend.f90
```diff
@@ -67,12 +67,12 @@ end subroutine transeq_ders
   end interface
 
   abstract interface
-    subroutine tds_solve(self, du, u, dirps, tdsops)
-         !! transeq equation obtains the derivatives direction by
-         !! direction, and the exact algorithm used to obtain these
-         !! derivatives are decided at runtime. Backend implementations
-         !! are responsible from directing calls to transeq_ders into
-         !! the correct algorithm.
+    subroutine tds_solve(self, du, u, tdsops)
```
semi-h, line 70: Removed dirps argument here, it had become redundant when we introduced mesh class a while ago.

src/vector_calculus.f90
```diff
@@ -0,0 +1,346 @@
...
+    type(vector_calculus_t) :: vector_calculus
+
+    vector_calculus%backend => backend
+
+  end function init
+
+  subroutine curl(self, o_i_hat, o_j_hat, o_k_hat, u, v, w, &
+                  x_der1st, y_der1st, z_der1st)
+    !! Curl of a vector field (u, v, w).
+    !! Inputs from velocity grid and outputs to velocity grid.
```
Nanoseb, line 39: Shall we use the `CELL`/`FACE` etc. nomenclature instead of velocity/pressure grids? Just to be consistent and general

src/vector_calculus.f90
```diff
@@ -0,0 +1,346 @@
...
+    call self%backend%allocator%release_block(dudy_x)
+
+  end subroutine curl
+
+  subroutine divergence_v2p(self, div_u, u, v, w, &
+                            x_stagder_v2p, x_interpl_v2p, &
+                            y_stagder_v2p, y_interpl_v2p, &
+                            z_stagder_v2p, z_interpl_v2p)
+    !! Divergence of a vector field (u, v, w).
+    !! Inputs from velocity grid and outputs to pressure grid.
```
Nanoseb, line 133: Same, convert to `CELL` and `FACE` data location instead of velocity/pressure

src/vector_calculus.f90
```diff
@@ -0,0 +1,346 @@
...
+    call self%backend%allocator%release_block(dw_z)
+
+  end subroutine divergence_v2p
+
+  subroutine gradient_p2v(self, dpdx, dpdy, dpdz, pressure, &
+                          x_stagder_p2v, x_interpl_p2v, &
+                          y_stagder_p2v, y_interpl_p2v, &
+                          z_stagder_p2v, z_interpl_p2v)
+    !! Gradient of a scalar field 'pressure'.
+    !! Inputs from pressure grid and outputs to velocity grid.
```
Nanoseb, line 228: same as above

src/vector_calculus.f90
```diff
@@ -0,0 +1,346 @@
...
+    class(field_t), pointer :: u_y, d2u_y, u_z, d2u_z
+
+    ! d2u/dx2
+    call self%backend%tds_solve(lapl_u, u, x_der2nd)
+
+    ! y directional temporary fields
+    u_y => self%backend%allocator%get_block(DIR_Y)
+    d2u_y => self%backend%allocator%get_block(DIR_Y)
+
+    call self%backend%reorder(u_y, u, RDR_X2Y)
```
Nanoseb, line 320: is it worth checking that `u` is in `DIR_X` before doing that? or at least add a comment in the header to explain that it is expected

Nanoseb (commented):
looks good to me, I just think it would be good to properly define what is expected from the input variables in terms of direction and dataloc.

src/vector_calculus.f90
semi-h: `tds_solve` is doing a check based on directions so added a comment in the laplacian subroutine header to explain

semi-h:
> looks good to me, I just think it would be good to properly define what is expected from the input variables in terms of direction and dataloc.

Thanks! Changed the namings from velocity/pressure grid to data_loc based VERT/CELL. Also added a bit more explanation in subroutine headers to explain expected `DIR`s and `data_loc`s for input and output arrays.

src/backend.f90
```diff
@@ -67,12 +67,12 @@ end subroutine transeq_ders
...
+      !! transeq equation obtains the derivatives direction by
+      !! direction, and the exact algorithm used to obtain these
+      !! derivatives are decided at runtime. Backend implementations
+      !! are responsible from directing calls to transeq_ders into
+      !! the correct algorithm.
       import :: base_backend_t
       import :: field_t
       import :: dirps_t
```
pbartholomew08, line 78: Based on the above comment, I think we can get rid of this `import :: dirps_t` too?

src/vector_calculus.f90
```diff
@@ -0,0 +1,368 @@
...
+  use m_base_backend, only: base_backend_t
+  use m_common, only: dp, DIR_X, DIR_Y, DIR_Z, &
+                      RDR_X2Y, RDR_X2Z, RDR_Y2X, RDR_Y2Z, RDR_Z2X, RDR_Z2Y
+  use m_tdsops, only: tdsops_t
+
+  implicit none
+
+  type :: vector_calculus_t
+    !! Defines vector calculus operators
+    class(base_backend_t), pointer :: backend
```
pbartholomew08, line 12: This is the kind of circular pointer that I think we should try and avoid (i.e. a pointer to the `backend` object in the parent class of `vector_calculus`). Could we pass `backend` as an argument to `vector_calculus` methods? e.g.
```
call vector_calculus%curl(..., backend)
```

pbartholomew08:
> Nanoseb: is it worth checking that u is in DIR_X before doing that? or at least add a comment in the header to explain that it is expected
> semi-h: tds_solve is doing a check based on directions so added a comment in the laplacian subroutine header to explain

I think it might be worth checking the conditions at the start of the curl/div/grad/xxx subroutines. A "invalid field locations passed to curl" will be easier to debug than "invalid fields in tdsops" (which you then need to figure out where it came from). Having the tdsops perform this last level check is good to ensure we catch everything, but a higher-level check might ultimately be beneficial in my opinion.

src/vector_calculus.f90
semi-h: We don't really have a circular pointer or dependency here. `backend`, `solver` and `vector_calculus` are all independent classes and there is no inheritance relationship among them. There are other classes too that use `backend`, like `time_integrator` and new wind turbine stuff. In a class diagram `backend` would be under all of them so I don't see a problem having a pointer to `backend` in these classes. Why do you think we should avoid?

PR #127, opened 2024-10-08
Compile based on present architecture

semi-h:


Nanoseb (approved):
yes, lgtm. I won't have to add it manually

Nanoseb:
hum, actually none of the flags in that line are picked up by cmake it seems. Let me double check

Nanoseb:
(but changing them directly in cmake also propagates it for the tests, so that part is fine)

Nanoseb:
Because these variables are internal ones defined already by cmake they had to be overwritten and a test had to be added so that users could still use `ccmake` or anything else to edit them.

semi-h:
I've tried locally and all good for me, but the GitHub action failed here in the PR, do you have any idea why?

Nanoseb:
do we know which version of cmake the CI uses? it doesn't seem to detect openmp related things properly. This isn't something that was introduced with this PR but flags set here never got actually used before so we see it only now

PR #130, opened 2024-11-22
Reduce memory use in transeq for both backends

### LINKED ISSUE #40: Reuse result field in transeq_component as a temporary field
In transeq we call _x _y and _z components separately, and each time we pass a field to store the result as well as 3 temporary fields that distributed algorithm requires. Because the result is assembled only at the very last step, we can actually pass only 2 temporary fields and use the result field in place of a temporary storage. **This eliminates 1 field size worth of memory usage, reducing the total down to 14** (See #37). It won't make any diference in terms of performance in the CUDA backend, however, OpenMP backend performance will be improved slightly. On the OpenMP backend reusing result as a temporary array removes an unnecessary single field sized read, reducing the total from 14 to 13 per transeq_component call and saving %7 of memory bandwidth. And I expect an equivalent speedup in this case.

https://github.com/xcompact3d/x3d2/blob/26ccadd4b033a99a63072ab47518bd092060978f/src/cuda/backend.f90#L259-L263
result in line 260
temporary fields in 263

https://github.com/xcompact3d/x3d2/blob/26ccadd4b033a99a63072ab47518bd092060978f/src/cuda/exec_dist.f90#L112-L113
And here in line 113 lets say we'll have `r_u, v, dud, d2u` only, because `du` will be in `r_u`.

semi-h:
Closes #40.

### DIFF
--- a/src/cuda/exec_dist.f90
+++ b/src/cuda/exec_dist.f90
@@ -66,7 +66,8 @@ contains
     implicit none
 
     ! r_du = -1/2*(v*d1(u) + d1(u*v)) + nu*d2(u)
-    real(dp), device, dimension(:, :, :), intent(inout) :: r_du
+    !> The result array, it is also used as temporary storage
+    real(dp), device, dimension(:, :, :), intent(out) :: r_du
     real(dp), device, dimension(:, :, :), intent(in) :: u, u_recv_s, u_recv_e
     real(dp), device, dimension(:, :, :), intent(in) :: v, v_recv_s, v_recv_e
 
--- a/src/cuda/exec_dist.f90
+++ b/src/cuda/exec_dist.f90
@@ -56,8 +56,8 @@ contains
   end subroutine exec_dist_tds_compact
 
   subroutine exec_dist_transeq_3fused( &
-    r_u, u, u_recv_s, u_recv_e, v, v_recv_s, v_recv_e, &
-    du, dud, d2u, &
+    r_du, u, u_recv_s, u_recv_e, v, v_recv_s, v_recv_e, &
+    dud, d2u, &
     du_send_s, du_send_e, du_recv_s, du_recv_e, &
     dud_send_s, dud_send_e, dud_recv_s, dud_recv_e, &
     d2u_send_s, d2u_send_e, d2u_recv_s, d2u_recv_e, &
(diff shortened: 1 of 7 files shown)


src/omp/exec_dist.f90
```diff
@@ -72,7 +72,7 @@ subroutine exec_dist_transeq_compact( &
     implicit none
 
     ! du = d(u)
-    real(dp), dimension(:, :, :), intent(out) :: rhs, du, dud, d2u
+    real(dp), dimension(:, :, :), intent(inout) :: rhs_du, dud, d2u
```
Nanoseb, line 75: From what I understand `dud` and `d2u` are still `intent(out)` only.

src/omp/exec_dist.f90
```diff
@@ -61,7 +61,7 @@ subroutine exec_dist_tds_compact( &
   end subroutine exec_dist_tds_compact
 
   subroutine exec_dist_transeq_compact( &
```
Nanoseb, line 63: Maybe add a comment explaining that rhs_du is where the results are also stored.

src/omp/backend.f90
```diff
@@ -233,30 +233,29 @@ subroutine transeq_halo_exchange(self, u, v, w, dir)
 
   end subroutine transeq_halo_exchange
 
-  subroutine transeq_dist_component(self, rhs, u, conv, &
+  subroutine transeq_dist_component(self, rhs_du, u, conv, &
```
Nanoseb, line 236: Same, add a comment explaining that rhs_du is where the output is stored.

src/cuda/exec_dist.f90
```diff
@@ -56,24 +56,24 @@ subroutine exec_dist_tds_compact( &
   end subroutine exec_dist_tds_compact
 
   subroutine exec_dist_transeq_3fused( &
```
Nanoseb, line 58: Add a comment explaining that r_du is where the output is

Nanoseb (changes requested):
Looking good to me beside a minor change and addition of some comments. Overall I think it would be good to document what all the arguments of these functions are supposed to be.

Also, in the issue you mention we might get a 7% performance uplift, is that about what you experienced?

src/cuda/exec_dist.f90
```diff
@@ -56,24 +56,24 @@ subroutine exec_dist_tds_compact( &
...
     du_send_s, du_send_e, du_recv_s, du_recv_e, &
     dud_send_s, dud_send_e, dud_recv_s, dud_recv_e, &
     d2u_send_s, d2u_send_e, d2u_recv_s, d2u_recv_e, &
     der1st, der2nd, nu, nproc, pprev, pnext, blocks, threads &
     )
     implicit none
 
-    ! r_u = -1/2*(v*d1(u) + d1(u*v)) + nu*d2(u)
-    real(dp), device, dimension(:, :, :), intent(out) :: r_u
+    ! r_du = -1/2*(v*d1(u) + d1(u*v)) + nu*d2(u)
```
pbartholomew08, line 68: Is this comment still correct? If so, then I think `r_du` can be `intent(out)` as before?

Also, I think this comment can satisfy @Nanoseb's suggestion, perhaps explicitly drawing attention to this being the output here

src/omp/backend.f90
```diff
@@ -233,30 +233,29 @@ subroutine transeq_halo_exchange(self, u, v, w, dir)
...
                                     u_recv_s, u_recv_e, &
                                     conv_recv_s, conv_recv_e, &
                                     tdsops_du, tdsops_dud, tdsops_d2u, dir)
       !! Computes RHS_x^u following:
       !!
       !! rhs_x^u = -0.5*(conv*du/dx + d(u*conv)/dx) + nu*d2u/dx2
```
pbartholomew08, line 242: And likewise, same the changed `intent` implies the equation described in the comment needs to be updated.

pbartholomew08 (changes requested):
Looks good! I would only suggest some updates to comments to match the code changes

src/omp/exec_dist.f90
semi-h: They're actually all `intent(out)` as before, fixed.

src/cuda/exec_dist.f90
semi-h: Yes, the comment is correct, fixed the `intent`.

semi-h:
Anything else here @Nanoseb @pbartholomew08?

Nanoseb:
Nope, seems ready to go to me

Nanoseb (approved):

pbartholomew08 (approved):
LGTM!

PR #132, opened 2024-12-05
Improve performance of transeq on OpenMP backend

### LINKED ISSUE #66: Performance improvement in transeq in the OpenMP backend
I'm copying my comment in #27 so that we don't forget about it.

https://github.com/xcompact3d/x3d2/blob/2d906a5cafe6060b8021b66aea8ff360558c3968/src/omp/exec_dist.f90#L164-L185

I realised that here we're writing 3 field sized arrays into main memory unnecessarily. It is potentially increasing the runtime %20.

In the second phase of the algorithm here we pass a part of the `du`, `dud`, and `d2u` into der_univ_subs, and they're all rewritten in place. Then later we combine them in `rhs` for the final result. Ideally, we want `du`, `dud`, and `d2u` to be read once and `rhs` to be written only once. However because of the way der_univ_subs work, the updated data in `du` arrays after der_univ_subs call gets written in the main memory, even though we don't need this at all.

There are three ways we can fix this
- In the parallel do loop in the second phase we can copy the relevant parts of `du`, `dud`, and `d2u` arrays into `(SZ, n)` sized temporary arrays. Then we pass temporary arrays into der_univ_subs, and at the end we use these temporaries to obtain final `rhs`. This is the easiest solution but it may not be the best in terms of performance.
- We can write an alternative der_univ_subs and separate input and output arrays. This way we can pass a part of the `du` arrays as we do now, and pass a small temporary array as the output one. Because `du` arrays will be input arrays no data will be written in main memory. Then we can combine the temporaries to get `rhs`.
- If we're writing an alternative der_univ_subs to be used in transeq, we can go one step further and have a fused version of it. This would probably the most performant solution. der_univ_subs is relatively lightweight so it isn't really hard to do so. The new subrotuine can input all `du`, `dud`, and `d2u`, and write the final result `rhs`.

semi-h:
Closes #66, implemented the third option as described in the issue.

### DIFF
--- a/src/omp/exec_dist.f90
+++ b/src/omp/exec_dist.f90
@@ -3,7 +3,8 @@ module m_omp_exec_dist
 
   use m_common, only: dp, VERT
   use m_omp_common, only: SZ
-  use m_omp_kernels_dist, only: der_univ_dist, der_univ_subs
+  use m_omp_kernels_dist, only: der_univ_dist, der_univ_subs, &
+                                der_univ_fused_subs
   use m_tdsops, only: tdsops_t
   use m_omp_sendrecv, only: sendrecv_fields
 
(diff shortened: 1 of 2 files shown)


src/omp/exec_dist.f90
```diff
@@ -164,31 +165,45 @@ subroutine exec_dist_transeq_compact( &
...
+    !    do i = 1, SZ
+    !      rhs_du(i, j, k) = -0.5_dp*(v(i, j, k)*rhs_du(i, j, k) &
+    !                                 + dud(i, j, k)) &
+    !                        + nu*d2u(i, j, k)
+    !    end do
+    !    !$omp end simd
+    !  end do
+
+    !end do
+    !!$omp end parallel do
```
Nanoseb, line 205: If the change is to be kept, this can be deleted

src/omp/kernels/distributed.f90
```diff
@@ -227,4 +227,104 @@ subroutine der_univ_subs(du, recv_u_s, recv_u_e, n, dist_sa, dist_sc)
...
+    real(dp) :: ur, bl, recp
+    real(dp), dimension(SZ) :: du_s, du_e, dud_s, dud_e, d2u_s, d2u_e, &
+                               temp_du, temp_dud, temp_d2u
+
+    !$omp simd
+    do i = 1, SZ
+      ! A small trick we do here is valid for symmetric Toeplitz matrices.
+      ! In our case our matrices satisfy this criteria in the (5:n-4) region
+      ! and as long as a rank has around at least 20 entries the assumptions
+      ! we make here are perfectly valid.
```
Nanoseb, line 262: would not complying with the assumption lead to wrong results or poor performance?
If it is wrong results it would probably be good to actively check for this somewhere (probably higher up in the call stack) and throw an error if the assumptions aren't met.

Nanoseb (changes requested):
looking good. Just a comment on the assumption made for the fused implementation. Thinking about it, maybe if the assumption isn't met we could revert back to the older implementation.

src/omp/kernels/distributed.f90
semi-h: It would lead to wrong results. Created an issue yo deal with this in a dedicated PR later. #133

semi-h:
> looking good. Just a comment on the assumption made for the fused implementation. Thinking about it, maybe if the assumption isn't met we could revert back to the older implementation.

So the assumption is not anything new in the fused implementation, it is because of the distributed algorithm and it would result in identical behaviour with non-fused basic implementation too. @Nanoseb

Nanoseb:
> So the assumption is not anything new in the fused implementation, it is because of the distributed algorithm and it would result in identical behaviour with non-fused basic implementation too. @Nanoseb

Ah ok, sure. Then it sounds good, specially with the issue you created.

Nanoseb (approved):

Issue #97, opened 2024-05-22
Potential misuse of `host_allocator`


semi-h:
#89 introduced a host_allocator concept. It is a way to make sure that we can get a field that lives in host memory when we need. Currently, its used for setting the initial conditions, but I anticipate it'll soon be used for other purposes too.

The open matter on this is that one can accidentally mix and use `get_block` and `release_block` from `backend%allocator` and `solver%host_allocator`. In OpenMP backend this won't cause any immediate issues, because both `allocator`s point to the same object, but this isn't the case in CUDA backend.

So lets discuss the best strategy to prevent such misuses if we can.

One thing I tried was having `field` pointers a type `type(field_t)` instead of `class(field_t)` so that a `get_field` from CUDA allocator will raise an error, it was okay with nvfortran but the gcc compiler didn't like this.

One thing I have in mind is implementing a `get_host_field` function in `allocator`, and using this instead with `host_allocator`. But this can still cause problems when we test things on OpenMP and try to run on CUDA. In the meeting we had a few recommendations but I'm not sure if I understand these, so could you explain here please?

PR #137, opened 2025-01-06
Add Cray support

pbartholomew08:


src/allocator.f90
```diff
@@ -125,12 +142,13 @@ function get_block(self, dir, data_loc) result(handle)
     integer, intent(in) :: dir
     integer, intent(in), optional :: data_loc
     integer :: dims(3)
+
     ! If the list is empty, allocate a new block before returning a
     ! pointer to it.
     if (.not. associated(self%first)) then
       ! Construct a field_t. This effectively allocates
       ! storage space.
-      self%first => self%create_block(next=self%first)
```
semi-h, line 133: It looks like there is a mistake in the original implementation. We don't really need to pass a `next` to `create_block`, because each time we need to allocate a field it means that the linked list is empty and the `next` will always be a null pointer. If we edit `create_block` and use `self%first` in place of receiving a `next` argument and using it in field instantiation it should work fine.

```
newblock = field_t(self%ngrid, self%first, id=self%next_id)
```

I guess the trouble here with Cray compiler is passing a null pointer as argument to `create_block` so this should solve the issue without requiring two new functions. If select type stuff is necessary with Cray then we can have that inside the original `create_block` functions. Each time we create a block we create the first one in the linked list anyways.

pbartholomew08: The issue with the Cray compiler stems from `create_block` taking a `type(field_t)` and the allocator's `first` field being `class(field_t)` - it is technically correct, this seemed the minimal change to me.

However, if as you suggest we don't need to pass `next` then I think this can be resolved more neatly.

semi-h: Actually this select type here is relevant to #97. Maybe using select type will solve the issue there too. My confusion however is that when we instantiate an allocator, there are initially no fields, and then the `first` pointer just points to a null pointer, so not sure how its type could be resolved down to `field_t` or `cuda_field_t`.

pbartholomew08: ah, we're considering two issues here:
1) Cray won't compile the code due to the different types
2) At runtime the first field should indeed be uncertain (I've not been able to test the code yet)

I think your suggestion of removing the `next` argument from `create_block` should resolve both issues

pbartholomew08: I think in the interest of building the code with Cray, I suggest we merge this as the minimum change, and open a new issue to consider changing how `create_block` is designed. What do you think @semi-h ?

semi-h: Sure, it makes sense. I guess this doesn't impact CUDA backend and doesn't cause any issues with gfortran right?

pbartholomew08:
I think this can be merged: Cray (ARCHER2, 1 node:128 ranks) agrees with NVHPC (Grace-Hopper, single GPU) output and shown is comparison with reference data from Dairay2017.

For reference the NVHPC run took 1324.9s, Cray took 5550.8s (both on 256^3 mesh).

nvhpc-cray

PR #145, opened 2025-01-22
Configuration module

semi-h:
This PR introduces a new module for reading user input from namelists. Processing all the parameters from a dedicated module simplifies the process and tidies the code a bit. It is quite similar to `parameters.f90` in Incompact3d.

### DIFF
--- a/src/config.f90
+++ b/src/config.f90
@@ -32,28 +32,28 @@ module m_config
   end type solver_config_t
 
   abstract interface
-    subroutine read(self, file_name, src) !&
+    subroutine read(self, nml_file, nml_string) !&
       !! Assigns the member variables either from a file or text source.
       !!
-      !! file_name can be an absolute or relative path
-      !! src is a character string that contains the namelist definition.
-      !! For example, src="&foobar_nml foo=0, bar='this'/"
+      !! nml_file can be an absolute or relative path
+      !! nml_string is a character string that contains the namelist.
+      !! For example, nml_string="&foobar_nml foo=0, bar='this'/"
       import :: base_config_t
 
       class(base_config_t) :: self
-      character(*), optional, intent(in) :: file_name
-      character(*), optional, intent(in) :: src
+      character(*), optional, intent(in) :: nml_file
+      character(*), optional, intent(in) :: nml_string
     end subroutine read
   end interface
 
 contains
 
-  subroutine read_domain_nml(self, file_name, src)
+  subroutine read_domain_nml(self, nml_file, nml_string)
     implicit none
 
     class(domain_config_t) :: self
-    character(*), optional, intent(in) :: file_name
-    character(*), optional, intent(in) :: src
+    character(*), optional, intent(in) :: nml_file
+    character(*), optional, intent(in) :: nml_string
 
     integer :: unit
 
@@ -66,15 +66,15 @@ contains
     namelist /domain_settings/ flow_case_name, L_global, dims_global, &
       nproc_dir, BC_x, BC_y, BC_z
 
-    if (present(file_name) .and. present(src)) then
+    if (present(nml_file) .and. present(nml_string)) then
       error stop 'Reading domain config failed! &
                  &Provide only a file name or source, not both.'
-    else if (present(file_name)) then
-      open (newunit=unit, file=file_name)
+    else if (present(nml_file)) then
+      open (newunit=unit, file=nml_file)
       read (unit, nml=domain_settings)
       close (unit)
-    else if (present(src)) then
-      read (src, nml=domain_settings)
+    else if (present(nml_string)) then
+      read (nml_string, nml=domain_settings)
     else
       error stop 'Reading domain config failed! &
                  &Provide at least one of the following: file name or source'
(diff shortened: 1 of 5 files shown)


src/config.f90
```diff
@@ -0,0 +1,130 @@
...
+
+      class(base_config_t) :: self
+      character(*), optional, intent(in) :: file_name
+      character(*), optional, intent(in) :: src
+    end subroutine read
+  end interface
+
+contains
+
+  subroutine read_domain_nml(self, file_name, src)
```
Nanoseb, line 46: just reading the definition it isn't clear what src is compared to file_name. So I'd add some comments here

src/solver.f90
```diff
@@ -111,35 +109,44 @@ function init(backend, mesh, host_allocator) result(solver)
     solver%v => solver%backend%allocator%get_block(DIR_X)
     solver%w => solver%backend%allocator%get_block(DIR_X)
 
-    call read_solver_input(Re, dt, n_iters, n_output, poisson_solver_type, &
-                           time_intg, der1st_scheme, der2nd_scheme, &
-                           interpl_scheme, stagder_scheme)
+    if (command_argument_count() >= 1) then
+      call get_command_argument(1, input_file)
```
Nanoseb, line 113: I think this logic of reading command line argument should be defined elsewhere (maybe inside the config module?) because the code is duplicated also in `xcompact.f90` at the moment. Right now it isn't a big deal but we might be something to think about for when we want to add other command line arguments, a help page etc.

Nanoseb (changes requested):
This is great, I was thinking it would be nice to have something like that when I wrote my `read_solver_input` function. Just a couple of minor comments

src/config.f90
```diff
@@ -0,0 +1,130 @@
...
+    namelist /domain_settings/ flow_case_name, L_global, dims_global, &
+      nproc_dir, BC_x, BC_y, BC_z
+
+    if (present(file_name) .and. present(src)) then
+      error stop 'Reading domain config failed! &
+                 &Provide only a file name or source, not both.'
+    else if (present(file_name)) then
+      open (100, file=file_name)
+      read (100, nml=domain_settings)
+      close (100)
```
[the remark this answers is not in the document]
semi-h: This is nice, thanks! All done

src/config.f90
semi-h: Added comments for the interface read_nml function to cover both

src/solver.f90
semi-h: Completely agree, not the best place. In general I think its better to avoid getting command arguments inside a module instead of from a program. For example test programs can still use these modules but in addition they may have their own stuff from the command line input, it would be hard to manage. We may instead read the solver namelist in the main program (we actually already do to get poisson solver) and pass the `solver_config_t` instance to solver, but not sure how this'll work when we have even more options/namelists. We may also deal with this later on and not in this PR.

semi-h:
@pbartholomew08 would be nice to have your review on this if you have time

semi-h:
Something interesting I observed is that it is possible to update a subset of the member arguments via the `src` option after reading from file. For example we can read solver config from file, then just change the number of iterations
```
call solver_cfg%read(file_name=input_file)
call solver_cfg%read(src='&solver_params n_iters=1/')
```
This only changes the n_iters member of the `solver_cfg`, doesn't mess up with all the variables that are previously set. Tested both on gfortran and nvfortran but not sure if this is standard or a typical behaviour of Fortran compilers.

This can be super useful when combined with command line argument support
https://fortranwiki.org/fortran/show/get_namelist

Then we can run solver like
```
xcompact ./config_nml.x3d n_iters=1,dt=0.01
``` 
to change a few options without editing the namelist file to test things quickly.

src/solver.f90
pbartholomew08: How about replacing this `if ...` with `call get_config()` which then runs this code? This could be defined in the `configuration` module and used/not used as required.

src/config.f90
```diff
@@ -0,0 +1,139 @@
...
+      !! Assigns the member variables either from a file or text source.
+      !!
+      !! file_name can be an absolute or relative path
+      !! src is a character string that contains the namelist definition.
+      !! For example, src="&foobar_nml foo=0, bar='this'/"
+      import :: base_config_t
+
+      class(base_config_t) :: self
+      character(*), optional, intent(in) :: file_name
+      character(*), optional, intent(in) :: src
```
pbartholomew08, line 45: Following the comments below - if we rename `src -> nml` then I think this would make it clear that it refers to the namelist within the file.

src/config.f90
```diff
@@ -0,0 +1,139 @@
...
+
+    class(solver_config_t) :: self
+    character(*), optional, intent(in) :: file_name
+    character(*), optional, intent(in) :: src
+
+    integer :: unit
+
+    real(dp) :: Re, dt
+    integer :: n_iters, n_output
+    character(3) :: poisson_solver_type = 'FFT', time_intg
```
pbartholomew08, line 104: Split these as they are unrelated, i.e.
```
character(3) :: poisson_solver_type = 'FFT'
character(3) :: time_intg
```
also makes it clearer that `poisson_solver_type` has a default value

pbartholomew08 (changes requested):
I think this really improves the code, have suggested a few minor changes

src/config.f90
semi-h: `read(nml, nml=solver)` would look ugly so renamed to `nml_string` instead, and for consistency renamed `file_name -> nml_file`.

pbartholomew08 (approved):

src/solver.f90
pbartholomew08: I think this addresses both concerns?

Nanoseb (approved):

PR #147, opened 2025-01-28
Issue error for formatting issues

ia267:
This PR will recover the previous behaviour that will issue an error during a PR if there are Formatting issues from fprettify (instead of just issuing a warning).  A meaningful error message is issued.

before
before

after
after


semi-h (approved):
Great to have this back!

PR #148, opened 2025-01-31
Make changes to contribution docs

ia267:
Contribution docs have been updated.
- Removed strict enforcing of conventional commits and removed commit-msg githook. Instead we are focusing on writing commit messages that's short and human-readable.
- Pre-commit checks with fprettify is made optional - however, fprettify checks will be done during a pull request and will lead to an error if incorrectly formatted
- Added section on rebasing feature branches on top of repo's main to keep commit history linear
- Made minor changes to the docs for readability, contribution process remains the same as before: we work with issues, submit PRs, assign for review (changed reviewers to 1 instead of 2).

see [RTD docs ](https://ia267-x3d2.readthedocs.io/en/chore-collaboration-docs/developer/index.html)with these changes:


docs/source/developer/contributing_guidelines.rst
```diff
@@ -1,145 +1,98 @@
...
+
+.. code-block:: bash
+
+    # Add the resolved files
+    git add .
+
+    # Continue the rebase
+    git rebase --continue
+
+* Squash commits before merging: If your feature branch has multiple commits, squash them into a single commit before merging. This keeps the commit history clean and concise.
```
semi-h, line 98: Sometimes PRs are large and then I think merging the changes keeping the commit history (so no squash) is helpful provided that:
- Individual commits represent a meaningful, complete and self-consistent change in the code
- At no point the code is broken (as in if a commit introduced a bug, there shouldn't be a new commit in the PR fixing the bug, instead, the commit introduced the bug should be edited and fixed)

I think following these points leads to a clean commit history.

Of course people should carry on with the way they prefer to work, so our guideline can be something like if the above recommendations are followed then the PR can be merged keeping the commit history, and if not squashed and merged so that we have a clean history.

Also, we can mention the linear history requirement here. I think we can set up some rulesets to prevent merges that would violate linear history so its not broken by mistake.

docs/source/developer/style.rst
```diff
@@ -4,108 +4,135 @@ Fortran style guide
...
-  characters each argument is defined on it own line.
+- Procedure prototypes spanning more than 79 characters should be split before the first dummy argument is defined, after the opening parenthesis. If the list of dummy arguments spans more than 79 characters, each argument should be defined on its own line.
 
   .. code:: fortran
 
      subroutine my_long_subroutine( &
-        argument1, argument3, argument4, argument5, argument5, argument6 &
-	)
+        argument1, argument2, argument3, argument4, argument5, argument6 &
+     )
```
semi-h, line 45: Although I prefer this way and find it prettier, `fprettify` raises a warning if the parenthesis is not indented together with the block above.

docs/source/developer/style.rst
```diff
@@ -4,108 +4,135 @@ Fortran style guide
...
-  return object is an array.
-- Functions are always defined with the `pure` prefix.  A procedure
-  with side effects must be a `subroutine`.
+        argument2, &
+        argument3, &
+        argument4, &
+        argument5, &
+        argument6, &
+        argument7 &
+     )
```
semi-h, line 55: we don't really follow this

docs/source/developer/style.rst
```diff
@@ -4,108 +4,135 @@ Fortran style guide
...
+        procedure :: some_procedure
+     end type stencil_t
 
 Custom structure constructors
 -----------------------------
 
-- Are named like `create_<type_root_name>[_<suffix>]`
-- Are declared with the `private` attribute
-- Are defined at the top of the module's `contains` block.
+- Name constructors as ``create_<type_root_name>[_<suffix>]``.
```
semi-h, line 124: I think we mostly name them `init` or `<type_root_name>_init` instead.

docs/source/developer/style.rst
```diff
@@ -120,46 +147,53 @@ Example
...
+       !! This is the first paragraph of the procedure's
+       !! documentation. Note that it starts with TWO !.
+       !! The addition operation is defined in-line as \( c = a + b \).
+       !! 
+       !! The following operation shows it as a displayed equation:
+       !! $$c = a + b$$
+       !! 
        real, intent(in) :: a, b !! Optional documentation for dummy argument.
-       real, intent(out) :: c !! The result of a + b
+       real, intent(out) :: c !! The result of \( a + b \)
```
semi-h, line 194: We can give an example to comment on the preceding line
```
!> The result of \( a + b )\
real, intent(out) :: c
```
produces same documentation as `!!` at the end of line. Useful when the definition is too long.

docs/source/report_a_bug.rst
```diff
@@ -1,11 +1,12 @@
...
 
-You can do so by adding a new issue to the x3d2 issue tracker on
-GitHub. See `(GitHub docs) Creating an issue <https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue>`_.
+To report a bug, please add a new issue to the x3d2 issue tracker on GitHub. For detailed instructions, see the `(GitHub docs) Creating an issue <https://docs.github.com/en/issues/tracking-your-work-with-issues/creating-an-issue>`_.
 
-If you are reporting a bug, be sure to include
+When reporting a bug, be sure to include the following information:
 
-- The version of x3d2 you are using
+- The version of x3d2 you are using.
```
semi-h, line 8: Commit hash can be more useful. (as long as we enforce a linear history)

semi-h (commented):
some minor comments

docs/source/developer/contributing_guidelines.rst
ia267: Done - yes I think we should set-up some ruleset to enforce rebasing for example

docs/source/developer/style.rst
ia267: Done

PR #149, opened 2025-02-03
Fix a bug with staggared 'v2p' operations

semi-h:
I observed this bug when working on the channel flow case. It is related to the tridiagonal solver operations, and surfaces when we do a staggared 'v2p' operation with non-periodic domain BCs.

'v2p' is unique because the resulting field length `n` is smaller than the input length. This was causing an issue with the right-hand-side build phase. The very last `cell` entry in a `v2p` operation requires the `next` vertex so that the corresponding right-hand-side can be built correctly. I made a logical bug here with the `tdsops` operator in `tdsops.f90` around line ~640. `coeff_e(6, 4)` tries to access the `next` vertex but the hardcoded implementation accesses the ghost cell instead because based on the tridiagonal solver length `n` it corresponds to a vertex outside of the domain.

The fix introduces a new lenght `n_rhs` to correctly build the right-hand-side in this particular setting. In the majority of the cases `n_tds` is equal to `n_rhs`, when the setup requires a different `n_rhs`, `tdsops` class figures out the correct dimension.

Added a test that would fail without the fix to make sure its all good. The previous test was passing mainly due to lack and two errors cancelling each other. It used a cos wave from 0 to 2pi and the ghost access I explained above simply didn't cause an issue with a periodic signal.


src/tdsops.f90
```diff
@@ -56,30 +56,31 @@ module m_tdsops
...
-      !! function (symmetric) accross the boundary. If it is .false. it means
-      !! that the field is assumed to be an odd function (anti-symmetric).
-      !! 'c_nu', 'nu0_nu' are relevant when operation is second order
-      !! derivative and scheme is compact6-hyperviscous.
+  function tdsops_init(n_tds, delta, operation, scheme, &
+                       bc_start, bc_end, n_halo, from_to, sym, c_nu, nu0_nu) &
+    result(tdsops)
+    !! Constructor function for the tdsops_t class.
+    !!
+    !! 'n_tds', 'n_rhs', 'delta', 'operation', and 'scheme' are necessary
```
pbartholomew08, line 64: `n_rhs` doesn't appear in the arguments list here - and it will be important to document (somewhere) the difference between `n_tds` and `n_rhs`.

src/tdsops.f90
```diff
@@ -88,9 +89,23 @@ function tdsops_init(tds_n, delta, operation, scheme, bc_start, bc_end, &
...
 
-    integer :: n_stencil
+    integer :: n, n_stencil
+
+    tdsops%n_tds = n_tds
 
-    tdsops%tds_n = tds_n
+    ! we need special treatment in the right-hand-side build stage for
+    ! the very last point in the domain if output length is smaller than
+    ! the input length
```
pbartholomew08, line 98: To check my understanding: `n_tds` is the number of equations to solve, `n_rhs` is the total number of points in the RHS stencils?

src/tdsops.f90
```diff
@@ -88,9 +89,23 @@ function tdsops_init(tds_n, delta, operation, scheme, bc_start, bc_end, &
...
+    if (present(from_to)) then
+      if ((bc_end == BC_NEUMANN .or. bc_end == BC_DIRICHLET) &
```
pbartholomew08, line 100: Does it not also apply in the case of `bc_start`?

pbartholomew08 (changes requested):
The documentation comment (or related subroutine call) requires updating

src/tdsops.f90
semi-h: Fixed, forgot to remove this after changing my approach slightly.

src/tdsops.f90
semi-h: Yes, provided an explanation in the `tdsops_t` class where they're declared as class members

src/tdsops.f90
semi-h: Nope, its only at the end. from v to p we have one less data point and we always deal with it at the very end.

semi-h:
Just to add to your question @pbartholomew08, the difference between `n_tds` and `n_rhs` comes down to the loop ranges in the RHS build phase, forward pass, and backward pass. The implementations both for the DistD2 algorithm and Thomas algorithm fuses the RHS build and forward pass. Therefore we need the larger length of the input and output fields in the RHS build phase. The forward pass operation at the very end is simply discarded. In the backward pass we use `n_tds` to make sure we solve the system correctly.

So an alternative would be creating `tdsops` instances as
```
tdsops  = tdsops_t(n_output, n_input, delta, ...)
```
Then in the `init` function in `tdsops_t` we set `n_rhs` to the larger one of the `n_output` and `n_input`.
At the solver level `n_output` and `n_input` are always one of `VERT` or `CELL` dim along the direction of the operation.

semi-h:
Just added two more tests for the `p2v` operations to make sure they're all good after the changes.

src/tdsops.f90
```diff
@@ -755,18 +753,14 @@ subroutine stagder_1st(self, delta, scheme, from_to, bc_start, bc_end, sym)
       case ('v2p')
         ! sym is always .false.
         dist_b(n) = 1._dp + alpha
```
ia267, line 755: Shouldn't this now be `dist_b(self%n_rhs) = 1._dp + alpha`?

tests/omp/test_omp_transeq.f90
```diff
@@ -3,7 +3,7 @@ program test_omp_transeq
   use mpi
 
   use m_allocator, only: allocator_t, field_t
-  use m_common, only: dp, pi, DIR_X, DIR_Y, DIR_Z, VERT
+  use m_common, only: dp, pi, DIR_X, DIR_Y, DIR_Z, VERT, BC_PERIODIC
```
ia267, line 6: Remove unnecessary `BC_PERIODIC`

ia267 (approved):
Approved

tests/omp/test_omp_transeq.f90
semi-h: Thanks, all done

pbartholomew08 (approved):

PR #153, opened 2025-02-06
Backend functionality for getting max and mean values of a field

semi-h:
Kind of continuation of #144. Back then relied on moving things back to CPU and doing some simple operations and now adding proper support in both backends.

Adds a new function `field_max_mean` to get maximum and mean values in field, takes care of the padding and considers only the actual region in a padded field. Also improved the existing CUDA scalar_product to support padded configurations.

@pbartholomew08, I noticed that the `scalar_product` in OpenMP backend supports `DIR_C` layout as well, do you ever call `scalar_product` on `DIR_C` arrays or just added this in case? On GPU supporting `DIR_C` would be a bit nasty, for now I just terminate the code if the field is `DIR_C` but if you need this for your stuff we need to find a solution later.

@ia267, added a FORD description to this new function `field_max_mean` I added in the `base_backend`, then simply referred to this description in the CUDA and OpenMP extensions. Otherwise it would be too much duplication, what do you think?

I tested these two functions by running a TGV case first with default input.x3d file to make sure divU (calls max_mean) and enstrophy (calls scalar_product) progresses correctly as before. Then I did another test this time with global dimensions that require padding in all directions. I can confirm that both div U and enstrophy looks correct both on OpenMP and CUDA backends.

### DIFF
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -618,6 +618,7 @@ contains
   end subroutine vecadd_cuda
 
   real(dp) function scalar_product_cuda(self, x, y) result(s)
+    !! [[m_base_backend(module):scalar_product(function)]]
     implicit none
 
     class(cuda_backend_t) :: self
@@ -625,8 +626,15 @@ contains
 
     real(dp), device, pointer, dimension(:, :, :) :: x_d, y_d
     real(dp), device, allocatable :: sum_d
+    integer :: dims(3), dims_padded(3), n, n_i, n_i_pad, n_j, ierr
     type(dim3) :: blocks, threads
-    integer :: n, ierr
+
+    if ((x%data_loc == NULL_LOC) .or. (y%data_loc == NULL_LOC)) then
+      error stop "You must set the data_loc before calling scalar product"
+    end if
+    if ((x%data_loc /= y%data_loc) .or. (x%dir /= y%dir)) then
+      error stop "Called scalar product with incompatible fields"
+    end if
 
     call resolve_field_t(x_d, x)
     call resolve_field_t(y_d, y)
@@ -634,10 +642,23 @@ contains
     allocate (sum_d)
     sum_d = 0._dp
 
-    n = size(x_d, dim=2)
-    blocks = dim3(size(x_d, dim=3), 1, 1)
+    dims = self%mesh%get_dims(x%data_loc)
+    dims_padded = self%mesh%get_padded_dims(DIR_C)
+
+    if (x%dir == DIR_X) then
+      n = dims(1); n_j = dims(2); n_i = dims(3); n_i_pad = dims_padded(3)
+    else if (x%dir == DIR_Y) then
+      n = dims(2); n_j = dims(1); n_i = dims(3); n_i_pad = dims_padded(3)
+    else if (x%dir == DIR_Z) then
+      n = dims(3); n_j = dims(2); n_i = dims(1); n_i_pad = dims_padded(1)
+    else
+      error stop 'scalar_product_cuda does not support DIR_C fields!'
+    end if
+
+    blocks = dim3(n_i, (n_j - 1)/SZ + 1, 1)
     threads = dim3(SZ, 1, 1)
-    call scalar_product<<<blocks, threads>>>(sum_d, x_d, y_d, n) !&
+    call scalar_product<<<blocks, threads>>>(sum_d, x_d, y_d, & !&
+                                             n, n_i_pad, n_j)
 
     s = sum_d
 
--- a/src/cuda/backend.f90
+++ b/src/cuda/backend.f90
@@ -21,7 +21,7 @@ module m_cuda_backend
   use m_cuda_tdsops, only: cuda_tdsops_t
   use m_cuda_kernels_dist, only: transeq_3fused_dist, transeq_3fused_subs
   use m_cuda_kernels_fieldops, only: axpby, buffer_copy, field_scale, &
-                                     field_shift, scalar_product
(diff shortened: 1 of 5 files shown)


ia267:
> @ia267, added a FORD description to this new function `field_max_mean` I added in the `base_backend`, then simply referred to this description in the CUDA and OpenMP extensions. Otherwise it would be too much duplication, what do you think?

Agreed - this is what we should be doing

src/cuda/backend.f90
```diff
@@ -666,6 +688,67 @@ subroutine copy_into_buffers(u_send_s_dev, u_send_e_dev, u_dev, n)
...
+    blocks = dim3(n_i, (n_j - 1)/SZ + 1, 1)
+    threads = dim3(SZ, 1, 1)
+    call field_max_sum<<<blocks, threads>>>(max_d, sum_d, f_d, & !&
+                                            n, n_i_pad, n_j)
+
+    max_p = max_d
+    sum_p = sum_d
+    mean_p = sum_p/product(self%mesh%get_global_dims(data_loc))
+
+    call MPI_Reduce(max_p, max_val, 1, MPI_DOUBLE_PRECISION, MPI_MAX, 0, &
```
ia267, line 745: In function `scalar_product_cuda` we are using `MPI_Allreduce`. Could you provide some context on why `MPI_Reduce` was chosen here?

src/cuda/kernels/fieldops.f90
```diff
@@ -79,25 +79,58 @@ attributes(global) subroutine field_shift(f, const, n)
...
+    max_pncl = 0._dp
+    sum_pncl = 0._dp
+    do j = 1, n
+      if (i + (b_j - 1)*blockDim%x <= n_j) then
+        val = abs(f(i, j, b))
+        sum_pncl = sum_pncl + val
+        max_pncl = max(max_pncl, val)
+      end if
+    end do
+    ierr = atomicadd(sum_f, sum_pncl)
```
ia267, line 131: The return value of `atomicadd` and `atomicmax` are set to `ierr`. I see NVIDIA examples using `istat`. But could we just use `call atomicadd` instead as we're not using the return value in `ierr`?

src/omp/backend.f90
```diff
@@ -549,6 +548,76 @@ subroutine copy_into_buffers(u_send_s, u_send_e, u, n, n_groups)
...
+    real(dp), intent(out) :: max_val, mean_val
+    class(field_t), intent(in) :: f
+    integer, optional, intent(in) :: enforced_data_loc
+
+    real(dp) :: val, max_p, sum_p, mean_p, max_pncl, sum_pncl
+    integer :: data_loc, dims(3), dims_padded(3), n, n_i, n_i_pad, n_j
+    integer :: i, j, k, k_i, k_j, ierr
+
+    if (f%data_loc == NULL_LOC .and. (.not. present(enforced_data_loc))) then
+      error stop 'The input field to field_max_mean does not have a valid &
```
ia267, line 565: Could we make the error message more backend-specfic? For example:

`[CUDA] The input field to field_max_mean...` and `[OpenMP] The input field to field_max_mean`.

src/omp/backend.f90
```diff
@@ -549,6 +548,76 @@ subroutine copy_into_buffers(u_send_s, u_send_e, u, n, n_groups)
...
+      n = dims(3); n_j = dims(1); n_i = dims(2); n_i_pad = dims_padded(2)
+    else
+      error stop 'field_max_mean does not support DIR_C fields!'
+    end if
+
+    sum_p = 0._dp
+    max_p = 0._dp
+    !$omp parallel do collapse(2) reduction(+:sum_p) reduction(max:max_p) &
+    !$omp private(k, val, sum_pncl, max_pncl)
+    do k_j = 1, (n_j - 1)/SZ + 1
```
ia267, line 593: For readability, could we create a new integer variable say `n_blocks_j` and set `n_blocks_j = (n_j - 1)/SZ + 1`?

src/cuda/kernels/fieldops.f90
```diff
@@ -79,25 +79,58 @@ attributes(global) subroutine field_shift(f, const, n)
...
     i = threadIdx%x
-    b = blockIdx%x
+    b_i = blockIdx%x
+    b_j = blockIdx%y
 
+    b = b_i + (b_j - 1)*n_i_pad
     s_pncl = 0._dp
     do j = 1, n
-      s_pncl = s_pncl + x(i, j, b)*y(i, j, b)
+      if (i + (b_j - 1)*blockDim%x <= n_j) then
```
ia267, line 99: Could we move this boundary check outside the loop?

semi-h: Definitely can, thanks. Did it both for max_mean and scalar_product.

src/cuda/backend.f90
semi-h: My initial thought was that `scalar_product` is a type of operation where when required all the ranks need to know about the final value, but `max_mean` is more like a debugging/reporting thing and these values are not really used as part of any further calculations. But the overhead between Reduce vs Allreduce is probably doesn't matter and this is not a function called very often so just changed these Reduce calls to Allreduce.

src/omp/backend.f90
semi-h: Is quite common practice to do basic arithmetic for the loop ranges in place. How about the in-line comment I added for clarification?

semi-h:
Edited the top comment to explain how I ensured that the added functions are behaving as expected for the record.

ia267 (approved):
Approved

PR #157, opened 2025-02-13
Resolve calling set field data drops fields data location when it reorders

pbartholomew08:


semi-h (approved):
The fix is simple and nice. My only worry with the tests like this is that they may become a drag for minor changes in the future. Maybe later on we can group procedures in a class together in a test program if we end up with too many.

Interesting that unit tests didn't catch the bug in the test but gitlab did. Probably failing because `arr` is not allocated.

tests/test_setget_field.f90
```diff
@@ -58,13 +54,19 @@ program test_setget_field
 #endif
   
   fld => backend%allocator%get_block(DIR_X, VERT)
+  fld_c => backend%allocator%get_block(DIR_C, VERT)
+  allocate(arr, mold=fld_c%data)
```
semi-h, line 58: This will fail on the CUDA backend. We should't access a `field`s data member unless we're certain that it's data stored on the host. You can instead use `field%get_shape` to get back a length 3 integer array with the correct dimensions.

tests/test_setget_field.f90
```diff
@@ -28,7 +28,7 @@ program test_setget_field
 #endif
   type(mesh_t) :: mesh
   
-  class(field_t), pointer :: fld
+  class(field_t), pointer :: fld, fld_c
   real(dp), dimension(:, :, :), allocatable :: arr
 
   integer :: nx, ny, nz
```
semi-h, line 34: Also suprised that this doesn't throw a unused variable warning and tests are passing

pbartholomew08: Done

tests/test_setget_field.f90
pbartholomew08: Done

Issue #12, opened 2023-10-24
Support for stretched meshes


semi-h:
Incompact3D supports stretched meshes in y-direction. My understanding is it is pretty straightforward. After obtaining the solution from the regular tridiagonal solver, every single line in the field along y direction gets pointwise-multiplied by an array to factor in the stretching in the final result.

I plan to add support for stretchinig by passing this array of stretching factors in the second phase of the distributed algorithm by fusing this operation in the final substitution phase.
https://github.com/xcompact3d/x3d2/blob/main/src/cuda/kernels_dist.f90#L145-L149
So when this support is added even when there is no stretching we'll be multipying the final result by an array of ones, but I think that in terms of performance there won't be a meaningful difference when the stretching support is added.

PR #163, opened 2025-02-26
 Stretching support for tridiagonal solvers

semi-h:
closing #12 

- [x] Add stretching stuff in mesh%geo
- [x] Add stretching in `tdsops_t`
- [x] Update tridiagonal solvers with stretching support

@slaizet, I've checked that all stretching types, centred, both-ends, and bottom, are giving a reasonable point distribution. If you have any suggestions or want any changes in the implementation let me know. I followed Incompact3d stretching implementation and it looks a bit different than the formulation in the 2009 paper.

Best way to move forward is merging #136 first and then this one. Only then I can start working on stretching support for the FFT stuff.

### DIFF
--- a/src/solver.f90
+++ b/src/solver.f90
@@ -165,7 +165,7 @@ contains
     character(*), intent(in) :: der1st_scheme, der2nd_scheme, &
                                 interpl_scheme, stagder_scheme
 
-    integer :: dir, bc_start, bc_end, n_vert, n_cell
+    integer :: dir, bc_start, bc_end, n_vert, n_cell, i
     real(dp) :: d
 
     dir = dirps%dir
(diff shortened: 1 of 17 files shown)


src/solver.f90
```diff
@@ -175,22 +175,40 @@ subroutine allocate_tdsops(dirps, backend, mesh, der1st_scheme, &
...
+      bc_end, from_to='p2v', stretch=mesh%geo%vert_ds(1:n_vert, dir) &
+      )
+    call backend%alloc_tdsops( &
+      dirps%interpl_v2p, n_cell, d, 'interpolate', interpl_scheme, &
+      bc_start, bc_end, from_to='v2p', stretch=[(1._dp, i=1, n_cell)] &
+      )
+    call backend%alloc_tdsops( &
+      dirps%interpl_p2v, n_vert, d, 'interpolate', interpl_scheme, &
+      bc_start, bc_end, from_to='p2v', stretch=[(1._dp, i=1, n_vert)] &
+      )
```
pbartholomew08, line 211: Should these not be using the midpoints?

semi-h: That's what I also thought initially but compact interpolation don't correct the result based on stretching in Incompact3d https://github.com/xcompact3d/Incompact3d/blob/06af91a0713a362af42ae7b04d491260d31b70ea/src/derive.f90#L4438.

semi-h:
As a final thing updated get_coordinates function in mesh.f90 to support stretched coordinates

PR #170, opened 2025-04-15
Add y-stretching support in Poisson-FFT

semi-h:


semi-h:
Fixed a few bugs and now I can maintain a uniform parabolic profile with a stretched mesh. This shows that the pentadiagonal coefficients are correct and the pentadiagonal solver in the spectral space works okay.

semi-h:
@slaizet I checked the pentadiagonal coefficients for `bottom` type stretching with the data you sent me and I confirm that they're identical. The complex modes before and after the pentadiagonal solver are also identical to Incompact3d data.
Then just to make sure I run a channel case with stretching for 100k time steps and it looks completely fine.

semi-h:
@slaizet I can edit the channel case input file here so that we have the identical setup with Incompact3d, is that okay?

examples/channel/input.x3d
```diff
@@ -15,6 +15,10 @@ nproc_dir = 1, 1, 1
 BC_x = 'periodic', 'periodic'
 BC_y = 'dirichlet', 'dirichlet'
 BC_z = 'periodic', 'periodic'
+
+! Stretching options: 'uniform' (default) | 'centred' | 'both-ends' | 'bottom'
+stretching(2) = 'both-ends'
```
slaizet, line 20: can we maybe call "both-ends" "top-bottom" to be consistent with "bottom"

slaizet:
yes for having the same input file as for xcompact3d (https://github.com/xcompact3d/Incompact3d/blob/master/examples/Channel/input_DNS_Re180_LR_explicittime.i3d)

Make it clear that there is no "distributed" capability for the stretched Poisson block (or add an item in the to-do list)

All good to go otherwise

slaizet:
All good to go!!

PR #179, opened 2025-05-05
Enable backend selection independently of compiler

### LINKED ISSUE #178: Enable user selection of backend
**Description:** Currently the backend is selected based on the detected compiler: nvfort -> GPU/CUDAFortran; other -> CPU/OMP. If we add a GPU/OMP backend this should be supported by nvfort (which could also support CPU/OMP too), therefore the user should be able to select the backend independently of the compiler.

**Requirements:**

1. Invalid backend/compiler combinations should be rejected (primarily only nvfort supports GPU/CUDAFortran)
2. The defaults should be preserved (nvfort->GPU/CUDAFortran, other->CPU/OMP)

pbartholomew08:
Closes #178 

Currently the backend is fixed based on the compiler. As we add backends the user should be able to select between them, for example currently we could have

NVHPC: CUDA/GPU or OMP/CPU
Other: OMP/CPU

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -7,3 +7,6 @@ set(WITH_2DECOMPFFT ON CACHE BOOL
 
 add_subdirectory(src)
 add_subdirectory(tests)
+
+list(APPEND CMAKE_MODULE_PATH "${CMAKE_CURRENT_SOURCE_DIR}/cmake")
+include(SelectBackend)
--- a/cmake/SelectBackend.cmake
+++ b/cmake/SelectBackend.cmake
@@ -4,12 +4,22 @@
 # - GPU/CUDAFortran (nvfort only)
 # - CPU/OMP (all)
 
+set(X3D2_BACKEND_OPTIONS "OMP/CPU")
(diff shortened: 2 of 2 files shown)


semi-h:
I think it'll be better to specify the backend at runtime rather than at compile time. Right now we don't have this functionality but it is trivial to add this in the main xcompact.f90 file (For example "mpirun -np 2 x3d2 --backend cuda input.x3d"). Runtime specification of backend is the only way to allow heterogeneous runs, and we can't rely on CMake level decisions for heterogeneous runs ("mpirun -np 1 x3d2 --backend cuda input.x3d :  -np 1 x3d2 --backend omp input.x3d2" so that one rank runs on CPU and the other on GPU, and for this we need both backend/omp and backend/cuda to be compiled and available). Also for example with the current CUDA compilers we run the OpenMP backend unit tests as well, it wouldn't be possible if we omitted OpenMP backend source files by specifying CUDA backend at compile time.

So the idea is from the CMake we allow the present compiler to compile all the parts of the codebase it can, then at runtime we decide which specific backend we want to go ahead.

CFD-Xing:
I think we can close this since #370 solved this issue.

PR #181, opened 2025-06-02
Add an explicit test for backend%vecadd

pbartholomew08:
Although the time integrators test exercises the `vecadd` functionality, it is mixed with other functions. This test will ensure issues can be isolated to `vecadd` if they arise.


tests/test_vecadd.f90
```diff
@@ -0,0 +1,167 @@
...
+    call backend%vecadd(1.0_dp, a, 1.0_dp, r)
+    if (any(r%data /= a%data)) then
+      print *, dirnames(d), ": a + 0 = a FAILED"
+      test_pass = .false.
+    end if
+    
+    ! a + a = 2a
+    r%data = a%data
+    call backend%vecadd(1.0_dp, a, 1.0_dp, r)
+    if (any(r%data /= 2 * a%data)) then
```
semi-h, line 92: `%data` is a host array and in `cuda_field_t` we have `%data_d` specifically for device storage. Also, I don't think CUDA Fortran allows operations like `a_dev /= 2*b_dev`, the CUDA Fortran programming guide describes such operations as ambigious as its not clear where the `2*` will take place.

semi-h (commented):
Implementation is not compatible with CUDA in its current state. See https://github.com/xcompact3d/x3d2/blob/3627692b30e91c4d6c152506eef142da63f4c9db/src/case/base_case.f90#L97 for how we handle setting fields on both OpenMP and CUDA backends. A similar strategy can also handle comparisons you need here.

semi-h:
Maybe this is more relevant for initialising with random values
https://github.com/xcompact3d/x3d2/blob/3627692b30e91c4d6c152506eef142da63f4c9db/src/case/channel.f90#L65

pbartholomew08:
This test uncovered a possible bug in the OMP backend (see #182)

semi-h (approved):

PR #185, opened 2025-06-13
Add veccopy backend

CFD-Xing:
Multi step/stage time-integration methods often required deep copy of solution vectors at previous step/stage. Currently this is emulated by using vecadd by setting the coefficients to 1.0 and 0.0. A more effective way is to provide a veccopy backend.


semi-h (approved):
Thanks!

PR #188, opened 2025-06-18
Create a general interface for the time integrator

### LINKED ISSUE #92: Create a general interface for the time integrator
Currently, the time integrator takes seven arguments as follow

`time_integrator%step(u, v, w, du, dv, dw, dt)`

We should consider instead to create a general interface taking `flist_t` as arguments such that the time integrator can be used for an arbitrary number of variables

`time_integrator%step(sol, deriv, dt)`

This will also simply the addition of unit test for the time integrator

EDIT:

We should also consider to generalize the `method` member variable to consider various type of integration techniques (AB, RK, IMEX, etc.)

CFD-Xing:
Closes #92

### DIFF
--- a/src/case/base_case.f90
+++ b/src/case/base_case.f90
@@ -6,7 +6,7 @@ module m_base_case
   use m_allocator, only: allocator_t
   use m_base_backend, only: base_backend_t
   use m_common, only: dp, DIR_X, DIR_Z, DIR_C, VERT
-  use m_field, only: field_t
+  use m_field, only: field_t, flist_t
   use m_mesh, only: mesh_t
   use m_solver, only: solver_t, init
 
@@ -191,7 +191,8 @@ contains
 
     class(base_case_t), intent(inout) :: self
 
-    class(field_t), pointer :: du, dv, dw
+    type(flist_t), allocatable :: curr(:)
+    type(flist_t), allocatable :: deriv(:)
     class(field_t), pointer :: u_out, v_out, w_out
 
     real(dp) :: t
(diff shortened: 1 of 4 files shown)


src/case/base_case.f90
```diff
@@ -230,30 +231,34 @@ subroutine run(self)
...
       do sub_iter = 1, self%solver%time_integrator%nstage
         ! first apply case-specific BCs
         call self%boundary_conditions()
 
-        du => self%solver%backend%allocator%get_block(DIR_X)
-        dv => self%solver%backend%allocator%get_block(DIR_X)
-        dw => self%solver%backend%allocator%get_block(DIR_X)
+        curr(1)%ptr => self%solver%u
+        curr(2)%ptr => self%solver%v
+        curr(3)%ptr => self%solver%w
```
semi-h, line 244: I think we can assign `curr`s outside of the do loop.

src/time_integrator.f90
```diff
@@ -167,33 +159,12 @@ function init(backend, allocator, method, nvars)
...
-
-  end subroutine step
-
-  subroutine runge_kutta(self, dt)
+  subroutine runge_kutta(self, curr, deriv, dt)
     implicit none
 
     class(time_intg_t), intent(inout) :: self
+    type(flist_t), allocatable, intent(inout) :: curr(:)
+    type(flist_t), allocatable, intent(in) :: deriv(:)
```
semi-h, line 167: If an object is allocated/deallocated outside we typically don't include `allocatable` attribute. Having `allocatable` attribute midly implies we may alloc/dealloc inside the subroutine, which is not the case here. Probably not going to interfere with the behaviour but I think its better to omit it here.

src/time_integrator.f90
```diff
@@ -247,10 +218,12 @@ subroutine runge_kutta(self, dt)
 
   end subroutine runge_kutta
 
-  subroutine adams_bashforth(self, dt)
+  subroutine adams_bashforth(self, curr, deriv, dt)
     implicit none
 
     class(time_intg_t), intent(inout) :: self
+    type(flist_t), allocatable, intent(inout) :: curr(:)
+    type(flist_t), allocatable, intent(in) :: deriv(:)
```
semi-h, line 226: See comment above

semi-h (approved):
Perfect! Because the chances included `get_block`/`release_block`, please run for example the TGV case for few hundred/thousand iterations just to make sure there is no memory leak issues (happens quite easily if we forget to release a field, and not yet have a mechanism in the code to detect this)

PR #189, opened 2025-06-23
Fix reordering for CUDA backen for SZ=64

### LINKED ISSUE #51: Reorder kernels in CUDA backend fail with SZ=64
This is due to us going above the maxiumum thread count per block when we set SZ=64. Some of the reorder kernels require a 2D thread with SZ*SZ threads. We need to fix these kernels so that when SZ=64 or above these kernels work on tiles of 32 by 32.

CFD-Xing:
Closes #51

### DIFF
--- a/tests/test_vecadd.f90
+++ b/tests/test_vecadd.f90
@@ -54,7 +54,7 @@ program test_vecadd
 
   call MPI_Init(ierr)
 
-  dims_global = [64, 64, 64]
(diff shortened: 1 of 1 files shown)


CFD-Xing:
@semi-h The do loop actually introduce significant overhead for SZ=32. So I instead unroll the loop for SZ=64.

CFD-Xing:
Yes, your proposed trick should work. Will do it shortly.

PR #192, opened 2025-07-02
Add basic ibm module

mathrack:
The present draft PR includes : 

- Basic IBM module : velocity in the solid is set to zero before computing the divergence
- The input for the IBM module is a ADIOS2 file named `ibm.bp` located in the folder of the simulation
- The basic IBM module can be activated when setting `ibm_on = T` in the solver namelist

For the basic IBM capability, the file `ibm.bp` should contain one integer (`iibm`) and the 3D mask (`ep1`). I have extended the ADIOS2 IO to allow reading a long integer from the file.

The input file can be generated with https://github.com/xcompact3d/py4x3d2.

The data workflow is not completely functional currently. I think there is a minor issue with data ordering. See https://github.com/ornladios/ADIOS2/issues/1665 for instance. I will convert the draft PR to a regular PR when I have a solution for the data ordering issue.

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -12,6 +12,7 @@ jobs:
     runs-on: ubuntu-22.04
     env:
       OMPI_VERSION: 4.1.5
+      ADIOS2_VERSION: 2.10.2
     steps:
 
     - uses: actions/checkout@v4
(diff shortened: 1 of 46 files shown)


src/module/ibm.f90
```diff
@@ -0,0 +1,132 @@
...
+
+      ! Get and fill a block on the host
+      ep1 => ibm%host_allocator%get_block(DIR_C)
+      do i = 1, dims(1)
+        do j = 1, dims(2)
+          do k = 1, dims(3)
+            ep1%data(i, j, k) = field_data(k, j, i)
+          end do
+        end do
+      end do
```
pbartholomew08, line 90: Could we add some explanation about the `ijk -> kji` swap?

pbartholomew08 (changes requested):
Looks good, I think only the ordering of indices needs clarification

src/module/ibm.f90
pbartholomew08: I think we should just add a comment to explain it, if we later figure out a better solution we can remove the comment, but I think this is fine otherwise

PR #198, opened 2025-07-04
Fix misleading BC rules in tdsops for the midpoint operators

### LINKED ISSUE #151: Staggered-interpolations and -derivations
This issue explains the current approach we have in x3d2, and it directly comes from Incompact3d. There are some non-obvious limitations because of our current approach, and in this issue I'm basically summarising our chats with @slaizet on this matter.

We use staggered interpolations and derivations in the divergence and gradient operations, and these are **always** applied to pressure-like fields in the current setup. Because we always apply them to pressure-like fields, the implementations have some inherent assumptions. Most importantly, even in the presence of Dirichlet BCs, we assume that pressure and pressure-like fields are subject to Neumann BCs, because with the FFT based Poisson solver we can only handle Neumann BCs and cannot deal with Dirichlet BCs.

Because we assume all pressure-like fields are always subject to Neumann BCs, we set up the staggered-interpolators and -derivatives such that they assume Neumann BCs in case of non-periodic global BCs. We may want to fix this, it'll require us to edit the part we instantiate staggered operators and superimpose Neumann BCs in case the global domain BCs are non-periodic.
https://github.com/xcompact3d/x3d2/blob/8b4e94368d117a247b8c019e76f8f04b39df77af/src/solver.f90#L181-L188

The final decision will probably depend on how we want to proceed with the iterative Poisson solver. Technically we won't have to enforce Neumann because iterative solver is not limited to Neumann, but its important to keep the current limitation and inherent assumptions with staggered operations in mind. @pbartholomew08

semi-h:
closes #151 

We now enforce BC_NEUMANN for the midpoint operators if the solver BCs are Dirichlet. If someone tries to use midpoint operators with Dirichlet BCs program terminates with error. We can simply come up with the dirichlet rules for the midpoint operators, but this is not used in xcompact3d. Can be extended in the future.


src/solver.f90
```diff
@@ -203,13 +204,27 @@ subroutine allocate_tdsops(dirps, backend, mesh, der1st_scheme, &
...
     real(dp) :: d
 
     dir = dirps%dir
+    d = mesh%geo%d(dir)
+
     bc_start = mesh%grid%BCs(dir, 1)
     bc_end = mesh%grid%BCs(dir, 2)
-    d = mesh%geo%d(dir)
+
+    ! BCs for the midpoint operators, enforce neumann only if the vertex BC
```
slaizet, line 216: I would provide a little bit more info here, to explain that because of the FFT, the BC for the pressure has to be NEUMANN. In practice it means that we impose a BC that is not compatible with Navier-Stokes, but that it does not affect the quality of the solution

slaizet (commented):
good to go, except adding a bit more info about why we do what we do :-)

src/solver.f90
semi-h: This is a good point, so maybe its not necessary for the iterative Poisson solver. Then this part might need to be bypassed if iterative solver is used instead of FFT.

Added a more detailed description as a comment.

PR #203, opened 2025-07-08
Refactor mesh and allocator

### LINKED ISSUE #166: Refactoring `mesh` and `allocator`
In the current design, a mesh object instantiation does not finalise its own initialisation, and requires an allocator to be instantiated such that the padding info is passed to the mesh object and only then its finalised and ready to use.

https://github.com/xcompact3d/x3d2/blob/2f1b375b1c8760d2f318ae02e04ee227606fa422/src/mesh.f90#L178-L180

https://github.com/xcompact3d/x3d2/blob/2f1b375b1c8760d2f318ae02e04ee227606fa422/src/allocator.f90#L91-L92

Additionally, when we use the allocator to get a `field`, its shape is set by the allocator by calling a mesh object function, so allocator relies on a mesh object. This creates a two-way circular dependency and its not ideal.

https://github.com/xcompact3d/x3d2/blob/2f1b375b1c8760d2f318ae02e04ee227606fa422/src/allocator.f90#L143-L144

We need to come up with a better abstraction so that these two classes can be independent, finalise their initialisation by themselves, and do not realy on each other when they're used.

I think the best way to achieve this is by moving the padded dimension info from mesh to allocator. Padding is a low level info that appears at backend level, so not mixing it with high level info like domain dimensions in the mesh is a good idea in my opinion. This is closely related to discussions we had before coming up with a mesh class, mainly around dimension info being scattared around. I think my current suggestion is okay because all the actual dimension info is still managed by mesh class, and only padded data is by allocator. What do you think, @pbartholomew08, @Nanoseb?

About implementation details of my suggestion, a simple search over the codebase reveals that `get_padded_dims` is used only in a handful of places, mainly in cuda/backend.f90 and omp/backend.f90. In most cases we ask the padding in `DIR_C` configuration, which is just padding in Cartesian layout. All those are super simple to change, we will simply ask the padding from the allocator instead of from the mesh. In OpenMP backend padding of a particular field and its directional layout is required, but it is also very simple to fix, we can just ask the shape of the `data` member of the field. Allocator is responsible for setting the field shape correctly so there will not be any issues.

```
../src/backend/cuda/backend.f90
457:    dims_padded = self%mesh%get_padded_dims(DIR_C)
574:    dims_padded = self%mesh%get_padded_dims(DIR_C)
596:    dims_padded = self%mesh%get_padded_dims(DIR_C)
653:    dims_padded = self%mesh%get_padded_dims(DIR_C)
722:    dims_padded = self%mesh%get_padded_dims(DIR_C)
863:    dims_padded = self%mesh%get_padded_dims(DIR_C)

../src/backend/omp/backend.f90
350:    dims = self%mesh%get_padded_dims(u)
406:    dims = self%mesh%get_padded_dims(u)
438:    dims = self%mesh%get_padded_dims(x)
580:    dims_padded = self%mesh%get_padded_dims(DIR_C)

../src/allocator.f90
69:    allocator%ngrid = product(allocator%mesh%get_padded_dims(DIR_C))
144:    dims = self%mesh%get_padded_dims(dir)

../src/mesh.f90
41:    procedure :: get_padded_dims_phi
42:    procedure :: get_padded_dims_dir
43:    generic :: get_padded_dims => get_padded_dims_dir, get_padded_dims_phi
300:  pure function get_padded_dims_dir(self, dir) result(dims_padded)
316:  pure function get_padded_dims_phi(self, phi) result(dims_padded)
323:    dims_padded = self%get_padded_dims(phi%dir)

../src/ordering.f90
89:    dims_padded = mesh%get_padded_dims(DIR_C)

```

More important aspect is the `n_group` we use in backends. We use it for two reasons, one is to allocate large enough buffers for communication, which can instead simply rely on getting this info from the allocator. The other reason is to figure out how many groups we have in the domain to run the tridiagonal solver over correct dimensions. I think we can remove all such uses and simply rely on the 3rd dimension of the `data` array. It is always the correct value, and simplifies everything. One alternative is that we can have an `n_group` member in `tdsops` class. Then we completely separate the data source from the 'rule' when applying a `tds_solve`. This will make our `tds_solve` operation similar in spirit to FFT libraries, where first a rule is defined, then the actual operation is called by passing plain arrays and a predefined rule, where user is responsible that the data layout in arrays match the predefined rule. In this strategy the arrays contain only the data, and the predefined rule contains all info about the size of the FFT, batch dimension, staggared or not and the amount, and dimensionality. Very similar to what we do currently except the `n_group` info.

semi-h:
closes #166

Ended up moving `get_padded_dims` and `get_n_groups` from mesh to allocator. Mesh is used for all the physical dimensions and allocator is used when padded dimensions are required. No more cyclic dependency between mesh and allocator (explained in detail in #166), and they can be instantiated independently.

### DIFF
--- a/src/allocator.f90
+++ b/src/allocator.f90
@@ -1,8 +1,7 @@
 module m_allocator
   use iso_fortran_env, only: stderr => error_unit
 
-  use m_common, only: dp, DIR_X, DIR_Y, DIR_Z, DIR_C, NULL_LOC, VERT
-  use m_mesh, only: mesh_t
+  use m_common, only: dp, DIR_X, DIR_Y, DIR_Z, DIR_C, NULL_LOC
   use m_field, only: field_t
 
   implicit none
@@ -35,14 +34,16 @@ module m_allocator
      !! [[m_allocator(module):release_block(subroutine)]].  The
      !! released block is then pushed in front of the block list.
 
-    integer :: ngrid
+    integer :: ngrid, sz
     !> The id for the next allocated block.  This counter is
     !> incremented each time a new block is allocated.
     integer :: next_id = 0
+    !> padded dimensions and n_groups in all 'dir's
+    integer, private :: dims_padded_dir(3, 4)
+    integer, private :: n_groups_dir(3)
     !> The pointer to the first block on the list.  Non associated if
     !> the list is empty
     ! TODO: Rename first to head
-    type(mesh_t), pointer :: mesh
     class(field_t), pointer :: first => null()
   contains
     procedure :: get_block
@@ -50,7 +51,8 @@ module m_allocator
     procedure :: create_block
     procedure :: get_block_ids
     procedure :: destroy
-    procedure :: compute_padded_dims
+    procedure :: get_padded_dims
+    procedure :: get_n_groups
   end type allocator_t
 
   interface allocator_t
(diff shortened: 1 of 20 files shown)


src/allocator.f90
```diff
@@ -59,39 +61,34 @@ module m_allocator
 
 contains
 
-  function allocator_init(mesh, sz) result(allocator)
-    type(mesh_t), target, intent(inout) :: mesh
-    integer, intent(in) :: sz
+  function allocator_init(nx, ny, nz, sz) result(allocator)
+    integer, intent(in) :: nx, ny, nz, sz
```
pbartholomew08, line 65: Would it be better to pass the grid dimensions as a triple? e.g.
```
allocator_init([nx, ny, nz], sz)
```
this would prevent mixing up where `sz` goes.

src/backend/cuda/backend.f90
```diff
@@ -479,7 +480,7 @@ subroutine tds_solve_dist(self, du, u, tdsops, blocks, threads)
     integer :: n_groups, dir
 
     dir = u%dir
-    n_groups = self%mesh%get_n_groups(u)
+    n_groups = self%allocator%get_n_groups(u%dir)
```
pbartholomew08, line 483: Is it neater just to pass `dir` here?

src/xcompact.f90
```diff
@@ -80,19 +80,21 @@ program xcompact
...
-  omp_allocator = allocator_t(mesh, SZ)
+  omp_allocator = allocator_t(dims(1), dims(2), dims(3), SZ)
   host_allocator => omp_allocator
 
   cuda_backend = cuda_backend_t(mesh, allocator)
   backend => cuda_backend
   if (nrank == 0) print *, 'CUDA backend instantiated'
 #else
-  omp_allocator = allocator_t(mesh, SZ)
+  omp_allocator = allocator_t(dims(1), dims(2), dims(3), SZ)
```
pbartholomew08, line 97: These could be called as
```
*allocator_t(dims, SZ)
```
(see previous comment)

pbartholomew08 (changes requested):
The "get `dims` array, then pass `dims(1),dims(2),dims(3)` seems to be the common pattern of use, I'd suggest that we change the interface to take the `dims` array as input to make this simpler in addition to distinguishing the arguments better. The few examples where this is not the case can be easily written as
```
allocator_t([8,8,8], 8)
```
etc.

semi-h:
All done!

PR #204, opened 2025-07-08
Add support in mesh%get_coordinates to return coords of any data_loc

### LINKED ISSUE #186: Consider `data_loc` of a field when obtaining coordinates
We need to add this support to `get_coordinates` in mesh so that we can pass `field%data_loc` and get the right global coordinates. 

_Originally posted by @semi-h in https://github.com/xcompact3d/x3d2/pull/173#discussion_r2150385557_

semi-h:
Closes #186 

`mesh%get_coordinates` returns VERT coordinates by default, but an arbitrary data_loc can be passed and then it returns the correct coordinates. I guess this can be useful in IO if we want to output a CELL field for example.

### DIFF
--- a/src/mesh.f90
+++ b/src/mesh.f90
@@ -305,15 +305,57 @@ contains
     end select
   end function get_n_dir
 
-  pure function get_coordinates(self, i, j, k) result(xloc)
-    !! Get the coordinates of a vertex with i, j, k local indices
+  pure function get_coordinates(self, i, j, k, data_loc_op) result(coords)
+    !! Get the coordinates of a vertex with i, j, k local cartesian indices
+    !! Avoid calling this in hot loops
     class(mesh_t), intent(in) :: self
     integer, intent(in) :: i, j, k
-    real(dp), dimension(3) :: xloc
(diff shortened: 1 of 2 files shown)


src/mesh.f90
```diff
@@ -305,15 +305,57 @@ pure function get_n_dir(self, dir, data_loc) result(n)
...
+    case (Y_EDGE)
+      coords(1) = self%geo%vert_coords(i, 1)
+      coords(2) = self%geo%midp_coords(j, 2)
+      coords(3) = self%geo%vert_coords(k, 3)
+    case (Z_EDGE)
+      coords(1) = self%geo%vert_coords(i, 1)
+      coords(2) = self%geo%vert_coords(j, 2)
+      coords(3) = self%geo%midp_coords(k, 3)
+    case default
+      error stop "Unknown location in get_dims_dataloc"
```
ia267, line 357: Could we change this to `"Unknown location in get_coordinates"`

ia267 (approved):
Approved

src/mesh.f90
semi-h: done

PR #206, opened 2025-07-09
Preliminary support for single precision

### LINKED ISSUE #70: Single precision support for MPI messages
There is the compile time seleciton mechanism for the precision in the code base, however the MPI modules are currently hardcoded for `MPI_DOUBLE_PRECISION`. We need to figure out a way to fix this so that we can run single precision simulations.

A very incomplete attemt is here (albeit the wrong place, it should be in base backend)
https://github.com/xcompact3d/x3d2/blob/85246f3b27fecee0095559029dbc4faadb7b5731/src/cuda/backend.f90#L26
Then based on the `dp` parameter `MPI_FP_PREC` can be set and used everywhere afterwards. This probably will require converting `mpi_sendrecv` modules into classes, so that they can be instantiated at the backends with the `MPI_FP_PREC`

Thanks @JamieJQuinn for pointing out this in #67.

semi-h:
Closes #70.

I tried running a TGV case with single precision. The first operation in a TGV case is calculating the enstrophy using the compact schemes, and based on the initial condition we get the expected value even at multiple ranks. However the simulation NaN's quickly. The fact that enstrophy calculation is correct showing us that there is nothing fundamentally wrong about the tridiagonal solvers, but still there is a but at some level. Some of the tests with tridiagonal solvers are also failing both on CUDA and OpenMP, while some of them are passing. Unfortunately I don't have enough time left to fix this bug, but this PR is at least adding a preliminary support and would be a good starting point for someone debugging the issue, so I think it can be merged into main, with a new issue detailing the bug with single precision. What do you think @slaizet @pbartholomew08 @ia267?

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -80,6 +80,10 @@ else()
   list(APPEND SRC ${DUMMYADIOS2SRC})
 endif()
 
+if(SINGLE_PREC)
+  add_definitions("-DSINGLE_PREC")
+endif()
+
 add_library(x3d2 STATIC ${SRC})
 target_include_directories(x3d2 INTERFACE ${CMAKE_CURRENT_BINARY_DIR})
 
--- a/src/backend/cuda/backend.f90
+++ b/src/backend/cuda/backend.f90
@@ -5,7 +5,7 @@ module m_cuda_backend
 
   use m_allocator, only: allocator_t
   use m_base_backend, only: base_backend_t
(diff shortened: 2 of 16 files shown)


pbartholomew08 (commented):
Changes all look OK

pbartholomew08:
It may not be worth implementing the fallback then. The `vecadd` passing and `time_integrator` failing suggests round-off error accumulation (one step vs multiple) to me

PR #208, opened 2025-07-11
Add documentation for DIR_X/Y/Z loops.

### LINKED ISSUE #71: Documentation regarding the loops over groups in the data structure.
@JamieJQuinn pointed out in #67 that its not clear for the user how to specify loops over groups.

I think we need a section in https://xcompact3d.github.io/x3d2/internals/ describing how we loop over groups in the domain both in CUDA and OpenMP backends. This is closely related to #68 and the subsequent changes that'll come in CUDA backend regarding reorderings as well.

semi-h:
Closes #71

### DIFF
--- a/docs/source/index.rst
+++ b/docs/source/index.rst
@@ -22,5 +22,6 @@ Contents:
    getting_started.rst
    user/index
    developer/index
+   internals/index
    report_a_bug
    API reference <https://xcompact3d.github.io/x3d2>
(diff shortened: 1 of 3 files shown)


ia267 (approved):
To make this visible you need to add `data_structure.rst` into `docs/source/internals/index.rst` 

This currently has `momentum_rhs_calculation.rst` and `allocator.rst` - if the momentum rhs is no longer applicable could you comment that line out (but still leave the .rst in case we decide to modify it).

PR #211, opened 2025-07-14
ADIOS2 XML snapshots for ParaView for uniform grids

### LINKED ISSUE #175: Enable ParaView integration for snapshots
Current visualisation snapshots lack metadata for optimal ParaView integration, making visualisation less intuitive for users.

**Implementation details:**

- [ ] Add XDMF metadata file generation for better ParaView loading
- [ ] Add proper variable naming and hierarchy
- [ ] Support for time series visualisation
- [ ] Explore writing 1D coordinate data with `ADIOSVTXReader`

**Acceptance criteria:**
- Snapshot files can be loaded directly in ParaView
- Time series data is properly interpreted
- Physical dimensions are preserved

ia267:
Writing snaphots have been modified to output using [VTK XML data model](https://adios2.readthedocs.io/en/latest/ecosystem/visualization.html). The resulting .bp files can be opened in ParaView.

It currently uses the VTK image data XML descriptor. This output will only work for uniform grids and it uses the grid origin and spacing. The origin is the coordinates of vertex (1,1,1) (in `mesh_content.f90` origin is defined but it's never set).

Some other changes were made to ADIOS2:

1. removed all variables when writing a new dataset (previously the variables were removed when closing, but I think removing them during writing is the proper way to do it)
2. Added a subroutine to define attributes using 1d arrays
3. Fixed a bug in `checkpoint_io.f90` where `release_block` was used outside the loop which would have caused memory leaks

The image below shows the paraview visualisation, three snapshots (for differen simulation times)

paraview_tgv

Closes #175

### DIFF
--- a/src/io/adios2_io.f90
+++ b/src/io/adios2_io.f90
@@ -64,7 +64,8 @@ module m_adios2_io
       write_array_3d_real, &
       write_array_1d_int, &
       write_array_4d_real
-    generic, public :: write_attribute => write_attribute_string
+    generic, public :: write_attribute => write_attribute_string, &
+                                          write_attribute_array_1d_real
 
     procedure, private :: write_scalar_int
     procedure, private :: write_scalar_real
@@ -74,6 +75,7 @@ module m_adios2_io
     procedure, private :: write_array_3d_real
     procedure, private :: write_array_4d_real
     procedure, private :: write_attribute_string
+    procedure, private :: write_attribute_array_1d_real
   end type adios2_writer_t
 
   !> ADIOS2 reader type
@@ -157,6 +159,13 @@ contains
     integer :: ierr, use_comm
     type(adios2_file_t) :: file   !! ADIOS2 file object
 
+    ! if opening in write mode, we are starting a new independent dataset
+    ! remove all old variables from the IO object
+    if (mode == adios2_mode_write) then
+      call adios2_remove_all_variables(self%io, ierr)
+      call self%handle_error(ierr, "Failed to remove old ADIOS2 variables before open")
+    endif
+
     use_comm = self%comm
     if (present(comm)) use_comm = comm
     if (.not. self%io%valid) &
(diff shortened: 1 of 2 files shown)


src/io/adios2_io.f90
```diff
@@ -394,6 +401,7 @@ subroutine write_array_3d_real( &
                                              count_dims
     type(adios2_variable) :: var
     integer :: ierr
+
     call adios2_define_variable(var, self%io, name, adios2_type_dp, &
                                 3, shape_dims, start_dims, &
                                 count_dims, adios2_constant_dims, ierr)
```
pbartholomew08, line 407: Should this not be guarded by a write mode / check for variable definition exists? I don't think you want to be redefining variables throughout the simulation.

src/io/checkpoint_io.f90
```diff
@@ -314,33 +318,41 @@ subroutine write_snapshot(self, solver, timestep, comm)
...
+    coords_max = solver%mesh%geo%L
+
+    do i = 1, 3
+      if (global_dims(i) > 1) then
+        original_spacing(i) = (coords_max(i) - origin(i))/ &
+                              real(global_dims(i) - 1, dp)
+      else
+        original_spacing(i) = 1.0_dp
+      end if
+    end do
```
pbartholomew08, line 332: I think grid spacing is (was?) stored in the `mesh` object

pbartholomew08 (changes requested):
Are we adding VTK to the checkpoints, I can't see it for snapshots, or does checkpoint manager also handle snapshots?

ia267:
> Are we adding VTK to the checkpoints, I can't see it for snapshots, or does checkpoint manager also handle snapshots?

The checkpoint manager contains routines for both snapshots and checkpoints, the changes were only made to the snapshot writing

src/io/checkpoint_io.f90
ia267: Yes, I'll need to change this. It is stored in `mesh%geo%d`

src/io/adios2_io.f90
ia267: I'll modify this to first call `adios2_inquire_variable` and only define it if it doesn't exist

pbartholomew08 (approved):
LGTM!

PR #215, opened 2025-07-21
Fix declaring the checkpoint_t object as public twice

pbartholomew08:
The new Cray compiler rejected the code when declaring the type public, and the symbol


ia267 (approved):
Approved - before you merge could you fix the fprettify error, this was introduced few changesets ago please?

PR #219, opened 2025-08-27
Change ep1 to be polymorphic to fix memcpy error

### LINKED ISSUE #218: type declaration for ep1 causes GPU copy failure
The pointer `ep1` is declared as `type(field_t)` in `ibm.f90`, i.e. monomorphic pointer that only point to objects of the exact `field_t` type. When it points to an extended type, the memory allocation and copy operations are based on the smaller, declared size of `field_t`, not the actual, dynamic size of the extended object. This causes the `memcpy` to write beyond the allocated device memory buffer.

### Proposed Solution

The pointer declaration should be changed to be polymorphic to handle extended types correctly:

**Change this:**
`type(field_t), pointer :: ep1 => null()`

**To this:**
`class(field_t), pointer :: ep1 => null()`

This ensures the runtime uses the object's dynamic type and size, preventing the memory access violation.

ia267:
Closes #218 

This PR fixes an illegal memory access error for IBM cases that occurs when a pointer to an extended data type is copied to device.

This is because pointer `ep1` is declared as `type(field_t)`. which prevents it from correctly handling objects of types that extend `field_t`. The `memcpy` operation consequently uses the wrong object size, leading to a buffer overrun.

This change modifies the declaration from `type(field_t)` to `class(field_t)`, making the pointer polymorphic. This allows it to correctly handle any object in the `field_t` inheritance hierarchy, ensuring that memory operations use the correct dynamic type and size.

### Summary of Changes

* Changed the declaration of the `ep1` pointer in `[filename.f90]` from `type(field_t)` to `class(field_t)`.

### Testing

* before fix, running in debug build gives: `0: copyin Memcpy (dev=0x72d55c000000, host=0x72d59bfffa70, size=134217728) FAILED: 700(an illegal memory access was encountered)`
* after fix code runs in debug build for IBM cases

### DIFF
--- a/src/module/ibm.f90
+++ b/src/module/ibm.f90
@@ -29,7 +29,7 @@ module m_ibm
     class(mesh_t), pointer :: mesh => null()
     type(allocator_t), pointer :: host_allocator => null()
     integer :: iibm = 0
-    type(field_t), pointer :: ep1 => null()
+    class(field_t), pointer :: ep1 => null()
   contains
     procedure :: body
   end type ibm_t


pbartholomew08 (approved):
LGTM

PR #223, opened 2025-09-23
Fix time-integrator and vecadd unit tests for single precision

CFD-Xing:


ia267 (approved):
Looks good. Resolves the truncation error being smaller than the round-off error for single-precision tests, and moves single-precision flag to root directory to cover test cases.

PR #224, opened 2025-09-24
Fix test coverage for test_thom and test_cuda_tridiag

CFD-Xing:
Most of the changes to cuda/test_thom.f90 and omp/test_thom.f90 are to improve consistencies between the two implementation to improve maintainability.

PS: Those unit tests are still failing when compiled with single precision:
	 10 - test_omp_tridiag_omp_4 (Failed)
	 11 - test_thom_omp_1 (Failed)
	 25 - test_thom_cuda_1 (Failed)
	 27 - test_cuda_tridiag_cuda_4 (Failed)

### DIFF
--- a/tests/cuda/test_thom.f90
+++ b/tests/cuda/test_thom.f90
@@ -2,7 +2,7 @@ program test_thom
 
   use iso_fortran_env, only: stderr => error_unit
   use cudafor
-  use m_common, only: dp, pi, BC_PERIODIC, BC_NEUMANN, BC_DIRICHLET, BC_HALO
+  use m_common, only: dp, nbytes, pi, BC_PERIODIC, BC_NEUMANN, BC_DIRICHLET, BC_HALO
   use m_cuda_common, only: SZ
   use m_cuda_exec_thom, only: exec_thom_tds_compact
   use m_cuda_tdsops, only: cuda_tdsops_t, cuda_tdsops_init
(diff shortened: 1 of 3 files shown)


tests/cuda/test_thom.f90
```diff
@@ -91,64 +89,85 @@ program test_thom
...
-    print'(a, f8.3, a)', 'Achieved BW: ', achievedBW/2**30, ' GiB/s'
+    integer :: ierr
+    integer :: nbytes, memClockRt, memBusWidth
+    real(dp) :: achievedBW, deviceBW
 
+    if (dp == kind(0.0d0)) then
+      nbytes = 8
+    else
+      nbytes = 4
+    end if
```
pbartholomew08, line 134: Is it worth making this a constant, e.g. where `dp` is defined? I can imagine there might be other uses of `nbytes` elsewhere

tests/omp/test_thom.f90
```diff
@@ -93,35 +90,41 @@ program test_thom
...
+    integer :: nbytes, memClockRt, memBusWidth
+    real(dp) :: achievedBW, deviceBW
 
     if (dp == kind(0.0d0)) then
-      nbytes = 8_dp
+      nbytes = 8
     else
-      nbytes = 4_dp
+      nbytes = 4
     end if
```
pbartholomew08, line 116: For example here we could use the predefined `nbytes`

pbartholomew08 (approved):
Only a minor change suggested - looks good otherwise

PR #227, opened 2025-09-29
Implement abstract IO

### LINKED ISSUE #216: Building without ADIOS2 is not possible
Currently attempting to build without enabling ADIOS2 fails. Do we want to
a) support building without ADIOS2 (requires redesigning/moving contents of IO modules)
b) make ADIOS2 a requirement
?

ia267:
IO implementation is refactored to use abstract I/O patterns with a session-based architecture. With this change we can build x3d2 without ADIOS2. When a user tries to run an IBM case (which requires ADIOS2) it will exit with an error if ADIOS2 is not available. In addition, if a user has `checkpoint_params` defined in the input file but did not compile ADIOS2 it will issue a warning (but continue running).

#### Abstract I/O architecture
- Session-based abstraction: new `reader_session_t` and `writer_session_t` provide a high-level interface that abstracts away backend complexity
- Polymorphic design - a new abstract `io_base.f90` defines all the interfaces that all I/O backends must implement. The session layer uses this interface to interact with the backends in a generic way.
- Compile-time backend selection: the specific I/O backend (currently only ADIOS2) is selected at compile time through the build system.

#### Modular separation of concerns
- Replaced `checkpoint_io.f90` with specialised managers:
  - `checkpoint_manager.f90`  - Handles simulation restart and recovery
  - `snapshot_manager.f90` - Handles visualisation output with VTK XML generation
  - `io_manager.f90` - Orchestrates both managers

#### Abstract session interface
```fortran
! Abstract usage pattern - same code works with any backend
use m_io_session, only: writer_session_t

type(writer_session_t) :: writer
call writer%open("output.bp", MPI_COMM_WORLD)
call writer%write_data("temperature", temp_field)
call writer%close()
```

#### New files
- `io_base.f90` - abstract base class for different backends
- `io_session.f90` - session-based I/O interface for reader and writer types
- `io_field_utils.f90` - shared field utilities between checkpoints and snapshots
- `io_manager.f90` - high-level I/O orchestrator for checkpoints and snapshots
- `checkpoint_manager.f90` - checkpoint/restart management
- `snapshot_manager.f90` - snapshot/visualisation management
- `adios2/io.f90` - ADIOS2 backend implementation
- `dummy/io.f90` - dummy backend
- `test_io_session.f90` - test case for io session functionality

#### Removed files
- checkpoint_io.f90 - old checkpoint/snapshot implementation
- checkpoint_dummy.f90 - old dummy implementation
- adios2_io.f90 - Old ADIOS2 implementation

Fixes #216

### DIFF
--- a/src/io/adios2/io.f90
+++ b/src/io/adios2/io.f90
@@ -155,21 +155,20 @@ contains
     integer, intent(in) :: comm
 
     class(io_file_t), allocatable :: file_handle
+    type(io_adios2_file_t) :: temp_handle
     integer :: ierr, use_comm
 
-    allocate (io_adios2_file_t :: file_handle)
     use_comm = comm
     if (.not. self%io_handle%valid) &
       call self%handle_error(1, "ADIOS2 IO object is not valid")
 
-    select type (file_handle)
-    type is (io_adios2_file_t)
-      call adios2_open( &
-        file_handle%engine, self%io_handle, filename, &
-        adios2_mode_read, use_comm, ierr)
-      call self%handle_error(ierr, "Failed to open ADIOS2 engine for reading")
-      file_handle%is_writer = .false.
-    end select
+    call adios2_open( &
+      temp_handle%engine, self%io_handle, filename, &
+      adios2_mode_read, use_comm, ierr)
+    call self%handle_error(ierr, "Failed to open ADIOS2 engine for reading")
+    temp_handle%is_writer = .false.
+
+    file_handle = temp_handle
   end function reader_open_adios2
 
   subroutine read_data_i8_adios2(self, variable_name, value, file_handle)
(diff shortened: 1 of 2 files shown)


src/io/io_session.f90
```diff
@@ -0,0 +1,215 @@
...
+  !!   call io_session%write_data("pressure", p_field, start_dims, count_dims)
+  !!   call io_session%write_attribute("ParaView", "vtk_xml_content")
+  !!   call io_session%close()
+  !!
+  !! The session automatically handles backend selection (currently just ADIOS2 but can be extended to MPI-IO, etc.)
+  !! and manages file handles, readers, and writers internally.
+  type :: io_session_t
+    private
+    class(io_reader_t), pointer :: reader => null()
+    class(io_writer_t), pointer :: writer => null()
```
pbartholomew08, line 44: Do we want a single object to support reading and writing or is the intention that you should have a reader session and a writer session? If the latter would it be better to specialise the type?

src/io/io_session.f90
```diff
@@ -0,0 +1,215 @@
...
+      self%file => self%reader%open(filename, io_mode_read, comm)
+      self%is_write_mode = .false.
+    else if (use_mode == io_mode_write) then
+      call allocate_io_writer(self%writer)
+      call self%writer%init(comm, "session_writer")
+      self%file => self%writer%open(filename, io_mode_write, comm)
+      self%is_write_mode = .true.
+    else
+      error stop "Invalid I/O mode"
+    end if
```
pbartholomew08, line 200: wrt above comment the user could try to use read operations from a session in write mode and vice-versa but this would not be caught until runtime.

src/io/adios2/io.f90
```diff
@@ -0,0 +1,659 @@
...
+  function reader_open_adios2(self, filename, mode, comm) result(file_handle)
+    class(io_adios2_reader_t), intent(inout) :: self
+    character(len=*), intent(in) :: filename
+    integer, intent(in) :: mode
+    integer, intent(in) :: comm
+
+    class(io_file_t), pointer :: file_handle
+    integer :: ierr, use_comm
+
+    allocate (io_adios2_file_t :: file_handle)
```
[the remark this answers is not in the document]
pbartholomew08: In fact, since it's a local variable, could it not just be `type(io_adios2_file_t)` and avoid allocation/deallocation, `select type`, etc?

PR #228, opened 2025-09-30
Initial OpenMP/GPU support

pbartholomew08:
This branch implements the basic offloading of data using openMP with the ability to perform the `vecadd` operation.


tests/omptgt/test_omptgt_allocator.f90
```diff
@@ -0,0 +1,95 @@
+program test_allocator_omptgt
+  use iso_fortran_env, only: stderr => error_unit
+
+  use m_allocator, only: allocator_t, field_t
+  use m_common, only: dp, pi, DIR_X
+  use m_mesh, only: mesh_t
+
+  use m_omptg_allocator, only: omptgt_allocator_t
```
ia267: typo: should be `m_omptgt_allocator`

I guess this file hasn't been added to `CMakeListst.txt` so it wasn't pickedup.

src/CMakeLists.txt
```diff
@@ -125,11 +132,16 @@ elseif(${CMAKE_Fortran_COMPILER_ID} STREQUAL "GNU" OR
   set(CMAKE_Fortran_FLAGS "-cpp -std=f2018")
   set(CMAKE_Fortran_FLAGS_DEBUG "-g -Og -Wall -Wpedantic -Werror -Wimplicit-interface -Wimplicit-procedure -Wno-unused-dummy-argument")
   set(CMAKE_Fortran_FLAGS_RELEASE "-O3 -ffast-math")
+  if (OMP_TGT)
+    # A bit of a hack - hardcoded for MI300A
```
ia267: Could we use CMake cache varibales (e.g. `OMP_TGT_ARCH`) so we don't need to edit CMake files?

tests/CMakeLists.txt
```diff
@@ -39,10 +39,19 @@ function(define_test testfile np backend)
     endif()
     find_package(OpenMP REQUIRED)
     target_link_libraries(${test_name} PRIVATE OpenMP::OpenMP_Fortran)
+
+    if(${backend} STREQUAL omp_tgt)
+      # Note this is somewhat of a hack - hardcoded to build against MI300A
```
ia267, line 48: Could we use CMake cache variables instead?

src/backend/omp/target/allocator.f90
```diff
@@ -0,0 +1,179 @@
...
+  subroutine omptgt_field_destroy(self)
+    class(omptgt_field_t) :: self
+
+    nullify (self%data_tgt)
+    nullify (self%p_data_tgt)
+    call omp_target_free(self%dev_ptr, self%dev_id)
+  end subroutine
+
+  ! Deallocates device-resident memory before deallocating the base type
+  subroutine destroy(self)
```
ia267: This `destroy` subroutine is not bound to `omptgt_allocator_t`. Need to add `procedure :: destroy => destroy` to `omptgt_allocator_t`

ia267 (changes requested):
Couple of things to address before merging.

tests/CMakeLists.txt
CFD-Xing: We can have a list of options as follow:
```
set(gpu_archs gfx942 CACHE STRING "")
set_property(CACHE gpu_archs PROPERTY STRINGS gfx942 gfx900 gfx902 gfx906 gfx908 gfx909 gfx90a gfx90c)
```

One other alternative is to add an additional option of the form cmake .. -Dgpu_arch=gfx942

This would avoid updating the list of toggled option every time there is a new architecture coming out but put the responsibility of setting the correct input on the user.

In both case, we should be able to set the parameter as follow
`      target_compile_options(${test_name} PRIVATE "--offload-arch=${gpu_arch}")`

tests/performance/perf_cuda_reorder.f90
```diff
@@ -99,17 +99,17 @@ subroutine prepare_input(case_id)
     case (CASE_Y2X, CASE_Y2Z)
       blocks = dim3(nx/SZ, nz, ny/SZ)
       threads = dim3(min(SZ, 32), min(SZ, 32), 1)
-      call reorder_x2y<<<blocks, threads>>>(u_temp_d, u_i_d, nz)
+      call reorder_x2y <  <  < blocks, threads >  >  > (u_temp_d, u_i_d, nz)
```
CFD-Xing: I think that this reformatting is going to break to CUDA compilation...

src/backend/omp/target/backend.f90
```diff
@@ -0,0 +1,310 @@
...
+    do k = 1, dims(3)
+      do j = 1, dims(2)
+        do i = 1, dims(1)
+          call get_index_reordering(out_i, out_j, out_k, i, j, k, &
+                                    dir_from, dir_to, SZ, cart_padded)
+          u_(out_i, out_j, out_k) = u(i, j, k)
+        end do
+      end do
+    end do
+    !$omp end target teams distribute parallel do
```
CFD-Xing: Observation: Re-ordering without using shared (scratchpad) memory is likely to be inefficient on GPU due to non-coalesced memory access

tests/performance/perf_cuda_reorder.f90
pbartholomew08: Oops!

src/CMakeLists.txt
pbartholomew08: Good point - this was added for development, and not intended as the actual implementation

tests/CMakeLists.txt
pbartholomew08: I have gone with the user must specify option, now when configuring the user should do
```
cmake -B build . <options> -DOMP_TGT=ON -DOMP_TGT_ARCH=<arch>
```

src/CMakeLists.txt
pbartholomew08: see comment below for resolution

tests/omptgt/test_omptgt_allocator.f90
pbartholomew08: Corrected - and added to the tests/CMakeLists.txt under `unit/`

CMakeLists.txt
```diff
@@ -25,6 +25,10 @@ else()
   message(STATUS "ADIOS2 is disabled")
 endif()
 
+set(OMP_TGT OFF CACHE BOOL
+  "Enable OpenMP target offloading.")
+set(OMP_TGT_ARCH "" CACHE STRING "Target architecture for OpenMP offloading, e.g. gfx942")
+
```
CFD-Xing, line 43: I believe that I need OpenMP 5.1, so I would add a CMake check on the OpenMP version

src/backend/omp/target/allocator.f90
```diff
@@ -0,0 +1,179 @@
...
+  end subroutine fill_omptgt
+
+  subroutine fill_omptgt_(p_data_tgt, c, n)
+    real(dp), dimension(:), intent(inout) :: p_data_tgt
+    real(dp), intent(in) :: c
+    integer, intent(in) :: n
+
+    integer :: i
+
+    !$omp target teams distribute parallel do has_device_addr(p_data_tgt)
```
CFD-Xing: It appears that `has_device_addr` is not supported by nvfortran, which only support OpenMP 5.0 (not 5.1): https://docs.nvidia.com/hpc-sdk//compilers/hpc-compilers-user-guide/index.html#using-openmp

So I would add a compilation check in the root CMakeLists.txt as currently noting prevent a user to use an OpenMP backend on a Nvidia GPU. 

On the flip side, this also mean that we can NOT currently use the department GitLab CI to test the OpenMP backend as it currently use a A40.

CFD-Xing: One alternative is to limit to OpenMP 5.0 and use is_device_ptr if that works.

tests/CMakeLists.txt
CFD-Xing: Observation: with the current logic, those cuda unit tests will be run if one use OpenMP on a Nvidia machine.

src/backend/omp/target/allocator.f90
pbartholomew08: Finally resolved this - using a finalise-bound destructor breaks the "Fortran constructor" model, I've moved to initialising the omptgt field type with a subroutine (as this is only called by the allocator when creating a new block it does not result in inconsistent user code).

src/backend/omp/target/allocator.f90
```diff
@@ -0,0 +1,166 @@
...
+    f%next => next
+    f%id = id
+
+    f%dev_id = omp_get_default_device()
+    if (f%dev_id == omp_get_initial_device()) then
+      error stop "Device ID is HOST"
+    end if
+
+    f%dev_ptr = omp_target_alloc(ngrid*c_sizeof(0.0_dp), f%dev_id)
+    call c_f_pointer(f%dev_ptr, f%p_data_tgt, shape=[ngrid])
```
CFD-Xing, line 91: I think this issue can be addressed in a separate PR, but the pre-mapping of a device pointer to an Fortran array does not seem to be fully portable.

src/backend/omp/target/allocator.f90
```diff
@@ -0,0 +1,166 @@
...
+    real(dp), intent(in) :: c
+    integer, intent(in) :: n
+
+    integer :: i
+
+    !$omp target teams loop has_device_addr(p_data_tgt)
+    do i = 1, n
+      p_data_tgt(i) = c
+    end do
+    !$omp end target teams loop
```
CFD-Xing, line 126: This can be addressed in a separate PR, but a more portable way would be to use `is_device_ptr` on the actual device pointer and to map the pointer to a Fortran within the OMP directive.

tests/CMakeLists.txt
ia267: if a user sets `-DOMP_TGT=ON` but forgets to set the `-DOMP_TGT_ARCH` it may lead to empty flag. Better to fail it early with something like:

```fortran
if(OMP_TGT AND OMP_TGT_ARCH STREQUAL "" AND
   NOT CMAKE_Fortran_COMPILER_ID STREQUAL "Cray")
  message(FATAL_ERROR
    "OMP_TGT=ON requires -DOMP_TGT_ARCH=<arch> for this compiler")
endif()
```

tests/CMakeLists.txt
```diff
@@ -120,6 +135,16 @@ if(WITH_2DECOMPFFT)
   define_verification_test(verification/test_fft.f90 4 omp)
 endif()
 
+if(${OMP_TGT})
```
ia267, line 138: An NVHPC OpenMP target can still pick up CUDA tests unintentionally. I think for the CUDA tests below (line 152) we can add `AND NOT OMP_TGT` to prevent this for now. A longer-term fix will be explicit test options per backend instead of using compiler identities - we can do this separately.

tests/CMakeLists.txt
```diff
@@ -120,6 +135,16 @@ if(WITH_2DECOMPFFT)
...
+  # OMP TGT unit tests
+  define_test(unit/test_vecadd.f90 1 omp_tgt)
+  define_test(unit/test_setget_field.f90 1 omp_tgt)
+  define_test(unit/test_omptgt_allocator.f90 1 omp_tgt)
+
+  # OMP TGT verification tests
+  define_test(verification/test_time_integrator.f90 1 omp_tgt)
```
ia267: this needs to be:
```suggestion
  define_verification_test(verification/test_time_integrator.f90 1 omp_tgt)
```

ia267 (changes requested):
3 small changes to be made - rest of the stuff we can do it in a separate issue.

tests/CMakeLists.txt
pbartholomew08: Done

tests/CMakeLists.txt
pbartholomew08: Done

pbartholomew08:
I believe all issues are addressed

ia267 (approved):
Approved - please merge

Issue #43, opened 2024-02-20
Use more standard build directories/layout


pbartholomew08:
CMake supports building following a standard `/path/lib`, `/path/bin`, etc. layout, e.g. `x3d2/build/lib`, `x3d2/build/bin`

This will make it easier to find things (e.g. the executable), ensure that generated (sub)modules are placed in a standard place for better editor integration.

kaanolgu:
Implemented in PR #233

PR #233, opened 2025-11-11
Enhancements to CMake for Organisation

kaanolgu:
Addressing the issue #43


pbartholomew08:
I have implemented the proposed changes in #234

PR #234, opened 2025-11-13
Place executables in relevant bin/ directories

pbartholomew08:
This PR creates the binaries (main, tests) in the relevant `bin/` directories and ensures tests are launched using this `bin` path


kaanolgu (approved):
Thank you for this, tested and all scenarios work

Edit: Gitlab build recipes need to be updated too I guess for the checks
Edit2: Seems like Gitlab is failing due to some usage limit

PR #246, opened 2025-12-10
Updating the checkpoint parameters section of the input parameters page in the user guide

### LINKED ISSUE #244: Typo in the checkpoint generation code snippet provided in the user guide
In the user guide, under input parameters ([https://x3d2.readthedocs.io/en/latest/user/input_file.html](https://x3d2.readthedocs.io/en/latest/user/input_file.html)), the code snippet provided states the checkpoint section is set using the "&checkpoint" heading. This, however, is not read for checkpoint / snapshot generation when used as the heading required should be "&checkpoint_params" instead, as discussed with @ia267.

ishaanalidina:
# Closes #244 

- Updated the input parameters page of the user guide so that the namelist block title now specifies ```checkpoint_params``` rather than ```checkpoint```. This should correct the typo in this section.

### DIFF
--- a/src/module/ibm.f90
+++ b/src/module/ibm.f90
@@ -87,6 +87,7 @@ contains
       ! Get and fill a block on the host
       ! The order of the data is corrected in the loop below
       ep1 => ibm%host_allocator%get_block(DIR_C)
+      call ep1%fill(1.0_dp)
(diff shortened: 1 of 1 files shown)


ia267 (approved):
Approved

PR #260, opened 2025-12-11
Example for ibm + cylinder with the current periodic poisson solver

### LINKED ISSUE #257: Example for IBM + cylinder with the current periodic poisson solver
Add one example for the IBM + cylinder flow using the current periodic poisson solver

mathrack:


ia267 (approved):
Approved - I was running this case and it diverged around iteration 3170. I think for now we could set the n_iters = 3000

ia267 (approved):
Make sense to limit it to 3000

PR #261, opened 2025-12-11
IBM: stop the simulation if the IBM input file si needed and missing

mathrack:


src/module/ibm.f90
```diff
@@ -60,6 +61,17 @@ function init(backend, mesh, host_allocator) result(ibm)
     ibm%mesh => mesh
     ibm%host_allocator => host_allocator
 
+    ! Stop if the file is missing
+    if (mesh%par%is_root()) inquire (file=ibm_file, exist=file_exists)
+    call MPI_Bcast(file_exists, 1, MPI_LOGICAL, 0, MPI_COMM_WORLD, ierr)
+    if (ierr /= 0 .or. (.not. file_exists)) then
```
ia267: If `ierr /=0` (`MPI_Bcast` error), this will fail even if the file exists - could be misleading. We can have two if-statements

ia267 (changes requested):
Tested it on a case without a `ibm.bp` file but with `ibm_on = .true.` and it ends the run giving an error message.

Requested change to the if statement.

ia267 (approved):
Approved

PR #266, opened 2026-01-16
Poisson_100 and poisson_110 with Poisson Boundary Condition Tests

kaanolgu:
## Description

This PR adds support for the **`poisson_100`** (non-periodic in X, periodic in Y/Z) and **`poisson_110`** (non-periodic in X and Y, periodic in Z) boundary condition cases, along with a 3D test using `cos(nπx)`, `cos(nπx)cos(nπy)`, and `cos(nπx)cos(nπy)cos(nπz)` where n = 2, 3 analytical solutions to validate both the Poisson solve and the `div(grad())` operator.

---

### What's Added

- **`poisson_100` via transpose reuse of `poisson_010`**: Rather than implementing a new spectral post-processing path, the `100` case transposes the input field from `(nx, ny, nz)` to `(ny, nx, nz)` on the GPU before the FFT, runs it through the existing `poisson_010` kernels (`process_spectral_010_fw/bw`, `enforce/undo_periodicity_y`), then transposes back. This avoids duplicating post-processing logic entirely.

- **cuFFTMp R2C setup for `poisson_100`**: The R2C plan is configured over the transposed domain so cuFFT sees the Dirichlet direction as the leading dimension. The output buffer is sized `(ny+2, nx, nz)` using standard Hermitian padding. After the transform, spectral post-processing and Poisson solve happen in-place, followed by C2R and the back-transpose.

- **New GPU kernels** in `spectral_processing.f90`:
  - `memcpy3D_with_transpose` / `memcpy3D_with_transpose_back` — lightweight device kernels to swap X/Y layout for the `100` case.
  - `process_spectral_110` — full Poisson pipeline for Dirichlet X+Y, periodic Z (normalisation, forward/backward Z post-processing, forward/backward Y post-processing, eigenvalue division).
  - `enforce_periodicity_x` / `undo_periodicity_x` — even-extension and its reversal along X, mirroring the existing Y-direction equivalents.

- **OpenMP backend**: Stub error messages added for `100` and `110` cases indicating these are not yet implemented on CPU.

- **Test suite**: New unified `test_poisson_bc.f90` replacing the old case structure, covering `000`, `010`, `100`, and `110` with analytical reference checks. `CMakeLists.txt` updated with new CTest targets and `input_000`/`input_110` namelists added.

---

### Notes

- `x_sp_st` is passed to `process_spectral_110` but currently unused — flagged for cleanup.
- A `VERBOSE` compile flag controls debug output in the test.


src/backend/cuda/kernels/spectral_processing.f90
```diff
@@ -604,6 +650,191 @@ attributes(global) subroutine process_spectral_010_bw( &
 
   end subroutine process_spectral_010_bw
 
+  attributes(global) subroutine process_spectral_110( &
+    div_u, waves, nx_spec, ny_spec, x_sp_st, y_sp_st, nx, ny, nz, &
```
ia267: Here we get `x_sp_st` but it's never used?

src/backend/cuda/poisson_fft.f90
```diff
@@ -85,13 +105,52 @@ function init(mesh, xdirps, ydirps, zdirps, lowmem) result(poisson_fft)
...
+      n_sp_st(2) = dims_loc(2)/mesh%par%nproc_dir(3)*mesh%par%nrank_dir(3)
+      n_sp_st(3) = 0
+    else if ((.not. periodic_x) .and. (.not. periodic_y) .and. periodic_z) then
+      ! 110 case: Dirichlet X, Dirichlet Y, Periodic Z
+      ! Standard spectral layout (no transpose needed)
+      n_spec(1) = dims_loc(1)/2 + 1  ! nx/2+1 (R2C)
+      n_spec(2) = dims_loc(2)/mesh%par%nproc_dir(3)    ! ny
+      n_spec(3) = dims_glob(3)        ! nz
+
+      n_sp_st(1) = dims_loc(1)/mesh%par%nproc_dir(3)*mesh%par%nrank_dir(3)
```
ia267: For 110 case, I don't quite understand why `n_sp_st(1)` is different to other cases (where it was set to 0). Note also that since `self%x_sp_st = n_sp_st(1)`, but we saw that in `process_spectral_110` this is actually not used?

ia267 (changes requested):
- Most comments are for me understand what's happening
- There are some places with incorrect comments that needs fixing
- The test structure is more suited to `tests` directory instead of `case` - this will need refactoring
- This PR is implementing core functionality and there are lots of changes here, it will be useful for others to know details of what has been implemented, please provide detailed description in this PR

ia267:
- made changes in [a1876f4](https://github.com/xcompact3d/x3d2/pull/266/commits/a1876f43cdc82ffcc7f99690cea903e20f0b09b7) to fix cufft path which was causing test_poisson_bc failures

- separate `r_dev` and `c_dev` device arrays - now each buffer has exactly the right size and type
- using `2*(nx_loc/2+1)` instead of `nx_loc + 2` - important for odd `nx_loc`. For even `nx`, `nx+2 = 2*(nx/2+1)` (i.e.  both expressions are identical). For odd e.g. `nx = 65`, `nx_loc+2` will give 67 but the correct value is `2*(32+1) = 66`. If we use 67 it over-indexes the `xtdesc` allocation and could lead to read/write one element past the end of the buffer
- odd dimension handling included in periodicity kernels - for `enforce_periodicity_x`  the old approach with `nx=65` will try to read `f_in(i=66..68, ..)` , i.e. out of bounds. For `undo_periodicity_x`, position `nx=65` was never written. The new kernel with `mod(nx, 2) == 1` handles both cases.
- For cuFFT path zeroed both `padded_dev` and `f_out_dev `- this seemed to be necessary - there would be a performance hit but it's limited to cufft - not ideal and we will need full deterministic kernels

ia267 (approved):
Implementation of non-periodic bcs look OK

cuFFT path gave issues for test_poisson_bc.f90 test case. These have been resolved, but there are some changes that for the cuFFT path specifically that might result in reduced  performance (for e.g. `f_out_dev` zero will cost a full-field GPU zero per Poisson solve call).

PR #267, opened 2026-01-29
Update developer documentation

### LINKED ISSUE #247: Improve contributor documentation
- The modules, subroutines and functions should describe what they do
- We should describe what the variables are 

This is currently missing for most of the code

ia267:
- Added comments to all the source files using FORD format (see https://forddocs.readthedocs.io/en/stable/) - this was done using GitHub Copilot with Claude Sonnet 4.5
- The process was semi-automatic (the AI agent had missed some parts of the documentation, or some of it was incorrect).  Most comments are not verified for correctness. When we are modifying some code we have to make sure that it is actually correct. For the code that I did look at, they were mostly correct (but for some parts of the code it had hallucinated and made up stuff)

For review I suggest looking at couple of source files that you're familiar with and check if the comments are OK (in terms of accuracy, verbosity).

Alternatively you could do the following:
- install ford using `pip install ford`
- go to the directory `x3d2/docs` and then run `ford ford.md`
- Then open the the `x3d2/docs/doc/index.html` in a browser

Close #247

### DIFF
--- a/src/backend/cuda/poisson_fft.f90
+++ b/src/backend/cuda/poisson_fft.f90
@@ -1,4 +1,9 @@
 module m_cuda_poisson_fft
+  !! FFT-based Poisson solver on GPU using cuFFT.
+  !!
+  !! Extends poisson_fft_t with device-resident spectral data and cuFFT plans.
+  !! Handles forward/backward transforms, spectral post-processing for different
+  !! boundary conditions, and periodic extensions.
   use iso_c_binding, only: c_loc, c_ptr, c_f_pointer, c_int, c_float, &
                            c_double_complex, c_float_complex
   use iso_fortran_env, only: stderr => error_unit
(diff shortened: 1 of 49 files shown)


src/backend/cuda/poisson_fft.f90
```diff
@@ -259,15 +284,19 @@ subroutine fft_backward_cuda(self, f)
   end subroutine fft_backward_cuda
 
   subroutine fft_postprocess_000_cuda(self)
+    !! Post-process spectral data for Dirichlet-Dirichlet-Dirichlet boundaries.
+    !!
+    !! Solves Poisson equation $\nabla^2 p = f$ in spectral space with homogeneous
+    !! Dirichlet boundaries in all directions.
```
kaanolgu:
```suggestion
    !! Periodic boundaries in all directions.
```

kaanolgu (changes requested):
Initial check for poisson_fft and poisson_fft_cuda

kaanolgu (changes requested):
The conflicts are coming from #266  I think it requires another iterate over the changed files since it has conflicts now

PR #277, opened 2026-02-17
Implement GPU-aware IO

### LINKED ISSUE #176: Implement CUDA-aware IO in ADIOS2
Current implementation treats CPU and GPU data identically, which may lead to unnecessary data transfers using the CUDA backend

**Implementation details:**
- Modify `adios2_io.f90` to detect backend type at runtime
- Add backend-specific implementation for write operations
- Implement CUDA-specific code path using ADIOS2's [GPU-aware functionality](https://adios2.readthedocs.io/en/latest/advanced/gpu_aware.html) 
- Add proper synchronisation for CUDA streams before I/O
- Add build-time detection of ADIOS2 CUDA support

Acceptance criteria:
- Data remains on GPU during I/O operations when using CUDA backend
- Performance improvement for checkpoint/snapshot operations with large datasets on GPU
- Graceful fallback when ADIOS2 lacks CUDA support
- No performance regression for OMP backend

### LINKED ISSUE #276: Allow mixed-precision input reading
Currently IO requires input variables to match the solver's precision. This prevents users from reading input files (like IBM masks or restarts) generated in double-precision when running a single-precision simulation (and vice-versa).

We should allow automatic type conversion in the ADIOS2 reader. It should detect the variable type in the file and convert it to solver's precision if a mismatch is found.

ia267:
This PR implements GPU aware IO (closes #176) where data is written directly from device memory to ADIOS2 without host staging.

Process:
- After defining the variable, tell ADIOS2 the buffer lives on GPU
- Pass the device pointer via C wrapper (the Fortran API call `adios2_put` does not accept CUDA `device` array.  Therefore a C wrapper `adios2_put_gpu` is created to receive the device pointer and calls the C `adios2_put`. Without the C wrapper it triggers a device-to-copy which defeats the purpose)
- At runtime, write-field subroutines detects the cuda backend and uses the GPU path to write. Non-cuda paths uses the default host path

In order to use GPU-aware IO we have to use the compiler flag: `-DWITH_ADIOS2_GPU_AWARE=ON`

As part of this PR I have also implemented single-precision writing from GPU, as well as reading mixed-precision (closes #276) and as well fixed a bug in checkpoints, where it was using the host pointer (%data)  see 20d98bf.


kaanolgu:
@x3d2bot fix style

kaanolgu:
TGV 256^3 with `ibm_on`, 20 iterations, snapshot every iteration, double precision, a single A100, one rank. GPU memory sampled with `nvidia-smi --query-gpu=memory.used -lms 200`

| run | peak GPU mem (MiB) | wall time (s) | 
|---|---|---|
| solver only, no I/O ( this PR build) | 6279  | 6.3 | 
| regular ADIOS2 (main) | 6279  | 42.6 | 
| this PR build, CPU I/O | 6279  | 39.8 | 
| this PR build, GPU I/O | 6407  | 31.1 | 

So with the GPU offloaded I/O we save `~27%` compared to CPU based I/O as we reduce data movement

kaanolgu:
image
Proof of GPU usage rather than fallback to CPU

ia267:
device_io
host_io

##### differences between GPU-aware IO and default

Setup: TGV 256^3 with IBM, double precision, 1 rank on an RTX A4000, `output_stride = 1,1,1`, snapshot every step (3 fields), profiled over 6 steps. T

Per snapshot (3 x 256^3 doubles):

| | CPU I/O | GPU-aware I/O |
|---|---|---|
| Preparing the fields | `IO_HostStage` 3 x ~180 ms (copy to host + reorder on the CPU)<br>`IO_HostPack` 3 x ~135 ms (copy into the write buffer) | `ADIOS2_DevicePack` 3 x ~3.8 ms (GPU kernel) |
| `ADIOS2_Put` | 3 x ~100 ms (ADIOS2 copies the host buffer) | 3 x ~88 ms (ADIOS2 copies from the GPU) |
| `ADIOS2_EndStep` (file write) | ~240 ms | ~240 ms |
| One snapshot step | ~1.5 s | ~0.58 s |

The CPU path handles each field four times: copy from the GPU, reorder on the CPU, copy into the write buffer, then ADIOS2 copies it again. The GPU path packs the field in a single kernel, and ADIOS2 copies it off the GPU once.

The same amount of data crosses PCIe in both builds (3.6 GB over the profiled run), and the file write takes the same time. The whole saving comes from replacing the CPU reorder and repack with the GPU pack kernel.

#### Notes
- Strided snapshots (`output_stride > 1`) still use the CPU path, so they don't benefit yet. Checkpoints always use the GPU path.
- ADIOS2 copies from the GPU into pageable host memory at ~1.5 GB/s. Pinned memory could make this copy faster in both builds.
- Numbers are from a single profiled run each.

src/CMakeLists.txt
```diff
@@ -166,6 +172,14 @@ if (${ENABLE_BACKEND} STREQUAL "CUDA")
     target_link_options(x3d2 INTERFACE "-cudalib=cufft")
   endif()
   target_compile_options(xcompact PRIVATE "-DCUDA")
+  # NVTX ranges label the I/O paths in Nsight Systems (see io/nvtx.f90).
+  target_compile_definitions(x3d2 PRIVATE X3D2_NVTX)
```
kaanolgu: CUDA backend builds always add `X3D2_NVTX` isn't this going to be extra  work for non ADIOS2 builds ?

kaanolgu (changes requested):

kaanolgu:
Tested results on A100 (https://github.com/xcompact3d/x3d2/pull/277/commits/21bdd5480a6e2858475dc5373f475608a1d065be) 
- Peak GPU memory (scenario https://github.com/xcompact3d/x3d2/pull/277#issuecomment-5665142288) :
  - 6279 MiB (no I/O) 
  -  6279 MiB (main, regular ADIOS2)
  -  6279 MiB (this PR, CPU I/O) 
  -  6535 MiB (this PR, GPU I/O)
- Wall time on TGV (scenario https://github.com/xcompact3d/x3d2/pull/277#issuecomment-5665142288): 
  - 6.4s (no I/O)
  - 41.0s (main) 
  - 40.8s (CPU I/O) 
  - 30.8s (GPU I/O) 
   
 - GPU-aware I/O saves ~24% wall time versus CPU I/O.

src/CMakeLists.txt
ia267: The runtime cost will be negligible without ADIOS2, but I'll set to use it only if ADIOS2 is on since it's not used anywhere else for now.

kaanolgu (approved):

PR #283, opened 2026-03-03
Add cylinder case with IBM inflow/outflow BC [CUDA]

kaanolgu:
## Summary

Adds (`m_case_cylinder`) for simulating flow around a circular cylinder using the solver's built-in Immersed Boundary Method (IBM).

## Boundary conditions

- **Inflow (x=0):** Dirichlet — `u=1, v=0, w=0`
- **Outflow (x=L):** BC condition on `u`, Dirichlet zero on `v` and `w`
- **y, z:** Periodic

The outflow BC discretises `du/dt + Uc * du/dx = 0` as:

```
u_N^{n+1} = u_N^n - (Uc * dt/dx) * (u_N^n - u_{N-1}^n)
```

The velocity `Uc` is computed dynamically each timestep as the global maximum of `u` at `nx-1`, following the approach used in [Xcompact3d's cylinder wake case](https://github.com/xcompact3d/Incompact3d/blob/master/src/Case-Cylinder-wake.f90). This prevents blowup that occurs when using a fixed `Uc`

The outflow is applied both in `boundary_conditions` (on `U^m`) and in `pre_correction` (on `U*` before the Poisson solve), which is required for the projection method to maintain consistency.

## Initial conditions

Uniform flow `u=1, v=0, w=0` with optional localized random noise (controlled via `init_noise` in the namelist)

## Configuration

- parameter `init_noise(3)` for per-component noise amplitude. Example input:

```fortran
&cylinder_nml
init_noise = 0.125 0. 0.
/End
```

Requires `ibm_on = T` and `BC_x = 'dirichlet', 'dirichlet'` in the solver settings (could work without IBM but it won't create data for visuals) 

## Tested

- Re = 300, 257×128×32 grid, AB3 time integration, `dt = 0.0075`, 30k iterations – stable with dynamic `Uc`.


kaanolgu:
@x3d2bot fix style

src/backend/cuda/kernels/spectral_processing.f90
```diff
@@ -75,6 +75,55 @@ attributes(global) subroutine memcpy3D_with_transpose_back( &
 
   end subroutine memcpy3D_with_transpose_back
 
+  attributes(global) subroutine transpose_xyz_to_zxy(dst, src, nx, ny, nz)
+    !! Transpose: src(nx, ny, nz) → dst(nz, nx, ny)
```
ia267, line 79: instead of the unicode character →, could we use standard ASCII -> for example?

src/backend/cuda/kernels/spectral_processing.f90
```diff
@@ -75,6 +75,55 @@ attributes(global) subroutine memcpy3D_with_transpose_back( &
...
+    if (j <= ny) then
+      do i = 1, nx
+        dst(k, i, j) = src(i, j, k)
+      end do
+    end if
+
+  end subroutine transpose_xyz_to_zxy
+
+  attributes(global) subroutine transpose_zxy_to_xyz(dst, src, nx, ny, nz)
+    !! Transpose back: src(nz, nx, ny) → dst(nx, ny, nz)
```
ia267: same here

src/backend/cuda/kernels/fieldops.f90
```diff
@@ -194,6 +194,43 @@ attributes(global) subroutine field_set_y_face(f, c_start, c_end, nx, ny, nz)
...
+    if (i <= i_max) then
+      ! --- Left face (j = 1) ---
+      select case (bc_start)
+        f(i, 1, b) = c_start
+      end select
+
+      ! --- Right face (j = nx) ---
+      select case (bc_end)
+      f(i, nx, b) = f(i, nx, b) - c_end*(f(i, nx, b) - f(i, nx - 1, b)) + fl_correction
+      end select
```
ia267, line 242: These `select case` blocks are incorrect - you must use `case` statements to define specific values, for example it could be:

```fortran
select case (bc_start)
case (BC_DIRICHLET)
  f(i, 1, b) = c_start
case (BC_NEUMANN)
  !this can be empty for now 
end select
```

src/backend/cuda/kernels/fieldops.f90
```diff
@@ -194,6 +194,43 @@ attributes(global) subroutine field_set_y_face(f, c_start, c_end, nx, ny, nz)
...
+
+    if (i <= i_max) then
+      ! --- Left face (j = 1) ---
+      select case (bc_start)
+        f(i, 1, b) = c_start
+      end select
+
+      ! --- Right face (j = nx) ---
+      select case (bc_end)
+      f(i, nx, b) = f(i, nx, b) - c_end*(f(i, nx, b) - f(i, nx - 1, b)) + fl_correction
```
ia267: I think it's confusing that `c_end` has a dual meaning. For example for the y-face, `c_end` is used to set the boundary face value (see line 192). But for the x-face it's as a multiplier on the convective outflow. This will make it confusing for someone using the interface later. I think it the idea was to have Dirichlet and Neumman BCs right? I think we need to add the case labels to make this clear

src/case/cylinder.f90
```diff
@@ -92,29 +92,134 @@ subroutine initial_conditions_cylinder(self)
...
+    call self%compute_outflow_params(out_vel, fl_correction)
+
+    call self%solver%backend%field_set_face( &
+      self%solver%u, 1._dp, out_vel, X_FACE)
+
+    call self%solver%backend%field_set_face( &
+      self%solver%v, 0._dp, out_vel, X_FACE)
+
+    call self%solver%backend%field_set_face( &
+      self%solver%w, 0._dp, out_vel, X_FACE)
```
ia267: Could we make the BC type explicity here? so `bc_start` and `bc_end` to `BC_DIRICHLET` (which it defaults to in `field_set_face_cuda`).

Also instead of hard-coding `u=1, v=0, w=0` could we not obtain these from the input file?

src/case/cylinder.f90
```diff
@@ -92,29 +92,134 @@ subroutine initial_conditions_cylinder(self)
...
+  ! ==========================================================================
+  subroutine boundary_conditions_cylinder(self)
+    implicit none
     class(case_cylinder_t) :: self
-    class(field_t), intent(inout) :: du, dv, dw
-    integer, intent(in) :: iter
+    real(dp) :: out_vel, fl_correction
 
-  end subroutine forcings_cylinder
+    call self%compute_outflow_params(out_vel, fl_correction)
```
ia267: `fl_correction` is computed but never applied.

src/case/cylinder.f90
```diff
@@ -92,29 +92,134 @@ subroutine initial_conditions_cylinder(self)
...
+  ! Same logic as boundary_conditions — Dirichlet inflow, convective outflow.
+  ! ==========================================================================
   subroutine pre_correction_cylinder(self, u, v, w)
     implicit none
-
     class(case_cylinder_t) :: self
     class(field_t), intent(inout) :: u, v, w
+    real(dp) :: out_vel, fl_correction
+
+    call self%compute_outflow_params(out_vel, fl_correction)
```
ia267: `compute_outflow_params` is called in both `boundary_conditions_cylinder` and `pre_correction_cylinder` - this launches 3 GPU kernels + 2 MPI_Allreduces each time. Could we cache the results if the fields haven't changed between the 2 calls or refactor the code so that the 2 subroutines share the computed values?

src/backend/cuda/kernels/fieldops.f90
```diff
@@ -194,6 +194,43 @@ attributes(global) subroutine field_set_y_face(f, c_start, c_end, nx, ny, nz)
...
+      end select
+    end if
+  end subroutine field_set_x_face
```
ia267, line 244: minor fromatting issue - could you add a blank line after this subroutine?

ia267 (changes requested):

src/backend/cuda/kernels/fieldops.f90
kaanolgu: The reason I kept the `c_end` name is that it would be good to have the same convention as `Y_FACE` and `X_FACE` the only change is that for `X_FACE` it is output of the computations with varying values, for the `Y_FACE` it is a float number set for all. Essentially it is the same logic so wanted to keep it same in the core subroutine level and provide different inputs in wrapper

src/case/cylinder.f90
kaanolgu: Thanks for this, removed while debugging!

src/backend/cuda/kernels/fieldops.f90
ia267: OK makes sense - could you add a comment at the call site in `cylinder.f90` to clarify. For example:
`! Outflow BC: out_vel is the convective CFD (uxmax * gdt / dx)`

and may in the kernel another comment saying that `c_end` is a boundary value that can come from different sources.

ia267 (approved):
Changes approved - just need to add a comment in the code regarding `c_end` use and fix fprettify issues

PR #284, opened 2026-03-04
ParaView Tutorial

kaanolgu:
This PR adds paraview tutorial steps I followed to have ParaView running on with CLI on the HPC systems to download the generated final video to my laptop.


docs/source/user/paraview.rst
```diff
@@ -0,0 +1,188 @@
...
+
+- Verify MPI uses GCC: ``mpicc -show`` should show ``gcc``, not ``nvc``.
+- If ``OPAL_PREFIX`` is set to an NVHPC path, unset it: ``unset OPAL_PREFIX``.
+
+Step 2: Build ADIOS2 with GCC + OpenMPI
+----------------------------------------
+
+.. code-block:: bash
+
+   ADIOS2_SRC=<path-to-x3d2>/build-gpu-adios2/adios2-build/adios2-src
```
ia267: Here you are asking to use the source downloaded by x3d2. I think it will be better to describe how to install adios2 without relying on x3d2 installation.

We can refer to the adios2 user manual here:
https://adios2.readthedocs.io/en/latest/setting_up/setting_up.html

The only difference will be just the cloning command:
`git clone https://github.com/ornladios/ADIOS2.git ADIOS2`

The rest will be the same

docs/source/user/paraview.rst
```diff
@@ -0,0 +1,188 @@
...
+     -DCMAKE_INSTALL_PREFIX=$ADIOS2_INSTALL
+
+   make -j$(nproc)
+   make install
+
+- Uses the ADIOS2 v2.10.2 source already downloaded by x3d2's ExternalProject.
+- Installs to ``$ADIOS2_INSTALL``.
+
+Step 3: Build ParaView
+----------------------
```
ia267: For paraview we could refer to this their official installation instructions: https://www.paraview.org/paraview-docs/latest/cxx/md__builds_gitlab-kitware-sciviz-ci_Documentation_dev_build.html

plus any specific flags we need (such as the adios2 installation directory)

kaanolgu:
> It might be better separate these into two separate `.rst` files. One can be this:
> 
> `Building ParaView with ADIOS2 for visualisation` - this will just explain how to install adios2 and paraview (up to step 4)
> 
> and the second could be:
> 
> `Visualising snapshots in ParaView` - where you can add the sections from step 5 onward
> 
> You will need to update `index.rst`. For example we can have a higher-level `.rst` called `post_processing.rst` and into which we can add these two sections (which are themselves separate `.rst` files)


I think it's best to keep this as a single document with all 6 steps, since the steps are meant to be followed end-to-end – I haven't covered any advanced visualisations, so a reader would need all 6 steps just to get a basic view of the data. Splitting it might add unnecessary navigation for what's currently a straightforward walkthrough. Happy to revisit the structure if we add more post-processing content later.

ia267:
> > It might be better separate these into two separate `.rst` files. One can be this:
> > `Building ParaView with ADIOS2 for visualisation` - this will just explain how to install adios2 and paraview (up to step 4)
> > and the second could be:
> > `Visualising snapshots in ParaView` - where you can add the sections from step 5 onward
> > You will need to update `index.rst`. For example we can have a higher-level `.rst` called `post_processing.rst` and into which we can add these two sections (which are themselves separate `.rst` files)
> 
> I think it's best to keep this as a single document with all 6 steps, since the steps are meant to be followed end-to-end – I haven't covered any advanced visualisations, so a reader would need all 6 steps just to get a basic view of the data. Splitting it might add unnecessary navigation for what's currently a straightforward walkthrough. Happy to revisit the structure if we add more post-processing content later.

I would split it to 2, because before the first release we will have to add additional post-processing such as the stats. I think it'll make sense to have a dedicated folder under user called post_processing and add `post_processing.rst` and `paraview.rst` and `visualising_snapshots.rst` for example

ia267 (changes requested):

ia267 (approved):
Approved - one change required to link x3d2 compilation with ADIOS2

PR #287, opened 2026-03-10
Add statistics and output post-processed fields

ia267:
Adds runtime stats accumulation and adds additional optional fields to snapshots (vorticity magnitude and Q-criterion), which closes #285 .  Note that this PR also fixes #286 

#### Stats module
- Created new module: `stats.f90` to create the running means
- Obtains mean velocities (`umean`, `vmean`, `wmean`), mean squares (`uumean`, `vvmean` and `wwmean`) and cross-moments (`uvmean`, `uwmean`, `vwmean`) 
- Pressure stats (`pmean`) - if pressure output is enabled
- Scalar species stats (`phimean`, `phiphimean`) - if `n_species >0`
- Stats parameters can be configured from the input file (`initstat`, `istatfreq`, `istatout`)
- Velocity fluctuations ($$u' = \sqrt{\overline{u^2} - \bar{u}^2}$$) and Reynolds-stresses ($$\overline{u'v'} = \overline{uv} - \bar{u}\,\bar{v}$$) are computed at write time (note that Xcompact3d writes the raw means `umean`, `uumean` etc. and rms velocity and Reynolds stresses are later computed using post-processing tools such as Py4Incompact3D.)
- Includes stats for restart runs (uses current checkpointing architecture) 
- In Xcompact3d stats output uses the same frequency as checkpoints, here the two are independent.
- We don't allow stats to be done on a coarse grid (possible aliasing issues)
- Added a new test case 

#### Post-process module
- moved some post-processing done in `solver.f90` to this new module
- added new post-processed fields
- vorticity magnitude: $$|\boldsymbol{\omega}| = \sqrt{\omega_x^2 + \omega_y^2 + \omega_z^2}$$
- Q-criterion: $$Q = -\frac{1}{2} \sum_{i,j} \frac{\partial u_i}{\partial x_j} \frac{\partial u_j}{\partial x_i}$$

#### Input changes
- Invdivdual boolean flag (`output_pressure` for example) replaced with a single extensible namelist array: `output_fields = 'pressure', 'vorticity', 'qcriterion'`

#### Tests
- Tested on simple TGV and channel cases to make sure that the stats are written to file and that restarts work. Full testing to be done on the cylinder case.

### DIFF
--- a/src/io/snapshot_manager.f90
+++ b/src/io/snapshot_manager.f90
@@ -125,6 +125,7 @@ contains
     if (self%config%snapshot_freq <= 0) return
     if (mod(timestep, self%config%snapshot_freq) /= 0) return
 
+    allocate (field_names(0))
     field_names = get_snapshot_fields(self%config)
 
     call MPI_Comm_rank(comm, myrank, ierr)
--- a/src/io/snapshot_manager.f90
+++ b/src/io/snapshot_manager.f90
@@ -11,7 +11,8 @@ module m_snapshot_manager
   use m_field, only: field_t
   use m_solver, only: solver_t
   use m_io_session, only: writer_session_t
-  use m_config, only: checkpoint_config_t, has_output_field
+  use m_config, only: checkpoint_config_t, has_output_field, &
+                      NUM_SNAPSHOT_FIELDS
   use m_io_field_utils, only: field_buffer_map_t, field_ptr_t, &
                               setup_field_arrays, cleanup_field_arrays, &
                               stride_data_to_buffer, get_output_dimensions, &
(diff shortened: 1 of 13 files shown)


src/io/snapshot_manager.f90
```diff
@@ -189,6 +186,28 @@ subroutine write_snapshot(self, solver, timestep, comm)
     deallocate (field_names)
   end subroutine write_snapshot
 
+  function get_snapshot_fields(config) result(names)
+    !! Build the list of field names for snapshot output
+    type(checkpoint_config_t), intent(in) :: config
+    character(len=32), allocatable :: names(:)
+
+    character(len=32) :: tmp(6)
```
kaanolgu: Currently there are 6 possible fields (u, v, w, p, vort, qcrit), which exactly fills the array. If a new output field is added in the future, it would be easy to miss this change. Could we add following to `config.f90` and call it in this function rather than hardcoding the 6 ? 

```fortran
integer, parameter :: NUM_SNAPSHOT_FIELDS = 6
```

kaanolgu (changes requested):
Great work! Noticed small things that would be good to clarify/edit.

kaanolgu (approved):

PR #288, opened 2026-03-11
Fix 110 case Poisson solver with R2C FFT and race-free spectral processing

kaanolgu:
~Replace R2C/C2R with C2C FFT for 110 boundary conditions (periodic Z,
non-periodic X/Y) to handle full spectral domain correctly.~
Transpose (nx, ny, nz) to (nz, nx, ny)  for reuse of R2C FFT 110 boundary conditions (periodic Z,
non-periodic X/Y) to handle full spectral domain correctly.

- ~Add C2C FFT plan and complex workspace for 110 case~ Add transpose for 110 case
- Implement dedicated fft_forward_110_cuda and fft_backward_110_cuda
- Split fft_postprocess_110_cuda into 5 sequential kernels to prevent
  race conditions in cross-block writes
- Add enforce/undo_periodicity_xy methods
- ~Update nx_spec to full domain size (not halved) for C2C~

Fixes pressure solution errors in 110 cylinder case caused by R2C symmetry
assumptions conflicting with non-periodic boundary conditions.


src/backend/cuda/kernels/spectral_processing.f90
```diff
@@ -752,63 +879,130 @@ attributes(global) subroutine process_spectral_110( &
...
+            - l_r*bx(ix_pair + x_sp_st) + l_c*ax(ix_pair + x_sp_st), &
+          kind=dp &
+          )
+      end do
+    end if
+
+  end subroutine process_spectral_110_x_pair_bw
+
+  attributes(global) subroutine process_spectral_110_z_bw( &
+    div_u, nz_h, nx, ny, nz, az, bz &
```
ia267: `nz` parameter is never used in this subroutine - consider removing it

ia267 (changes requested):
Looks good - added few comments for some fixes

ia267 (approved):
Looks good - ready to merge when the `cuFFT` checks are also complete

PR #300, opened 2026-03-31
Output global monitoring quantities to a csv file

ia267:
This PR adds global monitoring output to a csv file so that some key quantities can be tracked during the simulation.

#### Change

- add a new `m_monitoring` module that computes and writes:
  - time
  - enstrophy
  - maximum divergence
  - mean divergence
- write these quantities to `monitoring.csv` on output steps (it will still print to `stdout`)
- move the monitoring lifecycle into the base case so setup/finalisation is handled centrally
- replace the per-case enstrophy/divergence print helpers in the case implementations with the shared monitoring writer
- add user documentation for the new csv output, including column descriptions and a simple plotting example
- move `postprocess.f90` into `src/postprocess/` and update CMake accordingly

Closes #295

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -3,6 +3,7 @@ a.out
 *.mod
 build/
 build.*
+build*
 *.o
 
 # Emacs
--- a/.gitignore
+++ b/.gitignore
@@ -3,7 +3,6 @@ a.out
 *.mod
 build/
 build.*
-build*
 *.o
 
 # Emacs
--- a/.gitignore
+++ b/.gitignore
@@ -1,8 +1,7 @@
 *~
 a.out
 *.mod
-build/
-build.*
+build*/
 *.o
 
 # Emacs
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -7,7 +7,8 @@ set(SRC
   ordering.f90
   poisson_fft.f90
   solver.f90
-  postprocess.f90
+  postprocess/postprocess.f90
+  postprocess/monitoring.f90
   tdsops.f90
   time_integrator.f90
   ordering.f90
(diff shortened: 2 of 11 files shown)


src/case/base_case.f90
kaanolgu:
```suggestion
if(self%solver%n_output > 0) then
      if (mod(iter, self%solver%n_output) == 0) then
```
complains if we set n_output = 0 not divisible

kaanolgu, line 265:
```suggestion
          end if
      end if
```

.gitignore
```diff
@@ -3,6 +3,7 @@ a.out
 *.mod
 build/
 build.*
+build*
```
kaanolgu: Why this ?

src/CMakeLists.txt
```diff
@@ -113,7 +114,7 @@ target_link_libraries(xcompact PRIVATE x3d2_backends)
 if(${CMAKE_Fortran_COMPILER_ID} STREQUAL "PGI" OR
    ${CMAKE_Fortran_COMPILER_ID} STREQUAL "NVHPC")
   set(CMAKE_Fortran_FLAGS "-cpp -cuda")
-  set(CMAKE_Fortran_FLAGS_DEBUG "-g -O0 -traceback -Mbounds -Mchkptr -Ktrap=fp")
+  set(CMAKE_Fortran_FLAGS_DEBUG "-g -O0 -traceback -Mbounds -Mchkptr")
```
kaanolgu: why removed Ktrap=fp ?

kaanolgu (changes requested):
It looks promising, tested with latest version too all good.

src/CMakeLists.txt
ia267: After an update I did on my machine which has an NVIDIA RTX A4000 I noticed my cuFFTMp wasn't working in debug build (it was giving an error about a division by zero). I traced it to this flag, which for some reason causing the issue on this particular GPU (works on Kolmogorov though). But I think maybe it's better to leave the flag on for now and report it to NVIDIA and see what they say (locally I'll just disable this).

.gitignore
ia267: I often create custom build directories within my x3d2 directory and I wanted to prevent those build directories from being picked up. I think I'll remove this though as I could just call it `build.cuda` for example that will ignore it

src/case/base_case.f90
ia267: Made this change

ia267: Made this change

kaanolgu (approved):
LGTM!

PR #305, opened 2026-05-01
capture fprettify stderr in GitHub Actions check

### LINKED ISSUE #304: Fprettify check silently passes on formatting errors
fprettify writes its output to stderr, but the workflow captures only stdout:

```
FPRETTIFY_OUTPUT=$(fprettify --config .fprettify.ini --diff --recursive src)
```

So FPRETTIFY_OUTPUT is always empty and the failure condition is never triggered.

The GitLab CI job correctly uses > (which captures both stdout and stderr), which is why it catches the same issues that GitHub Actions misses.

Example PR #283 
```
$ FPRETTIFY_OUTPUT=$(fprettify --config .fprettify.ini --diff --recursive src)
$ echo $FPRETTIFY_OUTPUT 

$ FPRETTIFY_OUTPUT=$(fprettify --config .fprettify.ini --diff --recursive src 2>&1)
$ echo $FPRETTIFY_OUTPUT 
WARNING: File src/backend/backend.f90, line 253 auto indentation failed due to chars limit, line should be split (limit: 79) WARNING: File src/backend/backend.f90, line 253 auto indentation failed due to chars limit, line should be split (limit: 79) WARNING: File src/backend/backend.f90, line 253 auto indentation failed due to chars limit, line should be split (limit: 79) WARNING: Fi
...
```

kaanolgu:
The fprettify check in GitHub Actions was passing and GitLab is failing because
fprettify writes its failing line split WARNINGs output to stderr, which was not being captured
by the $() command substitution.

Fix by redirecting stderr to stdout with 2>&1:
```
- FPRETTIFY_OUTPUT=$(fprettify --config .fprettify.ini --diff --recursive src)
+ FPRETTIFY_OUTPUT=$(fprettify --config .fprettify.ini --diff --recursive src 2>&1)
```
It requires a lot of extra effort from user and it is not just a "safe" warning which is not ideal 

Closes #304

### DIFF
--- a/.github/workflows/check_policies.yml
+++ b/.github/workflows/check_policies.yml
@@ -15,7 +15,7 @@ jobs:
         id: fprettify-check
         run: |
           echo "Running fprettify..."
-          FPRETTIFY_OUTPUT=$(fprettify --config .fprettify.ini --diff --recursive src)
+          FPRETTIFY_OUTPUT=$(fprettify --config .fprettify.ini --diff --recursive src 2>&1)
           echo "$FPRETTIFY_OUTPUT"
           {
             echo "fprettify_output<<EOF"


kaanolgu:
Now it captured the warnings as error to the merged in changes from the #283 

https://github.com/xcompact3d/x3d2/actions/runs/25209575255/job/73917104232?pr=305

PR #307, opened 2026-05-14
Refactor case BC hooks: rename to define_BC/apply_BC and hoist BC fields to base

kaanolgu:
- Rename `boundary_conditions` -> `define_BC` and `pre_correction` -> `apply_BC` across base and child cases.
- Hoist 12 BC field pointers (`bc_{start,end}_{u,v,w}_{x,y}`) onto `base_case_t`; cases use only what they need.
- Add `field_set_face_from_field` backend hook (CUDA + OMP) for spatially-varying face BCs.
- Channel: y-wall BCs now use persistent device fields refreshed with noise per substep, applied via the new hook.
- Cylinder: inflow profile now built field-side using new `inlet_noise(3)` config (replaces `bc_start_u/v/w` scalars).
- Cylinder: `compute_outflow_params` moved from `apply_BC` into `define_BC`; results stored on `self%out_vel` / `self%flow_rate_diff` (drops `_cached` suffix).
- **Behavioural note:** outflow params sampled one substep earlier than before.
- **Breaking change:** existing `cylinder_nml` files must replace `bc_start_u/v/w` with `inlet_noise`.


src/case/cylinder.f90
```diff
@@ -145,86 +139,110 @@ subroutine compute_outflow_params(self, out_vel, flow_rate_diff)
...
+    ! so apply_BC (later in the substep) and postprocess can both read them.
+    ! Note: these are sampled pre-step here, whereas the old layout sampled
+    ! them post-step inside apply_BC; the resulting BC parameters drift by
+    ! one substep relative to the stamp.
+
+    ! TODO: remove cache data when the PR#300 merges in or add a commit
+    !      to fix it here too accordingly
+    call self%compute_outflow_params(self%out_vel, self%flow_rate_diff)
+
+    ! Allocate persistent device BC fields on first call.
```
ia267, line 182: `hu%data`, `hv%data`, `hw%data` are used before it's actually populated. We should do something similar to what's done in `initial_conditions_cylinder`:

```fortran
call random_number(hu%data(1, 1:dims(2), 1:dims(3)))
call random_number(hv%data(1, 1:dims(2), 1:dims(3)))
call random_number(hw%data(1, 1:dims(2), 1:dims(3)))
```

src/backend/omp/backend.f90
ia267, line 780: I think the block-index formula is wrong here. From `field_set_face_omp (see `k_end = k + (dims(2)-1)/SZ*dims(3))`  (line 852), the convention for DIR_X in OMP is `k = z + (y_block -1)*nz`, i.e. z varies fast in the third dimension. Here we have the opposite order. 

The correct formula should be `k = k_i + (k_j - 1)*n_i`

src/backend/cuda/backend.f90
ia267, line 897: I think the `atomicmax` is not fully implemented for double (works for integers), we need to check this.

src/backend/cuda/backend.f90
```diff
@@ -1057,6 +1061,79 @@ subroutine field_set_face_cuda(self, f, c_start, c_end, face, &
...
+    bc_e = BC_DIRICHLET
+    flow_rate_diff_val = 0._dp
+    if (present(bc_start)) bc_s = bc_start
+    if (present(bc_end)) bc_e = bc_end
+    if (present(flow_rate_diff)) flow_rate_diff_val = flow_rate_diff
+
+    call resolve_field_t(f_d, f)
+    call resolve_field_t(f_start_d, f_start)
+
+    ! NOTE: NULL_LOC == VERT == 0 in this codebase, so guarding on
```
ia267: This comment is not correct, `NULL_LOC = -0001` and `VERT = 0` (see common.f90). I think the code still works because VERT is used unconditionally, but we need to fix the comment

src/backend/cuda/kernels/fieldops.f90
```diff
@@ -195,7 +195,36 @@ attributes(global) subroutine field_set_y_face( &
     end if
 
   end subroutine field_set_y_face
+  attributes(global) subroutine field_set_y_face_from_field( &
+    f, f_start, flow_rate_diff, nx, ny, nz)
```
ia267: uses `flow_rate_diff` as an argument but it's never used in the kernel

src/backend/cuda/kernels/fieldops.f90
ia267, line 174: same here `flow_rate_diff` is not used in the subroutine

src/case/cylinder.f90
ia267: we can remove `CELL` as it's not used

src/config.f90
```diff
@@ -53,9 +55,7 @@ module m_config
 
   type, extends(base_config_t) :: cylinder_config_t
     real(dp) :: init_noise(3)
-    real(dp) :: bc_start_u = 1._dp
-    real(dp) :: bc_start_v = 0._dp
-    real(dp) :: bc_start_w = 0._dp
+    real(dp) :: inlet_noise(3)
```
ia267, line 58: We need to initialise `inlet_noise` in case an input file omits them

```fortran
real(dp) :: inlet_noise(3) = [0._dp, 0._dp, 0._dp]
````

File: src/backend/omp/backend.f90
Comment (ia267) [inline src/backend/omp/backend.f90:855]: should we do an error stop here like we do for `Y_FACE`?

**Review by ia267 — CHANGES_REQUESTED** [review]

File: src/case/cylinder.f90 (hunk 1 above)
Comment (kaanolgu) [reply src/case/cylinder.f90:182]: We populate the data at around line 197, It would be a duplicate work which eventually hinders the performance. The `initial_conditions_cylinder` is only run once but if we have `100k` iterations then we can't afford to fill in values in each run twice. That's why I avoided that 

Line 197- 203: 
```
    do k = 1, dims(3)
      do j = 1, dims(2)
        hu%data(1, j, k) = 1._dp + noise(1)*um*(2._dp*hu%data(1, j, k) - 1._dp)
        hv%data(1, j, k) = noise(2)*um*(2._dp*hv%data(1, j, k) - 1._dp)
        hw%data(1, j, k) = noise(3)*um*(2._dp*hw%data(1, j, k) - 1._dp)
      end do
    end do
```

File: src/config.f90 (hunk 4 above)
Comment (kaanolgu) [reply src/config.f90:58]: Then that means we also need to initialise `init_noise` too right ? Because in the original version that wasn't initialised too so I kept the same format

File: src/backend/cuda/kernels/fieldops.f90
Comment (kaanolgu) [reply src/backend/cuda/kernels/fieldops.f90:174]: Thanks for pointing out! I will `flow_rate_diff` into the  subroutine

File: src/backend/cuda/kernels/fieldops.f90 (hunk 3 above)
Comment (kaanolgu) [reply src/backend/cuda/kernels/fieldops.f90]: Thanks for pointing out! I will `flow_rate_diff` into the  subroutine

File: src/backend/cuda/backend.f90
Comment (kaanolgu) [reply src/backend/cuda/backend.f90:897]: Thanks for pointing out! I will check it and update if required

File: src/case/base_case.f90
Comment (kaanolgu) [inline src/case/base_case.f90]: Change this
```
    if (mod(iter, self%solver%n_output) == 0) then
        t = iter*self%solver%dt

        call self%postprocess(iter, t)
      end if
```
to this for safeguarding when we want to disable the postprocessing completely
```
      if (self%solver%n_output > 0) then
        if (mod(iter, self%solver%n_output) == 0) then
          t = iter*self%solver%dt

          call self%postprocess(iter, t)
        end if
      end if
```

File: src/backend/omp/backend.f90
Comment (kaanolgu) [reply src/backend/omp/backend.f90:780]: The storage is actually y-block-fast, so the existing line is right and this flip would break it. My evidence is `get_index_dir` in `ordering.f90`, which is the thing that actually decides where each value lands when a field is built (reorder_omp uses it). For `DIR_X` it computes the group as y-block fast.

```
DIR_X: n_j = dims(2) (y), n_i = dims(3) (z)
DIR_Y: n_j = dims(1) (x), n_i = dims(3) (z)
DIR_Z: n_j = dims(1) (x), n_i = dims(2) (y)
```

`k = k_i + (k_j - 1)*n_i` would make the leftover dim fastest, and that's wrong for all three directions, not just X. The existing line is the one that matches get_index_dir everywhere.

File: src/case/cylinder.f90 (hunk 1 above)
Comment (kaanolgu) [reply src/case/cylinder.f90:182]: made the required changes in the commit [fc13b64](https://github.com/xcompact3d/x3d2/pull/307/commits/fc13b64e521d68a639a73a732b7d2d7ebdf4bfac)

File: src/backend/cuda/backend.f90
Comment (kaanolgu) [reply src/backend/cuda/backend.f90:897]: I left this as it is, might create an issue ? because I don't have concrete answer

File: src/backend/cuda/backend.f90 (hunk 2 above)
Comment (kaanolgu) [reply src/backend/cuda/backend.f90]: made the required changes in the commit [fc13b64](https://github.com/xcompact3d/x3d2/pull/307/commits/fc13b64e521d68a639a73a732b7d2d7ebdf4bfac)

File: src/backend/cuda/kernels/fieldops.f90 (hunk 3 above)
Comment (kaanolgu) [reply src/backend/cuda/kernels/fieldops.f90]: made the required changes in the commit [fc13b64](https://github.com/xcompact3d/x3d2/pull/307/commits/fc13b64e521d68a639a73a732b7d2d7ebdf4bfac)

Removed it and we could add it if required in the future for `field_set_y_face_from_field`

File: src/config.f90 (hunk 4 above)
Comment (kaanolgu) [reply src/config.f90:58]: made the required changes in the commit [fc13b64](https://github.com/xcompact3d/x3d2/pull/307/commits/fc13b64e521d68a639a73a732b7d2d7ebdf4bfac)

File: src/case/base_case.f90
Comment (kaanolgu) [reply src/case/base_case.f90]: This is left for future PR

File: src/backend/omp/backend.f90
Comment (kaanolgu) [reply src/backend/omp/backend.f90:855]: We don't use this for channel and cylinder cases, I added `field_set_face_from_field` which enables the features that we need

File: src/backend/omp/backend.f90 [hunk 5]
```diff
@@ -863,6 +864,55 @@ subroutine field_set_face_omp(self, f, c_start, c_end, face, &
...
+    flow_rate_diff_val = 0._dp
+    if (present(flow_rate_diff)) flow_rate_diff_val = flow_rate_diff
+
+    dims = self%mesh%get_dims(f%data_loc)
+    n_mod = mod(dims(2) - 1, SZ) + 1
+    n_y_blocks = (dims(2) - 1)/SZ + 1
+
+    !$omp parallel do private(k_end, y_block, i_max)
+    do k = 1, n_y_blocks*dims(3)
+      y_block = (k - 1)/dims(3) + 1
```
ia267: I think this should be:

```suggestion
      y_block = mod(k - 1, n_y_blocks) + 1
```

OMP reordering uses `get_index_dir` (line 54), which maps a Cartesian (y, z) position to the third `DIR_X` index as:

```fortran
dir_k = n_y_blocks*(z-1) + y_block
```

Therefore, y-block is the fast-varying component. For n_y_blocks = 3 and nz = 2, increasing `k` from 1 to 6 gives y-blocks 1, 2, 3, 1, 2, 3. The current expression, (b - 1)/nz + 1 produces 1, 1, 2, 2, 3, 3 (opposite flattening order). For CUDA it calls `reorder_cuda` (doesn't call `get_index_dir`). The current expression assumes CUDA's different layout. 

Also, I think in `field_set_face_omp` (line 848), the `Y_FACE` block also uses the same incorrect CUDA-style indexing and should be changed to:

```fortran
case (Y_FACE)
  i_mod = mod(dims(2) - 1, SZ) + 1
  n_y_blocks = (dims(2) - 1)/SZ + 1

  !$omp parallel do private(k_start, k_end)
  do z = 1, dims(3)
    k_start = 1 + (z - 1)*n_y_blocks
    k_end = n_y_blocks + (z - 1)*n_y_blocks

    do j = 1, dims(1)
      f%data(1, j, k_start) = c_start
      f%data(i_mod, j, k_end) = c_end
    end do
  end do
  !$omp end parallel do
```

This follows the OMP's `DIR_X` layout.

src/backend/omp/backend.f90
ia267: Agreed, thanks. I was inferring from OMP layout from `field_set_face_omp`, but that routine itself uses CUDA-style z-fast indexing and is inconsistent with `reorder_omp`.

`reorder_omp` calls `get_index_reordering`, which uses `get_index_dir` so the existing y-block-fast formula here is correct here. I'll resolve this comment and raise the incorrect OMP face indexing separately.

src/backend/cuda/backend.f90
ia267: I checked NVHPC 25.3 docs. CUDA Fortran supports `atomicmax` for `real(8)` so this is OK. 

See https://docs.nvidia.com/hpc-sdk/archive/25.3/compilers/cuda-fortran-prog-guide/index.html#atomic-functions

"For atomicadd, atomicsub, atomicmax, atomicmin, and atomicexch, the data types may be integer(4), integer(8), real(4), or real(8). For atomicand, atomicor, and atomicxor, only integer(4) arguments are supported."

src/backend/omp/backend.f90
```diff
@@ -863,6 +864,55 @@ subroutine field_set_face_omp(self, f, c_start, c_end, face, &
...
+    integer, optional, intent(in) :: bc_start, bc_end
+    real(dp), optional, intent(in) :: flow_rate_diff
+    integer :: dims(3), k, i, i_max, n_mod, n_y_blocks, y_block, k_end
+    real(dp) :: flow_rate_diff_val
+
+    if (f%dir /= DIR_X) &
+      error stop 'field_set_face_from_field: only supported for DIR_X fields.'
+    if (f%data_loc == NULL_LOC) &
+      error stop 'field_set_face_from_field: requires a valid data_loc.'
+    if (face /= X_FACE) &
```
ia267: The channel case calls this routine with `Y_FACE` - we will need to implement `Y_FACE` path using OMP ordering:

```fortran
n_mod = mod(dims(2) - 1, SZ) + 1
n_y_blocks = (dims(2) - 1)/SZ + 1

!$omp parallel do private(k_start, k_end)
do z = 1, dims(3)
  k_start = 1 + (z - 1)*n_y_blocks
  k_end = n_y_blocks + (z - 1)*n_y_blocks

  do j = 1, dims(1)
    f%data(1, j, k_start) = f_start%data(1, j, k_start)
    f%data(n_mod, j, k_end) = f_start%data(n_mod, j, k_end)
  end do
end do
!$omp end parallel do
```

ia267 (changes requested):
The OMP boundary path incorrectly assumes CUDA’s z-fast block ordering. OMP uses y-block-fast ordering through get_index_dir, so the `y_block` calculation and Y-face indices must be corrected accordingly. Additionally, `field_set_face_from_field_omp` must support `Y_FACE`, since the channel case calls that path.

ia267 (approved):
Approved - please fix the formatting issues before merging.

kaanolgu:
@x3d2bot fix style

PR #310, opened 2026-06-04
Add per timestep execution time report

kaanolgu:
Adding per timestep average execution time printing like Incompact3d, 
`do_report` condition is set according to the input file variables so we only print every `nth` iteration
The printed `Time for this time step (s)` is the execution time of that one single iteration on which the report fires.

### DIFF
--- a/src/case/base_case.f90
+++ b/src/case/base_case.f90
@@ -5,7 +5,7 @@ module m_base_case
 
   use m_allocator, only: allocator_t
   use m_base_backend, only: base_backend_t
-  use m_common, only: dp, DIR_X, DIR_C, VERT
+  use m_common, only: dp, DIR_X, DIR_C, VERT, MPI_X3D2_DP
   use m_monitoring, only: monitoring_t
   use m_field, only: field_t, flist_t
   use m_mesh, only: mesh_t
@@ -13,7 +13,7 @@ module m_base_case
   use m_postprocess, only: compute_derived_fields, compute_pressure_vert
   use m_config, only: has_output_field
   use m_io_manager, only: io_manager_t
-  use mpi, only: MPI_COMM_WORLD
+  use mpi, only: MPI_COMM_WORLD, MPI_Wtime, MPI_Reduce, MPI_MAX
 
   implicit none
 
@@ -161,7 +161,7 @@ contains
 
   end subroutine set_init
 
-  subroutine run(self)
+subroutine run(self)
     !! Runs the solver forwards in time from t=t_0 to t=T, performing
     !! postprocessing/IO and reporting diagnostics.
     implicit none
@@ -174,9 +174,10 @@ contains
     real(dp) :: t
     integer :: i, iter, sub_iter, start_iter
     logical :: output_vorticity, output_qcriterion
-    real(dp) :: t_start, t_step0, t_step1, t_end
-    real(dp) :: t_total
-    logical :: do_report
+    real(dp) :: t_start, t_end, t_total_local, t_total
+    real(dp) :: t_step0, t_step1, t_step_local, t_step
+    logical :: do_timing_report, is_root
+    integer :: ierr
 
     output_vorticity = &
       has_output_field(self%io_mgr%snapshot_mgr%config, 'vorticity') &
@@ -191,6 +192,10 @@ contains
         ! for restarts current_iter is read from the checkpoint file
         print *, 'Continuing from iteration:', &
         self%solver%current_iter, 'at time ', t
+      if (self%solver%n_iters <= self%solver%current_iter) then
+        error stop &
+          'Restart requires n_iters greater than the restart iteration.'
+      end if
     else
       self%solver%current_iter = 0
       if (self%solver%mesh%par%is_root()) print *, 'initial conditions'
@@ -212,18 +217,23 @@ contains
       curr(3 + i)%ptr => self%solver%species(i)%ptr
     end do
 
-    call cpu_time(t_start)
+    is_root = self%solver%mesh%par%is_root()
+    t_start = MPI_Wtime()
 
     do iter = start_iter, self%solver%n_iters
-      ! nested guard so mod() is never evaluated when n_output == 0
-      do_report = (iter == start_iter .or. iter == self%solver%n_iters)
+      ! Report per-step timing on the final step and on every n_output-th
+      ! step. This is computed identically on every rank so that the
+      ! reduction below stays collective; the nested guard keeps mod()
+      ! from being evaluated when n_output is 0.
+      do_timing_report = (iter == self%solver%n_iters)
       if (self%solver%n_output > 0) then
-        if (mod(iter, self%solver%n_output) == 0) do_report = .true.
+        if (mod(iter, self%solver%n_output) == 0) do_timing_report = .true.
       end if
-      do_report = do_report .and. self%solver%mesh%par%is_root()
 
-      if (do_report) then
-        call cpu_time(t_step0)
+      ! Skip the first step: it carries one-off setup costs that would
+      ! make the per-step figure unrepresentative.
+      if (do_timing_report .and. iter > start_iter) then
+        t_step0 = MPI_Wtime()
       end if
       do sub_iter = 1, self%solver%time_integrator%nstage
         ! first apply case-specific BCs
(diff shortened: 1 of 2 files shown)


kaanolgu:
@x3d2bot fix style

src/case/base_case.f90
```diff
@@ -257,7 +260,19 @@ subroutine run(self)
...
 
+    call cpu_time(t_start)
+
     do iter = start_iter, self%solver%n_iters
+      ! nested guard so mod() is never evaluated when n_output == 0
+      do_report = (iter == start_iter .or. iter == self%solver%n_iters)
+      if (self%solver%n_output > 0) then
+        if (mod(iter, self%solver%n_output) == 0) do_report = .true.
+      end if
+      do_report = do_report .and. self%solver%mesh%par%is_root()
```
ia267: Current `do_report` combines both `do_report` and `is_root` - this is fine for the cpu time stuff, but it will also affect postprocess (i.e. only rank 0 will call post-process, which is wrong as the post-process uses MPI collectives). We need to split the behaviour so that only timing stuff checks for is root:

Lets introduce a new variable `do_postprocess` and rename `do_report` to `do_timing_report`. 

So we can do something like:

```fortran
logical :: do_postprocess, do_timing_report, is_root
is_root = self%solver%mesh%par%is_root()

do_postprocess = .false.
if (self%solver%n_output > 0) then
  do_postprocess = mod(iter, self%solver%n_output) == 0
end if

do_timing_report = (iter == start_iter .or. iter == self%solver%n_iters)
if (self%solver%n_output > 0) then
  do_timing_report = do_timing_report .or. &
                     (mod(iter, self%solver%n_output) == 0)
end if

if (do_timing_report .and. iter > start_iter .and. is_root) then
  call cpu_time(t_step0)
end if
```

and later (outside the time-step loop)

```fortran
if (do_timing_report .and. iter > start_iter .and. is_root) then
  call cpu_time(t_step1)
  print *, 'Time for this time step (s):', real(t_step1 - t_step0)
end if

if (do_postprocess) then
  t = iter*self%solver%dt
  call self%postprocess(iter, t)
end if
```

src/case/base_case.f90
```diff
@@ -304,16 +319,28 @@ subroutine run(self)
...
                                       iter, MPI_COMM_WORLD)
     end do
 
+    ! total + averaged-per-step over the whole run
+    call cpu_time(t_end)
+    t_total = t_end - t_start
+    if (self%solver%mesh%par%is_root()) then
+      print *, '==========================================================='
+      print *, 'Averaged time per step (s):', &
+        real(t_total/(self%solver%n_iters - (start_iter - 1)), 4)
```
ia267, line 325: Here we are dividing by `n_iters - (start_iter - 1)`, for a restart case, a user may erroneously not update `n_iter` (to a value greater than the start iteration for example. I think we need to add a check for this during restart, so inside the if-statement in line 236:

```fortran
if (self%io_mgr%is_restart()) then
  ...
  if (self%solver%n_iters <= self%solver%current_iter) then
    error stop 'Restart requires n_iters greater than restart iteration.'
  end if
end if
```

src/case/base_case.f90
```diff
@@ -304,16 +319,28 @@ subroutine run(self)
...
+      print *, 'Total wallclock (s):', real(t_total, 4)
```
ia267, line 326: For parallel runs I don't think `cpu_time` will give the `wallclock` time. Note also that these calls are only made for rank0. We should use `MPI_Wtime()`:

```fortran
real(dp) :: t_start, t_end, t_total_local, t_total
real(dp) :: t_step0, t_step1, t_step_local, t_step
integer :: ierr
``` 

before the loop

```fortran
t_start = MPI_Wtime()
```
for selected timestep timing, before timestep work:

```fortran
if (do_timing_report .and. iter > start_iter) then
  t_step0 = MPI_Wtime()
end if
```

after the timestep work:

```fortran
if (do_timing_report .and. iter > start_iter) then
  t_step1 = MPI_Wtime()
  t_step_local = t_step1 - t_step0

  call MPI_Reduce(t_step_local, t_step, 1, MPI_X3D2_DP, MPI_MAX, &
                  0, MPI_COMM_WORLD, ierr)

  if (is_root) then
    print *, 'Time for this time step (s):', real(t_step, 4)
  end if
end if
```

after the loop:

```fortran
t_end = MPI_Wtime()
t_total_local = t_end - t_start

call MPI_Reduce(t_total_local, t_total, 1, MPI_X3D2_DP, MPI_MAX, &
                0, MPI_COMM_WORLD, ierr)

if (is_root) then
  print *, 'Total wallclock (s):', real(t_total, 4)
end if
```

ia267 (changes requested):

kaanolgu:
@ia267  Thanks for the hint! I scratched the surface to figure out what was happening. The Incompact3d (earlier version of the project) uses `cpu_time` to capture execution times. Incompact3d is pure MPI implementation with no OpenMP. That's why the `cpu_time` and `mpi_wtime` results match on Incompact3d:

TGV example (Incompact3d) :

Run properties ( the ones not modified are removed) :
```
! Domain decomposition
p_row=0               ! Row partition
p_col=0               ! Column partition

! Mesh
nx=128                 ! X-direction nodes
ny=128                 ! Y-direction nodes
nz=128                 ! Z-direction nodes
istret = 0            ! y mesh refinement (0:no, 1:center, 2:both sides, 3:bottom)
beta = 0.259065151    ! Refinement parameter (beta)

! Domain
xlx = 6.28 ! Lx (Size of the box in x-direction)
yly = 6.28 ! Ly (Size of the boy in y-direction)
zlz = 6.28 ! Lz (Size of the boz in z-direction)

! Boundary conditions
nclx1 = 0 
nclxn = 0 
ncly1 = 0 
nclyn = 0 
nclz1 = 0 
nclzn = 0 

! Flow parameters
iin = 1               ! Inflow conditions (1: classic, 2: turbinit)
re = 1600.            ! nu=1/re (Kinematic Viscosity)
u1 = 8.               ! u1 (max velocity) (for inflow condition)
u2 = 8.               ! u2 (min velocity) (for inflow condition)
init_noise  = 0.0     ! Turbulence intensity (1=100%) !! Initial condition
inflow_noise = 0.0    ! Turbulence intensity (1=100%) !! Inflow condition

! Time stepping
dt = 0.005            ! Time step
ifirst = 1            ! First iteration
ilast = 100          ! Last iteration

! Enable modelling tools
ilesmod=0             ! if 0 then DNS
iscalar=0             ! If iscalar=0 (no scalar), if iscalar=1 (scalar)
iibm=0                ! Flag for immersed boundary method

! Enable io
ivisu=0               ! Store snapshots
ipost=0               ! Do online postprocessing

/End

!====================
&NumOptions
!====================

! Spatial derivatives
ifirstder = 4         ! (1->2nd central, 2->4th central, 3->4th compact, 4-> 6th compact)
isecondder = 4        ! (1->2nd central, 2->4th central, 3->4th compact, 4-> 6th compact, 5->hyperviscous 6th)

! Time scheme
itimescheme = 3       ! Time integration scheme (1->Euler,2->AB2, 3->AB3, 4->AB4,5->RK3,6->RK4)

/End

!=================
&InOutParam
!=================

! Basic I/O
irestart = 0          ! Read initial flow field ?
icheckpoint = 1000   ! Frequency for writing backup file
ioutput = 1000        ! Frequency for visualization
ilist = 1000             ! Frequency for the output
nvisu = 1             ! Size for visualisation collection

/End


```

Results:
|Run  |  MPI ranks | CPU Timer (s) | MPI Time (s)|
|--|--|--|--|
| A |                   1       |  79.6073761           |  79.6122971 |
| B  |                  2        |  48.4568710         |  48.4631577 |
| C  |                  3        |  38.8231468        |  38.8286133 |
| D | 4 | 37.8947716  | 37.9008751 |
 
 
 
 If we look at the x3d2 case, the `cpu_time` was carried over from Incompact3d but not a reliable way to capture the execution times now given that x3d2 has multi thread execution with OpenMP. 
 
 The TGV example verifies this:
 
 Results: (`OMP_PLACES=cores`)
|Run|  OMP_THREADS|   MPI ranks|  CPU Timer (s) | MPI Time (s)|
|--|--|--|--|--|
| A    |    1      |        1        |   79     |       79|
| B   |    24     |        1      |    123    |        69|
| C  |     48     |        1     |     127     |       72|
| D |       6       |      1     |     120       |     67|
| E |       6        |     2    |       65        |    36|
|F |       6   |          3       |    48   |         27|

kaanolgu:
@x3d2bot fix style

src/case/base_case.f90
ia267: This has been addressed in commit cfae532184a65b457426f863ac434352f196e831

src/case/base_case.f90
ia267: This has been addressed in cfae532184a65b457426f863ac434352f196e831

src/case/base_case.f90
ia267: Fixed in cfae532184a65b457426f863ac434352f196e831 - now only timing output is restricted to `is_root`

ia267 (approved):
Approved

PR #315, opened 2026-06-15
Add extensible checkpoint and scalar output support

### LINKED ISSUE #314: Make scalar monitoring output extensible
Currently monitoring file contains a fixed set of quantities: enstrophy and div mean/max.

Some flow cases need to write additional scalar data, such as turbulent kinetic energy, dissipation or turbine measurements. Introduce a re-usbale scalar-series writer with configurable column names so these output can be added without duplicating the file handling logic.

The implementation should support restart-safe behaviour.

ia267:
#### Changes

- Add a reusable root-only writer for named scalar time-series data.
- Refactor `monitoring.csv` output to use the new writer.
- Support appending output when restarting a simulation.
- Add an interface for case- and model-specific checkpoint state.
- Allow the I/O manager to register and restore additional checkpoint state.
- Added a unit test for the scalar time-series output infrastructure

#### Motivation

Monitoring output was previously hard-coded for enstrophy and the maximum and mean velocity divergence. Some flow cases and models need additional scalar outputs, such as turbulent kinetic energy, dissipation, or turbine measurements.

The new scalar-series abstraction centralises CSV file handling while allowing each caller to select its filename and columns. Separate outputs therefore remain in separate files rather than being combined into `monitoring.csv`.

The checkpoint extension allows model-specific state to participate in normal checkpoint and restart operations without coupling the I/O layer to a particular flow case or model.

This PR does not change how monitoring is called, including the existing n_output > 0 condition. It only generalises the CSV-writing functionality.

We are not adding new columns to monitoring.csv. Instead, the PR introduces a reusable `scalar_series_t` writer, allowing users to write their own computed scalar values at each output time step to separate files with user-defined columns.

To add a new scalar, a case would compute it and create its own `scalar_series_t` object. 

closes #314

### DIFF
--- a/src/CMakeLists.txt
+++ b/src/CMakeLists.txt
@@ -9,6 +9,7 @@ set(SRC
   solver.f90
   postprocess/postprocess.f90
   postprocess/monitoring.f90
+  postprocess/scalar_series.f90
   tdsops.f90
   time_integrator.f90
   ordering.f90
@@ -61,6 +62,7 @@ set(IOSRC
   io/io_session.f90
   io/io_field_utils.f90
   io/checkpoint_manager.f90
+  io/checkpoint_state.f90
   io/snapshot_manager.f90
   io/stats.f90
   io/io_manager.f90
--- a/src/io/checkpoint_manager.f90
+++ b/src/io/checkpoint_manager.f90
@@ -21,6 +21,7 @@ module m_checkpoint_manager
   use m_solver, only: solver_t
   use m_io_session, only: reader_session_t, writer_session_t
   use m_config, only: checkpoint_config_t
+  use m_checkpoint_state, only: checkpoint_state_t
   use m_stats, only: stats_manager_t
   use m_io_field_utils, only: field_buffer_map_t, field_ptr_t, &
                               setup_field_arrays, cleanup_field_arrays, &
@@ -49,6 +50,7 @@ module m_checkpoint_manager
     procedure :: init
     procedure :: handle_restart
     procedure :: handle_checkpoint_step
+    procedure :: restore_state
     procedure :: is_restart
     procedure :: finalise
     procedure, private :: write_checkpoint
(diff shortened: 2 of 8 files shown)


kaanolgu (changes requested):
The test is not added to CMakeLists, 
```
define_test(unit/test_get_block_2d.f90 1 omp)
``` 
After the addition I received the following error messages:

```[ 69%] Linking Fortran executable bin/test_omp_transeq_omp_4
/home/kolgu/x3d2-upstream/tests/unit/test_get_block_2d.f90:16:28:

   16 |   blk => alloc%get_block_2d(DIR_C)
      |                            1
Error: ‘get_block_2d’ at (1) is not a member of the ‘allocator_t’ structure
/home/kolgu/x3d2-upstream/tests/unit/test_get_block_2d.f90:22:23:

   22 |   if (size(blk%data_2d, 1) /= ny .or. size(blk%data_2d, 2) /= nz) then
      |                       1
Error: ‘data_2d’ at (1) is not a member of the ‘field_t’ structure; did you mean ‘data’?
/home/kolgu/x3d2-upstream/tests/unit/test_get_block_2d.f90:24:5:

   24 |   end if
      |     1
Error: Expecting END PROGRAM statement at (1)
/home/kolgu/x3d2-upstream/tests/unit/test_get_block_2d.f90:25:44:

   25 |   print *, "PASS: shape", shape(blk%data_2d), "== expected", ny, nz
      |                                            1
Error: ‘data_2d’ at (1) is not a member of the ‘field_t’ structure; did you mean ‘data’?
/home/kolgu/x3d2-upstream/tests/unit/test_get_block_2d.f90:29:14:

   29 |   blk%data_2d(1, 1) = sentinel
      |              1
Error: ‘data_2d’ at (1) is not a member of the ‘field_t’ structure; did you mean ‘data’?
/home/kolgu/x3d2-upstream/tests/unit/test_get_block_2d.f90:30:17:

   30 |   if (blk%p_data(1) /= sentinel) then  ! p_data is private; expose via get if needed
      |                 1
Error: Component ‘p_data’ at (1) is a PRIVATE component of ‘field_t’
/home/kolgu/x3d2-upstream/tests/unit/test_get_block_2d.f90:32:5:

   32 |   end if
      |     1
Error: Expecting END PROGRAM statement at (1)
```

ia267:
Carried out runs to determine whether there was performance hit with writing .csv files (with enstropy and div u max/mean).

TGV case with 100 iterations and priting/writing monitoring diagnostics every 10th iteration

For CUDA (single GPU): NVIDIA RTX A4000
1. No monitoring calcs/output (n_output = 0): 18.60499 s
2. With monitoring output/stdout: 19.09602 s
3. With monitoring output/ no stdout: 19.07139 s
4. Only enstropy/div u max/mean calcs but no monitoring output or stdout: 19.09439 s

For OMP (`-np 2`): CPU Intel(R) Xeon(R) w3-243
1. No monitoring calcs/output (n_output = 0): 345.402069 s
2. With monitoring output/stdout:  356.805 s
3. With monitoring output/ no stdout:  359.7260 s
4. Only enstropy/div u max/mean calcs but no monitoring output or stdout: 358.189819 s

Writing to .csv file does not cause a significant decrease in performance. There is a 2-4% increase in runtime due to the enstropy and div u max/mean computations.

kaanolgu (approved):

PR #328, opened 2026-06-19
Fix qcrit/vort snapshots in post-process

### LINKED ISSUE #327: Vorticity and Q-criterion calculations should use device-host transfer
On CUDA backend, the derived fields (q-criterion and vorticity) calculated in `postprocess.f90` were computed from gradient field arrays without explicity device-to-host transfer, .e. these values are incorrect.

The fix is to copy gradient data via backend `get_field_data` into host, compute vort/qcrit here and write back with `set_field_data`

ia267:
This PR fixes a bug where qcrit/vort fields did not change with time (i.e. remained frozen). Fixes #327 

- In `postprocess.f90` vorticity and q-criterion were computed from gradient fields without explicitly transfering it from CUDA backend
- As a result, these derived fields written to snapshots  were unchanged over time 

#### Changes
- Add host fields for all 9 velocity gradient components
- Uses `get_field_data` to copy/reorder gradients from backend fields into host fields
- Computed vorticity/q-criterion on host data
- Uses `set_field_data` to write computed vort/qcrit back to backend fields
- Releases all temporary host blocks and backend gradient blocks
- The OMP behaviour is correct

#### Future considerations
- Additional transfer of data from device to host - need to be optimised via kernels for larger simulations. Alternatively, we create a python module that does it after the simulation (similar to Incompact3d)

#### Testing 
Cylinder case at two different time-steps showing qcrit (coloured with u-velocity)

Screenshot from 2026-06-20 00-25-37
Screenshot from 2026-06-20 00-25-27

### DIFF
--- a/src/postprocess/postprocess.f90
+++ b/src/postprocess/postprocess.f90
@@ -6,7 +6,7 @@ module m_postprocess
   !! and visualisation (e.g. pressure on the vertex grid, vorticity
   !! magnitude, Q-criterion).
 
-  use m_common, only: dp, DIR_X, DIR_Y, DIR_Z, VERT, &
+  use m_common, only: dp, DIR_X, DIR_Y, DIR_Z, DIR_C, VERT, &
                       RDR_X2Y, RDR_X2Z, RDR_Y2X, RDR_Z2X
   use m_field, only: field_t
   use m_solver, only: solver_t
@@ -38,6 +38,11 @@ contains
       u_y, dudy_y, u_z, dudz_z, &
       v_y, dvdy_y, v_z, dvdz_z, &
       w_y, dwdy_y, w_z, dwdz_z
+    class(field_t), pointer :: &
+      dudx_h, dudy_h, dudz_h, &
+      dvdx_h, dvdy_h, dvdz_h, &
+      dwdx_h, dwdy_h, dwdz_h, &
+      vort_h, qcrit_h
 
     ! Allocate output fields on first use
     if (output_vorticity .and. .not. associated(solver%vort)) &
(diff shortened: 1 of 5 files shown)


src/postprocess/postprocess.f90
```diff
@@ -132,31 +137,67 @@ subroutine compute_derived_fields(solver, output_vorticity, &
...
     if (output_qcriterion) then
-      solver%qcrit%data = &
-        -0.5_dp*(dudx%data**2 &
-                 + dvdy_x%data**2 &
-                 + dwdz_x%data**2) &
-        - dudy_x%data*dvdx%data &
-        - dudz_x%data*dwdx%data &
-        - dvdz_x%data*dwdy_x%data
+      qcrit_h => solver%host_allocator%get_block(DIR_C, solver%qcrit%data_loc)
+      qcrit_h%data = &
```
kaanolgu: It would be better to make this backend dependent because with GPU, the overhead of moving 9 different data to host for calculation in every postprocessing step is going to be costly.  

The following are the execution time differences for a small example which I anticipate would make huge difference for a longer simulation run: 

```
// TGV Case Base ( 128^3)
Re = 1600d0
time_intg = 'AB3' ! 'AB[1-4]' | 'RK[1-4]'
dt = 0.001d0
n_iters = 100
n_output = 5
```

```
// TGV Case Base Snapshot Enabled Base ( 128^3)
Re = 1600d0
time_intg = 'AB3' ! 'AB[1-4]' | 'RK[1-4]'
dt = 0.001d0
n_iters = 100
n_output = 5

&checkpoint_params
snapshot_freq = 5
/End
```


```
// TGV Case This PR ( 128^3)
Re = 1600d0
time_intg = 'AB3' ! 'AB[1-4]' | 'RK[1-4]'
dt = 0.001d0
n_iters = 100
n_output = 5

&checkpoint_params
snapshot_freq = 5
output_fields = 'vorticity', 'qcriterion'
/End
```


| Method       | GPUs  | Execution Time (s) |
|--------------|-------|--------------------|
| Base         | 1 GPU | 1.975934           |
| Snp Base     | 1 GPU | 3.608796           |
| This PR      | 1 GPU | 10.93888           |

kaanolgu (changes requested):

src/postprocess/postprocess.f90
ia267: I've made the changes to compute voriticity/q-criterion post-processing on the device to avoid device to host copies. For the TGV case this is what I get (100 iterations, with snapshot frequency set to 5):

| Method	| GPUs	| Execution Time (s) |
| :--- | :--- | :--- |
| Without vort & qrit	|1 GPU	| 37.42250 |
| With vort & qcrit	| 1 GPU	| 51.69613 |

This is still a 38% slow down with these outputs. For large simulations, this is not practical and might be better to post-process after the simulation (similar to Incompact3d).

kaanolgu (approved):
On `Kolmogorov` tested both the CPU and GPU. The results match as previously mentioned cases:

The performance comparisons for reference:

| Method       | GPUs  | Execution Time (s) |
|--------------|-------|--------------------|
| Snp Base     | 1 GPU | 3.633177            |
| This PR      | 1 GPU | 5.055154           |


| Method       | CPUs  | Execution Time (s) |
|--------------|-------|--------------------|
| Snp Base     | 1 CPU | 70.1194611            |
| This PR      | 1 CPU | 93.6137390          |

Issue #78, opened 2024-03-18
Investigate OMP reordering performance


Nanoseb:
The OMP backend is using a single function for any reorder. Specialised functions may offer better performance (better inlining, remove of the `select` statements etc.). 
See https://github.com/xcompact3d/x3d2/pull/68#discussion_r1528187837

ia267:
This was implemented in #329

PR #329, opened 2026-06-23
Faster OpenMP Reorder and Accumulation Kernels

kaanolgu:
Speeds up the OpenMP `reorder` and `sum_intox` kernels by removing the per element index computation. The transpose used to call `get_index_reordering` for every element, which nests two more routines doing integer `mod` and division per element and blocks vectorisation

This addresses the same bottleneck as #131 (related to #78), where reorders and accumulations were measured at roughly half of the transeq runtime while the per element path sustains only 10 to 20% of peak bandwidth. #131 specialises the hot transeq mappings with explicit cache blocked kernels. This PR takes an alternative, more general route: the index map advances by a fixed stride along the leading index, so the kernels compute a base and stride once per column and run a plain inner loop. Because the same routine covers every directional mapping rather than a selected few, the speedup extends across the divergence and gradient reorders as well as transeq, broadening the gain over the full step while keeping the implementation compact

Padded dimensions are exact multiples of `SZ`, so the leading index stays inside one `SZ` block and never crosses a `mod` or division boundary. One output index then advances by a fixed stride while the other two stay constant. Evaluating the map at `i = 1` and `i = 2` recovers the base and stride, and the inner loop reproduces the original writes exactly. This needs `dims(1) == SZ` (a directional field). A `DIR_C` field has a leading dimension of `nx_padded`, so the kernels guard on `dims(1) == SZ` and fall back to the per element loop there

## Performance (TGV 256^3, 100 iterations, generic 2decomp FFT)

1 rank, 24 threads:

| phase       | before (s) | after (s) | change |
|-------------|-----------:|----------:|-------:|
| transeq     |       22.0 |      17.3 |   -21% |
| divergence  |       10.0 |      7.32 |   -27% |
| gradient    |        9.5 |      6.79 |   -28% |
| poisson FFT |        113 |       111 | (untouched) |

8 ranks, 3 threads each:

| phase        | before (s) | after (s) | change |
|--------------|-----------:|----------:|-------:|
| transeq      |       19.1 |     15.35 |   -20% |
| divergence   |       8.55 |      6.22 |   -27% |
| gradient     |        7.8 |      5.68 |   -27% |
| poisson FFT  |       18.6 |     18.44 | (untouched) |
| **avg step** |      0.576 |     0.508 | **-12%** |

At one rank the generic FFT is about 74% of the step so the total barely moves. At eight ranks the FFT has scaled down and the saving becomes a 12% step reduction, so the benefit grows with rank count


ia267 (approved):
LGTM!

Issue #332, opened 2026-07-07
Ensure portability of OpenMP Target backend across Nividia and AMD GPUs


CFD-Xing:
The `has_device_addr` pragma is not portable to Nvidia architecture,  a more portable way would be to use `is_device_ptr` on the actual device pointer and to map the pointer to a Fortran within the OMP directive

CFD-Xing:
closing as addressed by #335

PR #335, opened 2026-08-03
332 ensure portability of openmp target backend across nividia and amd gpus

pbartholomew08:
I'm opening this for visibility/discussion.

I think this implements the changes suggested by #332. However, with this I'm no longer able to compile the code with Cray (cce 20) or AMD (flang 22) compilers due to `call c_f_pointer` not being offloaded and `is_device_addr` not being supported, respectively. Of course, there may be a mistake in the implementation in which case I'll correct it and retest.

### DIFF
--- a/src/backend/omp/target/allocator.f90
+++ b/src/backend/omp/target/allocator.f90
@@ -155,6 +155,7 @@ contains
   function get_shape_omptgt(self) result(dims)
     class(omptgt_field_t) :: self
     integer :: dims(3)
+    !$omp declare target
 
     dims = self%dims
   end function
--- a/src/backend/omp/target/allocator.f90
+++ b/src/backend/omp/target/allocator.f90
@@ -33,14 +33,15 @@ module m_omptgt_allocator
 
   type, extends(field_t) :: omptgt_field_t
     ! A device-resident field
+    integer, private :: n
+    integer, dimension(3), private :: dims
     integer, private :: dev_id
     type(c_ptr), private :: dev_ptr = c_null_ptr
-    real(dp), pointer, private :: p_data_tgt(:) => null()
-    real(dp), pointer, contiguous :: data_tgt(:, :, :) => null()
   contains
     procedure :: fill => fill_omptgt
     procedure :: get_shape => get_shape_omptgt
     procedure :: set_shape => set_shape_omptgt
+    procedure :: get_dev_ptr
     final :: omptgt_field_destroy
   end type omptgt_field_t
 
@@ -87,16 +88,14 @@ contains
       error stop "Device ID is HOST"
     end if
 
-    f%dev_ptr = omp_target_alloc(ngrid*c_sizeof(0.0_dp), f%dev_id)
-    call c_f_pointer(f%dev_ptr, f%p_data_tgt, shape=[ngrid])
+    f%n = ngrid
+    f%dims = -1
+    f%dev_ptr = omp_target_alloc(f%n*c_sizeof(0.0_dp), f%dev_id)
 
   end subroutine omptgt_field_init
 
   subroutine omptgt_field_destroy(self)
     type(omptgt_field_t) :: self
-
-    nullify (self%data_tgt)
-    nullify (self%p_data_tgt)
 
     if (c_associated(self%dev_ptr)) then
       call omp_target_free(self%dev_ptr, self%dev_id)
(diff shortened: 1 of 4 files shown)


CFD-Xing:
I did some investigation and it seems that for the Cray compiler, the device pointer MUST be a dummy argument. So, having something like this does the trick:

```
  subroutine vecadd_offload_helper(self, a, cp_x, dims_x, b, cp_y, dims_y, dims)

    class(omptgt_backend_t) :: self
    real(dp), intent(in) :: a
    type(c_ptr), intent(in):: cp_x
    integer, intent(in) :: dims_x(3)
    real(dp), intent(in) :: b
    type(c_ptr), intent(in):: cp_y
    integer, intent(in) :: dims_y(3)
    integer, intent(in) :: dims(3)

    integer :: i, j, k
    real(dp), dimension(:,:,:), pointer :: p_x, p_y

    !$omp target teams is_device_ptr(cp_x, cp_y)
    call c_f_pointer(cp_x, p_x, shape=dims_x)
    call c_f_pointer(cp_y, p_y, shape=dims_y)
    !$omp loop collapse(3)
    do k = 1, dims(3)
      do j = 1, dims(2)
        do i = 1, dims(1)
          p_y(i, j, k) = a*p_x(i, j, k) + b*p_y(i, j, k)
        end do
      end do
    end do
    !$omp end loop
    !$omp end target teams

  end subroutine
```

```
  subroutine vecadd_offload(self, a, x, b, y)

    class(omptgt_backend_t) :: self
    real(dp), intent(in) :: a
    type(omptgt_field_t), intent(in) :: x
    real(dp), intent(in) :: b
    type(omptgt_field_t), intent(inout) :: y

    type(c_ptr) :: cp_x, cp_y
    integer :: dims_x(3), dims_y(3)
    integer :: dims(3)

    dims = self%allocator%get_padded_dims(x%dir)

    cp_x = x%get_dev_ptr()
    cp_y = y%get_dev_ptr()
    dims_x = x%get_shape()
    dims_y = y%get_shape()

    call vecadd_offload_helper(self, a, cp_x, dims_x, b, cp_y, dims_y, dims)
  end subroutine
```

EDIT: Sorry, I am probably experiencing a different issue as I am compiling with Cray 16 on Archer2.

CFD-Xing:
@pbartholomew08 What facilities are you using to compile with cce 20 and flang22? I am trying to compile on ARCHER2 and the compilation fails because ROCM_PATH can't be found. Rather than spending time to figure out how to compile on ARCHER2, I think I am better off to use a more recent facility.

pbartholomew08:
Hi @CFD-Xing - I've not had a chance to try your suggested changes, I'll hopefully do tomorrow. I'm building using the latest flang I can get from AMD on Dirac, and I think if you log into a GPU node on ARCHER2 you should get CCE 20 for the Cray compilers

CFD-Xing:
@pbartholomew08 Ok thanks for the information. 

EDIT: I don't think I can access GPU node on ARCHER2. 

PS: You can use my patch to give a quick try if you want.
[omp.patch](https://github.com/user-attachments/files/30793551/omp.patch)

pbartholomew08:
Thanks @CFD-Xing - I've tried building with the patch on ARCHER2 using the Cray compiler and still have the issue around offloading `c_f_pointer` - have you been able to build this under amdflang, or something that isn't nvhpc?

CFD-Xing:
@pbartholomew08 I finally managed to sort out how to compile on ARCHER2. The previous error that I was seeing was probably due to using CCE 16 instead of CCE 20. And after looking into it, it seems that the Cray compiler does not support calling c_f_pointer inside a OMP TARGET construct... 


EDIT:  Spend a large part of the day looking into this and I think that there is no way around with the Cray compiler. `c_f_pointer` must be outside the target directive.  NVHPC seems to support both, just tested locally on my laptop with NVHPC 25.11. However the following test is failing:  `test_omptgt_allocator_omp_tgt_1`

Here is a patch fixing a bug and moves the `c_f_pointer` outside the target directive.
[patch.patch](https://github.com/user-attachments/files/30845621/patch.patch)

src/backend/omp/target/allocator.f90
```diff
@@ -107,60 +106,77 @@ subroutine fill_omptgt(self, c)
...
-    dims = shape(self%data_tgt)
+    dims = self%dims
   end function
 
   subroutine set_shape_omptgt(self, dims)
     class(omptgt_field_t) :: self
     integer, intent(in) :: dims(3)
 
-    call c_f_pointer(self%dev_ptr, self%data_tgt, shape=dims)
+    if (product(dims) < self%n) then
```
CFD-Xing, line 167: This should be  `  if (product(dims) <= self%n) then`

CFD-Xing:
@pbartholomew08 With some help from Claude, I think I managed  to have something working with both NVHPC and Cray: [patch2.patch](https://github.com/user-attachments/files/31833922/patch2.patch)

CFD-Xing:
> @pbartholomew08 With some help from Claude, I think I managed to have something working with both NVHPC and Cray: [patch2.patch](https://github.com/user-attachments/files/31833922/patch2.patch)

@pbartholomew08 Here is the updated patch. It also fixes the bug in [src/backend/omp/target/allocator.f90](https://github.com/xcompact3d/x3d2/pull/335/files/7840bc9b64cda7eb3ed478ce666ea4c8f89e3bcc#diff-ae181416d6d153f2baab486f1965bd45b6c112eb279b11197e8eb26133df0e06)

CFD-Xing:
@ia267 I have merged main into this branch and updated it. The branch is now working with both NVHPC and CRAY compilers. I have also tested it on our GitLab CI: https://gitlab.ae.ic.ac.uk/xcompact3d/x3d2/-/pipelines/215

ia267 (approved):
Approved

PR #336, opened 2026-08-06
Fix single-precision test failures and spectral Poisson thresholds

### LINKED ISSUE #210: Bug with single precision
#206 added single precision support, but there is a bug in x3d2 when we compile with single precision support. Here is the list of tests that fail:
```
The following tests FAILED:
	  8 - test_time_integrator_omp_1 (Failed)
	  9 - test_omp_tridiag_omp_4 (Failed)
	 11 - test_omp_transeq_omp_4 (Failed)
	 12 - test_omp_transeq_species_omp_4 (Failed)
	 13 - test_omp_dist_transeq_omp_4 (Failed)
	 19 - test_time_integrator_cuda_1 (Failed)
	 20 - test_fft_cuda_1 (Failed)
	 23 - test_thom_cuda_1 (Failed)
	 25 - test_cuda_transeq_cuda_4 (Failed)
```
I would start by debugging the time integrator test first, as it only uses vecadd and the vecadd test itself is passing. 
Also, the cuda test with the tridiagonal solver passes, and omp test with the thomas algorithm passes, so I guess there is nothing fundamentally wrong with the distd2 algorithm or thomas algorithm with single precision.

ia267:
From issue #210 the following test cases were failing in single-precision:

```
OMP backend:

 5 - test_mesh_omp_4
12 - test_statistics_omp_1
17 - test_thom_omp_1
18 - test_omp_tridiag_omp_4
19 - test_omp_penta_omp_4
20 - test_omp_transeq_omp_4
21 - test_omp_transeq_species_omp_4
22 - test_omp_dist_transeq_omp_4

CUDA backend:

34 - test_poisson_bc_cuda_1     <----- exposed a real bug
35 - test_cuda_tridiag_cuda_4
36 - test_cuda_penta_cuda_4
37 - test_cuda_transeq_cuda_4
```

With the exception of `test_poisson_bc_cuda_1`, these failures were caused by test tolerances that were not suitable for single precision. Each failing case was compared with its theoretically expected single-precision roundoff floor, and the observed errors were consistent with floating-point roundoff rather than implementation errors.

Several tests used hardcoded tolerances such as `tol = 1e-8`, which are appropriate for double precision but can be below the achievable error floor in single precision. A derivative operator amplifies input roundoff by approximately `1/dx` for a first derivative and `1/dx^2` for a second derivative. At `n_glob = 1024`, single-precision epsilon (`eps_sp ~ 1.19e-7`) therefore produces an irreducible error floor of approximately `3e-3` for second derivatives.

The following table compares the observed errors with the expected single-precision roundoff floors:

| Test | Check | Observed error | Predicted floor | Amplification mechanism |
|---|---|---|---|---|
| test_thom (n_glob=1024) | 2nd deriv, periodic | 3.5e-3 | ~3.2e-3 | `eps/dx^2`, `dx = 2pi/1024` |
| test_thom (n_glob=1024) | 2nd deriv, Dirichlet | 8.6e-3 | ~3e-3 (+boundary rows) | `eps/dx^2` |
| test_omp_tridiag | 1st deriv | 1.0–1.3e-5 | ~2e-5 | `eps/dx` |
| test_omp_tridiag | 2nd deriv | 3.2e-3 | ~3.2e-3 | `eps/dx^2` |
| test_omp_tridiag | interpolation | 9.2e-8 | ~1.2e-7 | `eps` (no amplification) |
| test_omp_tridiag | staggered deriv | 2.4–2.7e-5 | ~2e-5 | `eps/dx` |
| test_omp_tridiag | hyperviscous 2nd deriv | 0.57 | ~0.2 | `eps/dx^2 x nu0_nu (=63)` |
| transeq family (96³) | conv + diff RHS | 1.1–6.0e-4 | ~3e-5 × O(10) coeffs | `eps/dx^2` |
| test_statistics | running mean, 200 samples | 2.8e-9 | <=2e-5 | accumulation roundoff |
| test_mesh | `geo%d(Y) - 0.25` | 1.5e-8 | 1 ulp of 0.25 | fast-math reciprocal |

#### Fast-math behaviour in `test_math`
One test case (`test_mesh_omp_4`) expected the exactly representable value `1/4 = 0.25`. In a `Release` build, however, `-ffast-math` allows GCC to vectorise divisions using hardware reciprocal approximations. The resulting value may differ by approximately one ulp. For example with `-ffast-math`, the runtime division can produce: `2.49999985E-01` instead of `2.50000000E-01`. This behaviour was only found to affect this particular test. The `-ffast-math` flag is retained for performance reasons, and the tolerance is now scaled with the working precision.

#### Spectral Poisson solver bug
A real bug was found in both backends’ spectral Poisson kernels. The kernels used a hardcoded threshold when testing modified wave numbers for zero.

These wave numbers include staggered-interpolation transfer functions that vanish identically at Nyquist-type modes. In double precision, the corresponding entries evaluate to approximately `1e-20` because of roundoff and are therefore below the existing `1e-16` guard. In single precision, however, the same expressions leave approximately `1e-13` of roundoff noise, which is above that guard.

As a result, the solver divided the FFT of the RHS by values of approximately `1e-13`. In `test_poisson_bc`, this produced normalised errors as large as `6.9e+3`.

The threshold now scales with the working precision using the module-level parameter: `eps_wave = epsilon(1._dp)`. This evaluates to approximately `1.2e-7` in single precision and `2.2e-16` in double precision. The hardcoded `1.e-16_dp` guards have been replaced with `eps_wave` in both backends. After this fix, the affected error drops from `6.9e+3` to `7.1e-6`.

#### `test_poisson_bc` tolerance
While retuning `test_poisson_bc`, a related constraint was identified: its tolerance must lie between the single-precision roundoff floor of passing cases -- approximately `3e-7` in the test’s normalised norm -- and the expected aliasing error of approximately `2.4e-6`. The tolerance is now `1e-6`. This leaves a margin of roughly three on either side. If the test becomes flaky in the future, this narrow separation should be revisited.

After these fixes, all tests pass in single precision with both the OMP and CUDA backends. Closes #210.

### DIFF
--- a/src/backend/cuda/kernels/spectral_processing.f90
+++ b/src/backend/cuda/kernels/spectral_processing.f90
@@ -5,6 +5,12 @@ module m_cuda_spectral
 
   implicit none
 
+  ! Threshold below which a spectral coefficient is treated as zero. Must
+  ! scale with the working precision: single-precision roundoff leaves
+  ! ~1e-13 noise in wave numbers that should vanish (e.g. Nyquist modes),
+  ! and dividing by that noise corrupts the pressure field.
+  real(dp), parameter :: eps_wave = epsilon(1._dp)
+
 contains
 
   attributes(global) subroutine memcpy3D(dst, src, nx, ny, nz)
@@ -186,7 +192,7 @@ contains
         ! Solve Poisson
         tmp_r = real(waves(i, j, k), kind=dp)
         tmp_c = aimag(waves(i, j, k))
-        if ((tmp_r < 1.e-16_dp) .or. (tmp_c < 1.e-16_dp)) then
+        if ((tmp_r < eps_wave) .or. (tmp_c < eps_wave)) then
           div_r = 0._dp; div_c = 0._dp
         else
           div_r = -div_r/tmp_r
@@ -311,12 +317,12 @@ contains
 
         tmp_r = real(waves(i, j, k), kind=dp)
         tmp_c = aimag(waves(i, j, k))
-        if (abs(tmp_r) < 1.e-16_dp) then
+        if (abs(tmp_r) < eps_wave) then
           div_r = 0._dp
         else
           div_r = -div_r/tmp_r
         end if
-        if (abs(tmp_c) < 1.e-16_dp) then
+        if (abs(tmp_c) < eps_wave) then
           div_c = 0._dp
         else
           div_c = -div_c/tmp_c
@@ -480,12 +486,11 @@ contains
     integer, value, intent(in) :: nx_spec, n, nx, ny, nz
 
     integer :: i, j, k, jm, nm
-    real(dp) :: tmp_r, tmp_c, div_r, div_c, epsilon
+    real(dp) :: tmp_r, tmp_c, div_r, div_c
 
     i = threadIdx%x + (blockIdx%x - 1)*blockDim%x
     k = blockIdx%y ! nz_spec
 
-    epsilon = 1.e-16_dp
 
     ! Solve Poisson
     if (i <= nx_spec) then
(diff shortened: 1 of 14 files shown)


CFD-Xing (commented):
Looks good to me.

PR #337, opened 2026-08-11
fix nvlink cudadevrt ordering error on NVHPC GPU builds

ia267:
This problem was produced on NVHPC 23.1 and CMake 3.26.5, however, it is a real issue that might cause issues even with later versions of NVHPC. 

The code was calling `find_package(OpenMP)` in `src/CMakeLists.txt` after `-cuda` is added to `CMAKE_Fortran_FLAGS`. Because `-cuda` was active `FindOpenMP` tested the compiler with it and picked up NVHPC's CUDA runtime libaries (`cudafor_120;cudafor;acchost;cudadevrt;cudart;cudafor2`) as part of OpenMP. These then landed on every link line with `libcudadevrt.a` not last. `nvlink` needs `cudadevrt` to come last, so the link failed with: `nvlink fatal: unexpected object after cudadevrt`

#### Fix
- Find OpenMP just once, at the top level, so `src/` and `tests/` share one target. This also removes the duplicate `find_package(OpenMP)` calls that kept overwriting it.
- For NVHPC/PGI, remove the `cuda*` libraries from the OpenMP target. Each target already passes `-cuda`, which links the CUDA runtime itself with `cudadevrt` in the right place (last).

The cleanup runs on the finished target, so it works whether or not `-cuda` was active when OpenMP was found. Only the host OpenMP library (`libacchost.so`) is kept, and `-mp` still comes from the compile flags - so OpenMP still works, and
non-NVHPC/PGI compilers are untouched.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -42,5 +42,21 @@ if (${OMP_TGT} AND NOT
 endif()
 
 set(CMAKE_Fortran_MODULE_DIRECTORY ${PROJECT_BINARY_DIR}/include)
+
+# Resolve OpenMP once here so src/ and tests/ share a single import target.
+find_package(OpenMP REQUIRED)
+if(CMAKE_Fortran_COMPILER_ID STREQUAL "NVHPC" OR
+   CMAKE_Fortran_COMPILER_ID STREQUAL "PGI")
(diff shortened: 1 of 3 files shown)


kaanolgu (approved):
Reproduced the error and tested the fix

PR #338, opened 2026-08-11
Add static Smagorinsky LES model

### LINKED ISSUE #321: Add Smagorinsky LES with wall damping
Following on from Incompact3d, start with static Smagorinsky + wall damping. Backend-agnostic module selectable from config.

#### Scope
- Strain-rate tensor `S_ij` from the resolved velocity and its magnitude `|S|`
- Smagorinsky eddy viscosity $\nu_t = (C_s \Delta)^2 |S|$, with the filter width $\Delta$ from the local mesh spacing
- Wall damping for `nut` $\nu_t$ near a wall - the Mason-Thomson mixing-length form `(C0, n) = (0.14, 3)` used in Deskos (2019) . This is what supresses the log-law overshoot and is required for the ABL validation to reproduce the curves from Deskos (2019)
- Add the SGS  stress divergence $\partial_{x_j}(2 \nu_t S_{ij})$  to the momentum RHS (via `forcings`/`transeq`) on both OMP and CUDA backends
- Expose the computed `nut` field so the ABL wall model #317 can read it
- Config: LES model selector + Smagorinsky constant + wall-damping `(C0, n)` in `src/config.f90`

#### Validation
- Validate against TGV at Re=5000, full periodic on `[0, 2pi]^3` using existing 'case_tgv_t`. At Re=5000 the flow is under-resolved.
- Diagonostic: the dissipation rate over the transition/breakdown $t \approx 0\text{-}20$. The enstrophy based form $\varepsilon(t) = 2\nu\,E(t)$ is available directly. The kinetic-energy form $\varepsilon(t) = -dE_k/dt$ needs $E_k$ added to the monitoring output
- Reference: a well-resolved DNS $\varepsilon(t)$ curve 
- Criterion: the coarse-mesh Smagorinsky LES reproduces DNS dissipation rate peak (timing + magnitude) within tolerance, where the coarse-mesh no-model does not - i.e. the SGS model demonstrably improves the under-resolved result toward the DNS

#### Acceptance criteria
- [ ] `m_les` is backend agnostic
- [ ] Smagorinsky `nut` + SGS stress divergence computed on both backends; unit test on a known strain field recovers the analytical `nut`
- [ ] Wall damping reduces `nut` to ~0 at the wall with the `(0.14, 3)` profile
- [ ] TGV Re=5000 LES reproduces the DNS dissipation-rate peak $\varepsilon(t)$ (timing + magnitude) within tolerance, improving on the coarse no-model run
- [ ] Eddy-viscosity field exposed for the ABL wall model

ia267:
#### Summary
- Add static Smagorinsky LES model with optional wall damping
- Couple the SGS stress divergence into the momentum RHS for OpenMP and CUDA
- Add kinetic-energy for monitoring and reduce the cost of kinetic energy and enstrophy calculations
- Add unit/verification tests and update the input-file documentation
- Add `input_static_smagorinsky.x3d` as an LES TGV example
- Closes #321 

#### Validation
Validated in double precision against the reference DNS data. The reference DNS is the $1280^3$, $Re=5000$ TGV calculation reported by Dairay et al.:

> T. Dairay, E. Lamballais, S. Laizet, and J. C. Vassilicos,
> “Numerical dissipation vs. subgrid-scale modelling for large eddy
> simulation,” *Journal of Computational Physics*, 337 (2017), 252–274.  
> https://doi.org/10.1016/j.jcp.2017.02.035

The reference data were taken from the corresponding Incompact3d [`TGV_Re5000.dat`](https://github.com/xcompact3d/Incompact3d/blob/abb010e615cff520f949a210278945346995966c/examples/TGV-Taylor-Green-vortex/TGV_Re5000.dat) dataset.

Against this reference DNS:

- Kinetic-energy relative L2 error: 3.59%
- Total-dissipation relative L2 error: 11.31%
- Dissipation-peak relative error: 3.16%
- Mean divergence remained approximately $10^{-14}$

The same static Smagorinsky case was compared with Incompact3d:

- Kinetic-energy relative L2 difference: $1.3\times10^{-6}$
- Total-dissipation relative L2 difference: $7.3\times10^{-5}$

The example uses `interpl_scheme = 'aggressive'`, corresponding to Incompact3d's `ipinter=3`, so both solvers use the same midpoint interpolation scheme.

comparison

### DIFF
--- a/docs/source/user/post-processing/monitoring.rst
+++ b/docs/source/user/post-processing/monitoring.rst
@@ -13,7 +13,7 @@ The file is a comma-separated CSV with a header line:
 
 .. code-block:: text
 
-   # time, enstrophy, div_u_max, div_u_mean
+   # time, kinetic_energy, enstrophy, div_u_max, div_u_mean
 
 Each subsequent row contains one record per output step.
 
@@ -30,6 +30,9 @@ Quantities
    * - ``time``
      - Simulation time
      - :math:`t`
+   * - ``kinetic_energy``
+     - Spatially-averaged kinetic energy
+     - :math:`E_k = \frac{1}{2N} \sum |\mathbf{u}|^2`
    * - ``enstrophy``
      - Spatially-averaged enstrophy
      - :math:`\mathcal{E} = \frac{1}{2N} \sum |\nabla \times \mathbf{u}|^2`
@@ -39,6 +42,15 @@ Quantities
    * - ``div_u_mean``
      - Mean of :math:`|\nabla \cdot \mathbf{u}|`
      - Divergence-free check
+
+.. note::
+
+   For a freely decaying periodic flow such as Taylor--Green vortex, the
+   kinetic-energy decay rate is
+   :math:`\varepsilon = -\mathrm{d}E_k/\mathrm{d}t`.  In LES this includes
+   the effects of molecular, subgrid-scale, and any numerical dissipation,
+   whereas an estimate based only on enstrophy does not include the
+   subgrid-scale contribution.
 
 .. note::
 
@@ -58,19 +70,23 @@ Example: plotting with Python
    import pandas as pd
    import matplotlib.pyplot as plt
 
-   columns = ["time", "enstrophy", "div_u_max", "div_u_mean"]
+   columns = ["time", "kinetic_energy", "enstrophy",
+              "div_u_max", "div_u_mean"]
    df = pd.read_csv("monitoring.csv", comment="#", names=columns)
 
-   fig, axes = plt.subplots(3, 1, figsize=(8, 8), sharex=True)
+   fig, axes = plt.subplots(4, 1, figsize=(8, 10), sharex=True)
 
-   axes[0].plot(df["time"], df["enstrophy"])
-   axes[0].set_ylabel("Enstrophy")
+   axes[0].plot(df["time"], df["kinetic_energy"])
+   axes[0].set_ylabel("Kinetic energy")
 
-   axes[1].plot(df["time"], df["div_u_max"])
-   axes[1].set_ylabel("div(u) max")
+   axes[1].plot(df["time"], df["enstrophy"])
+   axes[1].set_ylabel("Enstrophy")
 
-   axes[2].plot(df["time"], df["div_u_mean"])
-   axes[2].set_ylabel("div(u) mean")
+   axes[2].plot(df["time"], df["div_u_max"])
+   axes[2].set_ylabel("div(u) max")
+
+   axes[3].plot(df["time"], df["div_u_mean"])
+   axes[3].set_ylabel("div(u) mean")
 
    axes[-1].set_xlabel("Time")
 
--- a/src/les/les.f90
+++ b/src/les/les.f90
@@ -27,6 +27,7 @@ module m_les
     class(field_t), pointer :: nut => null()
     class(field_t), pointer :: mixing_length_sq => null()
   contains
+    procedure :: mixing_length
     procedure :: nut_from_gradient
     procedure :: compute_nut
     procedure :: apply_sgs_stress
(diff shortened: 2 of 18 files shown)


src/les/les.f90
```diff
@@ -0,0 +1,426 @@
...
+
+    call compute_velocity_gradients( &
+      backend, u, v, w, xdirps, ydirps, zdirps, &
+      dudx, dudy, dudz, dvdx, dvdy, dvdz, dwdx, dwdy, dwdz)
+    call backend%compute_smagorinsky_nut( &
+      self%nut, self%mixing_length_sq, &
+      dudx, dudy, dudz, dvdx, dvdy, dvdz, dwdx, dwdy, dwdz)
+
+    call add_normal_stress(backend, du, self%nut, dudx, xdirps)
+    call add_normal_stress(backend, dv, self%nut, dvdy, ydirps)
```
kaanolgu, line 236: ```compute_velocity_gradients -> derivative_to_x -> der1st```,
then
``` add_normal_stress / add_shear_stress -> add_stress_derivative -> derivative_to_x -> der1st```
Same operator, applied twice in a row why don't we use `der2nd`?

File: src/les/les.f90 [hunk 2]
```diff
@@ -0,0 +1,426 @@
...
+      y_lower = -0.5_dp*mesh%geo%L(2)
+
+    do k = 1, dims(3)
+      spacing(3) = spacing_at_vertex(mesh, k, 3, dims(3))
+      do j = 1, dims(2)
+        spacing(2) = spacing_at_vertex(mesh, j, 2, dims(2))
+        wall_distance = max(mesh%geo%vert_coords(j, 2) - y_lower, 0._dp)
+        do i = 1, dims(1)
+          spacing(1) = spacing_at_vertex(mesh, i, 1, dims(1))
+          delta = filter_width(spacing)
```
kaanolgu: The delta calculation exists twice, at `les.f90:120` and `les.f90:379` why ?

tests/verification/test_les_field.f90
```diff
@@ -0,0 +1,232 @@
...
+  class(field_t), pointer :: u, v, w, rhs_u, rhs_v, rhs_w
+
+  integer, parameter :: dims_global(3) = [32, 33, 32]
+  integer, parameter :: nproc_dir(3) = [1, 1, 1]
+  real(dp), parameter :: lengths(3) = [1._dp, 1._dp, 1._dp]
+  real(dp), parameter :: shear = 2.5_dp
+  real(dp), parameter :: curvature = 1.75_dp
+  character(len=9), parameter :: bc_periodic(2) = &
+    ['periodic ', 'periodic ']
+  character(len=9), parameter :: bc_wall(2) = &
```
kaanolgu, line 49: The wall damping is not exercised by the TGV case (`wall_damping = .false.`). The only place it runs is here, which sets bc_wall in y purely to have a wall present. Does LES support `channel` case as well if yes this should also be noted or an example of channel with LES could be produced as well?

docs/source/user/post-processing/monitoring.rst
```diff
@@ -40,6 +43,15 @@ Quantities
      - Mean of :math:`|\nabla \cdot \mathbf{u}|`
      - Divergence-free check
 
+.. note::
+
+   For a freely decaying periodic flow such as Taylor--Green vortex, the
```
kaanolgu, line 48: We could compare the kinetic-energy decay rate against a reference data to make sure that the run is OK, but the PR contains no reference data. Could we have some simple reference data as well to test against?

kaanolgu (changes requested):

src/les/les.f90
ia267: The `der2nd` is a constant-coefficient operator. The SGS term has a variable coefficient `nu_t` (which is a field that depends on all 9 gradients through `|S|`), sandwiched between 2 derivatives so it needs both. This is equivalent to the conservative form used in Incompact3d (`iconserv = 1`).

But what you suggested is right for the non-conservative form (`iconserv = 0` in Incompcat3d - this is the default) , where we can use what you're suggesting here. It applies the product rule analytically first then discretise, where we can move the non-constant `nu_t` out of the the derivative (then we can use `der2nd`). But this re-arrangement uses the principle of continuity to get rid of a term. This is exact in the continuum, however discretely before pressure project the divergence isn't exactly zero, so this form will carry a small error (on top of not conserving momentum discretely). At the same time the conservative form also has its own weakness, 2 compound first derivatives give an operator that vanishes at the Nyquist wavenumber so it can't damp the 2delta mode the SGS model exists to remove. 

At the moment I have only implemented the conservative form because the atmospheric boundary layer wall model needs it. I'll raise the non-conservative form as a separate issue. #353

src/les/les.f90
ia267: Yes this is a good point, these 2 sections are repeating the same code. The `nut_from_gradient` is pointwise and `intialise_mixing_length` is field-wise. However we can use a common code for both and I've implemented that now.

tests/verification/test_les_field.f90
ia267: Yes LES closure is case agnostic - so we can include an LES version for each of the examples. Wall damping is not used for TGV as it doesn't have walls but it will be used for the atmospheric boundary layer.

The wall distance is currently measured from the lower y boundary only (matching Incompact3d's ABL). This will be wrong for a channel LES example that uses. Incompact3d doesn't use wall damping for the LES case - but it may be useful for us so I'll open an issue. #354

docs/source/user/post-processing/monitoring.rst
ia267: I compared against the reference DNS data both kinetic energy L2 error is 3.59%. The reference data source is given in the PR intro.

In terms of adding reference data to the repo itself, I have added this to the regression test #341. But this is for the DNS case (golden from x3d2). When we add more cases to regression testing we can add LES cases (but this will compare against an LES golden). In terms of validation for the PR I believe the current comparison against both a reference DNS and Incompact3d LES is sufficient.

kaanolgu (approved):

PR #339, opened 2026-08-11
Fix vtk.xml attribute missing on restart runs

### LINKED ISSUE #333: ADIOS2 vtk.xml schema attribute not written on restarted runs
Snapshot output from a restarted run produces a valid BP5 file containing all field variables, but without the vtk.xml attribute. ParaView's VTX reader then fails to open it. Fresh runs of the same case write the attribute correctly.

Suggested fix: define the vtk.xml attribute unconditionally at IO setup, independent of the restart flag. Since it's derived from grid dimensions and mesh spacing, both known at initialisation in either case, there should be no ordering constraint.

Workaround: writing the schema by hand into the .bp directory as a vtk.xml file restores loading - the reader checks for the file as well as the attribute

ia267:
Currently `vtk.xml` is only written for the very first snapshot of a fresh run. On restart runs, the first snapshot happens at a later timestep, so the condition `timestep == self%config%snapshot_freq` is never met. Without a valid `vtk.xml` attribute, ParaView's VTX reader cannot open the BP5 file.

Fix: introduce a `first_step_of_run` flag derived from the file-open state, and write the attribute on the first snapshot of the current run (fresh or restarted) instead of at a fixed timestep. This is safe because `vtk.xml` is derived from grid dimensions and mesh spacing, both known at initialisation in either case, so there is no ordering constraint.

Closes #333

### DIFF
--- a/src/io/snapshot_manager.f90
+++ b/src/io/snapshot_manager.f90
@@ -6,8 +6,9 @@ module m_snapshot_manager
 !! data to files intended for analysis and visualisation
 !! Unlike checkpoints, which are always full-resolution for exact restarts,
 !! snapshots can be strided to reduce file size.
-  use mpi, only: MPI_COMM_WORLD, MPI_Comm_rank
-  use m_common, only: dp, i8, DIR_C, VERT, get_argument
+  use mpi, only: MPI_COMM_WORLD, MPI_Comm_rank, MPI_Allreduce, MPI_IN_PLACE, &
+                 MPI_MIN, MPI_SUCCESS
+  use m_common, only: dp, i8, MPI_X3D2_DP, DIR_C, VERT, get_argument
   use m_field, only: field_t
   use m_solver, only: solver_t
   use m_io_session, only: writer_session_t
@@ -120,7 +121,7 @@ contains
     real(dp), dimension(3) :: origin, original_spacing, output_spacing
     real(dp) :: simulation_time
     logical :: snapshot_uses_stride = .true.
-    logical :: first_step_of_run
+    logical :: is_first_snapshot_of_run
     integer :: i
 
     if (self%config%snapshot_freq <= 0) return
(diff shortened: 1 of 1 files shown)


kaanolgu (commented):
- Has this been run on more than one rank on a real restart? 

- `first_step_of_run` is derived from `is_snapshot_file_open`, so it means "the file was not open when we entered", not anything about the run. Could it  be better to rename it as `first_snapshot_of_run`

ia267:
> * Has this been run on more than one rank on a real restart?

Good point regarding multiple ranks - no it didn't work, I've now modified to generate global origin across multiple ranks and define vtk.xml attribute on every rank. It requires a one-time three-value `MPI_Allreduce` (it reduces 3 floating-point values and exactly once per code execution.

> * `first_step_of_run` is derived from `is_snapshot_file_open`, so it means "the file was not open when we entered", not anything about the run. Could it  be better to rename it as `first_snapshot_of_run`

I've renamed it to `is_first_snapshot_of_run`

Tested on a multiple rank case and now able to load a snapshot file in ParaView after restart.

ia267:
I have verified that restarts are OK for multiple ranks.

#### Multi-rank restart validation at timestep 4

| Backend | Variable | Before restart | After restart | Absolute difference |
|---|---|---:|---:|---:|
| OMP | Enstrophy | `0.37499502345493219` | `0.37499502345493219` | `0.0` |
| OMP | Div u max | `8.9428898314425354e-14` | `8.9428898314425354e-14` | `0.0` |
| OMP | Div u mean | `5.3920377527920394e-15` | `5.3920377527920394e-15` | `0.0` |
| CUDA | Enstrophy | `0.3749950234549316` | `0.3749950234549330` | `1.4e-15` |
| CUDA | Div u max | `9.1648063649859418e-14` | `9.1648063649859418e-14` | `0.0` |
| CUDA | Div u mean | `4.7622317381795163e-15` | `4.7622317381795210e-15` | `4.7e-30` |

The multi-rank restart results agree for both OMP and CUDA. OMP reproduces the reported diagnostics exactly. CUDA shows only round-off-level differences in enstrophy and mean divergence because these quantities use `atomicAdd` reductions, whose accumulation order can vary between GPU kernel executions. Since floating-point addition is not associative, this can change the final few bits. The maximum divergence is identical because the maximum operation is order-independent. These differences are expected GPU reduction behaviour and do not indicate an error in the restarted solution.

kaanolgu (approved):

PR #340, opened 2026-08-13
Update 2DECOMP-FFT to v2.0.3.2

ia267:
Update the 2DECOMP-FFT dependency from `v2.0.3.1` to `v2.0.3.2`, which includes the fix for the non-portable -`march=native` flag in NVHPC builds.

This restores compatibility with non-x86 platforms and older NVIDIA HPC SDK versions.

Validation:
- CUDA configure and build passed.
- FFT and Poisson tests passed.


kaanolgu (approved):
LGTM!

PR #341, opened 2026-08-14
add TGV regression suite with golden monitoring comparison

### LINKED ISSUE #232: Regression Test Cases
It would be great to have an example scenario with expected outputs to compare if our build or modifications are correct. For FP there should be some variations but we could agree on some range of tolerance. The Spack package for the project would be great addition for the ease of use and P3 analysis given that we support several backends and different hardware ( with new additions too) 

- [ ] Creating a test case with expected output
- [ ] Spack package for ease of installation 
- [ ] Ramble or ReFrame packages for regression testing

ia267:
Adds full-solver regression testing for x3d2. Until now the suite had unit and verification tests but nothing that catches unintended changes to the numbers a real run produces. This uses a TGV case whose `monitoring.csv` is diffed against a committed golden file, at two tiers:

| Tier | Trigger | Case | Purpose |
|---|---|---|---|
| Smoke | every PR (existing CI, Debug) | `tgv_smoke` — 32^3, ~100 steps (~2 s) | catch crashes, NaNs, gross numerical drift |
| Nightly | `schedule` cron (new workflow) | `tgv_nightly` — 64^3, 2000 steps (~3 min) | catch subtle drift from any merge; report timing |

#### Changes
- `tests/regression/compare_monitoring.py` - dependency-free golden comparator:
  - per-column, name-matched `allclose`-style comparison (`|p-g| <= atol + rtol·|g|`);
  - requires every golden column to exist in the produced output;
  - rejects non-finite values (`NaN`/`Inf`);
  - checks time-column alignment for clearer failures;
  - supports `--bless` to regenerate a golden after an intentional numerical change.
- `tgv_smoke.x3d` / `tgv_nightly.x3d` + their `*_golden_omp.csv` (tiny text, ~1–2 KB each).
- `define_regression_case_test()` in `tests/CMakeLists.txt` - runs the `xcompact` binary then the comparator. Guarded to a CPU/OMP build, because the goldens are OMP output while the shared `xcompact` target compiles as CUDA under NVHPC. Nightly case is behind `-DWITH_NIGHTLY_REGRESSION=ON` so PR CI stays fast.
- `build_and_test.yml` - the Debug step now also runs the `regression` label.
- `nightly.yml` - scheduled workflow running the full regression, with per-step timing surfaced as a non-gating metric (CI timing is too noisy to fail on).

#### Notes

- Tolerances: `rtol=1e-4`, `atol=1e-10`. Enstrophy is the real signal; the near-zero divergence columns are checked primarily through the absolute tolerance.
- Golden portability (first-run watch item): the goldens were blessed locally with GCC 13; PR CI is GCC 11 at `-O0`. This *should* stay within `rtol=1e-4`. If the smoke diff fails on first run, re-bless in the CI environment (temporary bless step --> upload artifact --> commit) rather than loosening the tolerance blindly.
- Kinetic energy: once #338 merges and adds a `kinetic_energy` column to `monitoring.csv`, re-bless to bring it under regression coverage - the comparator picks up new columns automatically once they exist in the golden.
- Note that we are still only using CPU runners - for GPU coverage we need a self-hosted GPU runner plus a blessed `*_golden_cuda.csv`

#### How to (re)bless a golden

```bash
# run the case, then accept its output as the new reference
mpirun --oversubscribe -np 2 ./xcompact tests/regression/tgv_smoke.x3d
python3 tests/regression/compare_monitoring.py golden \
  --produced monitoring.csv \
  --golden tests/regression/tgv_smoke_golden_omp.csv --bless
```

#### Testing

- `ctest -L regression` passes locally on a GNU/double CPU build: `regression_tgv_smoke_omp_2` (~2 s) and `regression_tgv_nightly_omp_2` (~188 s).
- Validated the nightly workflow end-to-end via a temporary `pull_request` trigger: run [#32018214659](https://github.com/xcompact3d/x3d2/actions/runs/32018214659)  passed (5m47s) — build, `ctest -L regression`, timing report, and artifact upload  all green. Temporary trigger reverted in 4940ce5, so `main` will get only `schedule` (nightly 02:30 UTC) + `workflow_dispatch`.

Closes #232


kaanolgu (changes requested):
https://github.com/xcompact3d/x3d2/actions/runs/32018214659/job/95352130515

```
Run set -o pipefail
Internal ctest changing into directory: /home/runner/work/x3d2/x3d2/build-nightly
Test project /home/runner/work/x3d2/x3d2/build-nightly
    Start 22: regression_tgv_smoke_omp_2
1/2 Test #22: regression_tgv_smoke_omp_2 .......   Passed    2.24 sec
    Start 23: regression_tgv_nightly_omp_2
2/2 Test #23: regression_tgv_nightly_omp_2 .....   Passed  243.59 sec

100% tests passed, 0 tests failed out of 2

Label Time Summary:
omp           = 245.84 sec*proc (2 tests)
regression    = 245.84 sec*proc (2 tests)

Total Test time (real) = 245.87 sec
```

We could add verification of the execution times too that would enable us to notice if any change merged in is causing a slow down. 

It might be good to have 8% tolerance (arbitrary) since 5% could be variance between run to run ?

ia267:
> We could add verification of the execution times too that would enable us to notice if any change merged in is causing a slow down.
> 
> It might be good to have 8% tolerance (arbitrary) since 5% could be variance between run to run ?

I have avoided failing on execution times since it may vary significantly on the shared GitHub runners (we may not be be able to even guarantee an 8% tolerance). However, I've added a change that will report execution times (written to the workflow run's GitHub Actions summary), allowing us to track them manually. Once we have enough data we can introduce a threshold based on this. Created issue #357 to look into this.

The proper implementation will be to do hard performance thresholds on a self-hosted runners ideally using repeated tests.

kaanolgu (approved):

PR #344, opened 2026-08-19
Import DECOMP_2D_COMM_CART_X from decomp_2d_mpi

ia267:
Import `DECOMP_2D_COMM_CART_X` from `decomp_2d_mpi`, where it is defined, instead of re-exporting it through `decomp_2d` in `decomp_2decompfft.f90`.

This makes the code more robust - some 2decomp&fft builds/environments do not re-export `DECOMP_2D_COMM_CART_X` via `decomp_2d`, causing a build failure.


ia267:
Closing because this change is incompatible with the upstream 2decomp&fft `v2.0.3.2` dependency pinned by x3d2.

PR #346, opened 2026-08-20
Refactor tests around shared MPI and backend helpers

### LINKED ISSUE #293: Simplifying the test cases
The test cases have lots of common calls which are shared by many tests. We could generate a `test_common.f90` or `test_base.f90`
 to store these common routines for all of the tests. For example this is shared by the tests:
``` fortran
  subroutine initialise_mpi()
    call MPI_Init(ierr)
    call MPI_Comm_rank(MPI_COMM_WORLD, nrank, ierr)
    call MPI_Comm_size(MPI_COMM_WORLD, nproc, ierr)

    if (nrank == 0) print *, 'Performance benchmark with', nproc, 'ranks'

    pnext = modulo(nrank - nproc + 1, nproc)
    pprev = modulo(nrank - 1, nproc)
  end subroutine initialise_mpi 
  ```

This way we could reduce the complexity of standalone tests

### PR DESCRIPTION
**Pull request by ia267**
Closes #293 

- Centralise MPI initialisation, result reporting, and finalisation in shared test utilities
- Add shared runtime setup for backend, allocator, and CUDA device selection while retaining OMP, CUDA, and OMP-target support
- Use backend-neutral field access and solver operations to remove repeated preprocessor and `select type` blocks
- Isolate backend-specific Fortran modules in the test build.

#### Testing

Built and ran the affected GNU/OpenMP and NVHPC/CUDA tests locally. OMP-target paths are covered by CI.

### DIFF
--- a/src/backend/backend_runtime.f90
+++ b/src/backend/backend_runtime.f90
@@ -56,6 +56,27 @@ module m_backend_runtime
 
 contains
 
+#ifdef CUDA
+  subroutine select_cuda_device(nrank, devnum)
+    !! Select a CUDA device round-robin by MPI rank.  The optional result
+    !! returns the device that CUDA reports as current after selection.
+    integer, intent(in) :: nrank
+    integer, optional, intent(out) :: devnum
+
+    integer :: ierr, ndevs, selected_device
+
+    ierr = cudaGetDeviceCount(ndevs)
+    if (ndevs < 1) then
+      error stop 'select_cuda_device: no CUDA devices available'
+    end if
+
+    ierr = cudaSetDevice(mod(nrank, ndevs))
+    ierr = cudaGetDevice(selected_device)
+
+    if (present(devnum)) devnum = selected_device
+  end subroutine select_cuda_device
+#endif
+
   subroutine init(self, mesh, separate_host_allocator)
     class(backend_runtime_t), target, intent(inout) :: self
     type(mesh_t), target, intent(inout) :: mesh
@@ -67,7 +88,7 @@ contains
 #endif
 
 #ifdef CUDA
-    integer :: ierr, nrank, ndevs, devnum
+    integer :: ierr, nrank
 #endif
 
     dims = mesh%grid%vert_dims
@@ -80,12 +101,7 @@ contains
 
 #ifdef CUDA
     call MPI_Comm_rank(MPI_COMM_WORLD, nrank, ierr)
-    ierr = cudaGetDeviceCount(ndevs)
-    if (ndevs < 1) then
-      error stop 'backend_runtime_t%init: no CUDA devices available'
-    end if
-    ierr = cudaSetDevice(mod(nrank, ndevs))
-    ierr = cudaGetDevice(devnum)
+    call select_cuda_device(nrank)
 
     self%backend_name = 'CUDA'
     self%cuda_allocator = cuda_allocator_t(dims, SZ)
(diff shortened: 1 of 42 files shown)

### CODE REVIEW

File: tests/common/test_utils.f90 [hunk 1]
```diff
@@ -1,14 +1,94 @@
...
+    do_finalize = .true.
+    if (present(finalize_mpi)) do_finalize = finalize_mpi
+
+    if (allpass) then
+      if (is_root) write (stderr, '(a)') 'ALL TESTS PASSED SUCCESSFULLY.'
+    else
+      error stop 'SOME TESTS FAILED.'
+    end if
+
+    if (do_finalize) call MPI_Finalize(ierr)
```
kaanolgu: if rank 1 fails and rank 0 passes, rank 0 executes `MPI_Finalize` while rank 1 does error stop.

tests/CMakeLists.txt
```diff
@@ -23,7 +27,10 @@ function(define_test testfile np backend)
   get_filename_component(test_name ${testfile} NAME_WE)
   string(CONCAT test_name ${test_name} _ ${backend} _ ${np})
 
-  add_executable(${test_name} ${testfile})
+  add_executable(${test_name} ${testfile}
+                 ${CMAKE_SOURCE_DIR}/src/backend/backend_runtime.f90)
```
kaanolgu: This adds the file as a source to every single test target but it is already included in the xcompact build so there is duplicate work. 

Might be better to move it to tests/common/ and fold it into the existing test utils library (which is already built once and already has a module directory).

src/backend/backend_runtime.f90
```diff
@@ -0,0 +1,146 @@
...
+    integer :: dims(3)
+#if !defined(CUDA) && !defined(OMP_TGT)
+    logical :: need_separate_host_allocator
+#endif
+
+#ifdef CUDA
+    integer :: ierr, nrank
+#endif
+
+    dims = mesh%grid%vert_dims
```
kaanolgu:
```suggestion
    dims = mesh%get_dims(VERT)
```

tests/performance/perf_thom.f90
```diff
@@ -1,188 +1,153 @@
...
-    n_iters = 500
-#endif
+    if (backend_is_cuda) then
+      backend_label = 'cuda'
+      periodic_bw = 6.0_dp
+      dirichlet_bw = 4.0_dp
+      n_groups = 512*512/backend_sz
+      n_iters = 1000
+    else
+      backend_label = 'omp '
```
kaanolgu:
```suggestion
      backend_label = 'omp'
```

kaanolgu (changes requested):

tests/common/test_utils.f90
ia267: Good catch - I will aggregate `allpass` across the ranks first.

tests/CMakeLists.txt
ia267: Agreed - currently it was compiling once per test.

- Moved `backend_runtime.f90` to `tests/common/backend_runtime.f90`
- Removed it from every test executable's source list
- Added a shared runtime library compiled once per backend

It cannot safely be folded into the single backend-independent `x3d2_test_utils` library because one build can contain test targets for multiple backends, while `backend_runtime.f90` is compiled different for each backend using preprocessor definitions such as `CUDA` and `OMP_TGT`

kaanolgu (approved):
LGTM!

PR #349, opened 2026-08-21
Add multi GPU support for the cylinder case CUDA backend

kaanolgu:
Adds the support to run the 100 case (dirichlet in x, periodic in y and z) across multiple ranks on the CUDA backend. Until now poisson solver stopped with `Multiple ranks are not yet supported for non-periodic BCs!` for any non periodic case.

Rather than gathering the whole spectral field, each rank fetches a mirror of its own slab and reads the partner out of that:
- The single postprocess kernel is split into three launches, `process_spectral_100_fw`, `process_spectral_100_pair_fw` and  `process_spectral_100_pair_bw`, which leaves room for an exchange either  side of the Poisson solve
- `exchange_mirror_100` fills `c_mirror_dev` with the other rank's slab in  one `MPI_Sendrecv`, plus a plane sized `MPI_Sendrecv` for the single mode that falls outside the mirrored range. Slab `r` of `P` pairs with slab `P - 1 - r`, and the extra plane comes from slab `P - r`, so both partners are symmetric and nothing has to be sent back afterwards
- One GPU execution keeps a dedicated single kernel, `process_spectral_100`, which reads straight out of `div_u` with no exchange


kaanolgu:
@x3d2bot fix style

src/backend/cuda/poisson_fft.f90
```diff
@@ -396,6 +432,24 @@ function init(mesh, xdirps, ydirps, zdirps, lowmem) &
...
+        error stop 'The 100 case on multiple ranks needs cuFFTMp, plain &
+                    &cuFFT cannot decompose the transform.'
+      end if
+      if (mod(dims_loc(1), mesh%par%nproc_dir(3)) /= 0) then
+        error stop 'The 100 case on multiple ranks needs the number of x &
+                    &cells to divide by the number of ranks.'
+      end if
+      if (mesh%par%nrank /= mesh%par%nrank_dir(3)) then
+        error stop 'The 100 case mirror exchange assumes the MPI rank &
+                    &matches the z slab index.'
```
ia267: In line 234 we print if the decomposition is not 1D `if (mesh%par%nproc_dir(2) /= 1) print *, 'nproc_dir in y-dir must be 1'` and here we independently check every rank. If for example `nproc_dir` was `[1,2,N]` some ranks will give `error stop` while others will continue into FFT/MPI operations. We need to validate `nproc_dir == [1, 1, nproc]` uniformly before creating cuFFTMp plans

src/backend/cuda/poisson_fft.f90
```diff
@@ -805,20 +859,133 @@ subroutine fft_postprocess_100_cuda(self)
...
+      mpi_cplx = MPI_DOUBLE_COMPLEX
+    end if
+
+    n_slab = self%nx_spec*self%ny_spec*self%nz_spec
+    n_plane = self%nx_spec*self%nz_spec
+
+    ! Make sure the staged kernel and the packing above have landed before
+    ! MPI reads the device buffers.
+    if (self%mesh%par%nproc > 1) then
+      ierr = cudaStreamSynchronize(default_stream)
```
ia267: CUDA sync here and the two `MPI_Sendrecv` calls below will overwrite `ierr` without checking it. A failed synchronisation can expose unfinished buffers to MPI, while an MPI error can leave stale mirror data feeding the pressure solve. Check every results immediately through shared CUDA/MPI error helpers.

src/backend/cuda/kernels/spectral_processing.f90
```diff
@@ -705,6 +705,439 @@ attributes(global) subroutine process_spectral_010_bw( &
 
   end subroutine process_spectral_010_bw
 
+  attributes(global) subroutine process_spectral_100( &
```
ia267: We are duplicating around 160 lines from `process_spectral_010`. Before this change the 100 case re-used the 010 kernel with swapped arguments. The new staged kernels repeat the same normalisation and forward/backward transforms again. This creates 3 representations of the same maths that can drift apart. Keep the existing 010 kernel for the single-rank path and extract the shared pointwise transformations into CUDA device procedures for the staged path.

src/backend/cuda/poisson_fft.f90
```diff
@@ -805,20 +859,133 @@ subroutine fft_postprocess_100_cuda(self)
...
+      c_dev, self%c_mirror_dev, self%c_plane_recv_dev, &
+      self%nx_spec, self%ny_spec, self%sp_st(2), &  ! dims and dim2 offset
+      self%ny_glob, self%nz_glob, &  ! SWAP nx <-> ny
+      self%ay_dev, self%by_dev, &  ! dim1 uses Y coefficients
+      self%ax_dev, self%bx_dev, &  ! dim2 uses X coefficients
+      self%az_dev, self%bz_dev &   ! dim3 uses Z coefficients
       )
   end subroutine fft_postprocess_100_cuda
 
+  subroutine exchange_mirror_100(self, c_dev)
```
ia267, line 921: Now this module owns the stream synchronisation, MPI datatype selection, tags and `MPI_Sendrecv`, even though we have `m_cuda_sendrecv` already used as the CUDA communication abstraction. Can you add a generic complex-device exchange operation there? This would centralise synchronisation, error handling, tags and later make the OpenMP implementation cleaner.

src/backend/cuda/poisson_fft.f90
```diff
@@ -396,6 +432,24 @@ function init(mesh, xdirps, ydirps, zdirps, lowmem) &
       end if
     end if
 
+    ! exchange_mirror_100 relies on cuFFTMp handing out the dim2 modes as
+    ! equal sized blocks in rank order, so check that here rather than
+    ! producing a quietly wrong pressure field.
+    if (poisson_fft%is_100_case .and. mesh%par%nproc > 1) then
+      if (.not. poisson_fft%use_cufftmp) then
+        error stop 'The 100 case on multiple ranks needs cuFFTMp, plain &
```
ia267: Could improve this slightly we if use a shared-error routing calling `MPI_Abort` rather than plain `error stop` to guarantee that the complete MPI job terminates cleanly.

ia267 (changes requested):
Requires some changes 

Also could we add a multi-GPU test case (that we can run locally). 

We can add a focussed `test_poisson_100`, with 2 or 4 ranks. Label it `cuda;multigpu`. Or something like this setup conceptually:


```cmake
define_verification_test(
  verification/test_poisson_bc.f90 1 cuda --config all
)

define_verification_test(
  verification/test_poisson_bc.f90 ${CMAKE_CTEST_NPROCS} cuda --config 100
)
```

Two MPi ranks sharing a single GPU can provide a smoke test of the mirror-exchange path.

src/backend/cuda/kernels/spectral_processing.f90
kaanolgu: There are 2 reasons for not reusing the 010: 
- It might change in the future with multi rank channel case high chance that they diverge
- I didn't want to cause a confusion like "why are we using 010 cases spectral process in 100" 

I could rename it to a common name and reuse it in 010 and 100, by having separate cases if things diverge. Do you think that way is better?

ia267: Yes that would be better - the risk with the current approach is that we update something in one part and forget to do the same in the other. But if it's shared code then it can be re-used.

kaanolgu:
Addressed the comments, except the test routine modification I will create an issue and work out that one in a separate PR

kaanolgu:
@x3d2bot fix style

src/backend/cuda/poisson_fft.f90
```diff
@@ -805,20 +860,144 @@ subroutine fft_postprocess_100_cuda(self)
...
+    if (self%mirror_plane_rank == nrank) then
+      self%c_plane_recv_dev = self%c_plane_send_dev
+    else
+      call MPI_Sendrecv(self%c_plane_send_dev, n_plane, mpi_cplx, &
+                        self%mirror_plane_rank, tag_plane, &
+                        self%c_plane_recv_dev, n_plane, mpi_cplx, &
+                        self%mirror_plane_rank, tag_plane, &
+                        MPI_COMM_WORLD, MPI_STATUS_IGNORE, ierr_mpi)
+      if (ierr_mpi /= MPI_SUCCESS) then
+        error stop 'The 100 mirror slab exchange failed.'
```
ia267: This should say `the 100 mirror plane exchange failed`

Could we use `MPI_Abort` here instead of `error stop`  for the CUDA synchronisation failure so that other ranks cannot remain blocked in `MPI_Sendrecv`.  
```suggestion
  write(stderr, '(a,i0)') &
    'The 100 mirror plane exchange failed: ', ierr_mpi
  call MPI_Abort(MPI_COMM_WORLD, 1, ierr_abort)
```

src/backend/cuda/poisson_fft.f90
```diff
@@ -805,20 +860,144 @@ subroutine fft_postprocess_100_cuda(self)
...
+
+    n_slab = self%nx_spec*self%ny_spec*self%nz_spec
+    n_plane = self%nx_spec*self%nz_spec
+
+    ! Make sure the staged kernel and the packing above have landed before
+    ! MPI reads the device buffers. This is only reached on multiple ranks,
+    ! fft_postprocess_100_cuda returns early for the single rank path.
+    ierr = cudaStreamSynchronize(default_stream)
+    if (ierr /= cudaSuccess) then
+      error stop 'Stream synchronisation before the 100 mirror exchange &
```
ia267: Could we use `MPI_Abort` here instead of `error stop`  for the CUDA synchronisation failure so that other ranks cannot remain blocked in `MPI_Sendrecv`.  

This issue exists even in `sendrecv.f90` where the `ierr` is overwritten by `MPI_Waitall` - this needs to be fixed in a separate issue. #356 

```suggestion
  write(stderr, '(a,i0)') &
    'CUDA synchronisation before 100 mirror exchange failed: ', ierr
  call MPI_Abort(MPI_COMM_WORLD, 1, ierr_abort)
```

src/backend/cuda/poisson_fft.f90
```diff
@@ -805,20 +860,144 @@ subroutine fft_postprocess_100_cuda(self)
...
+    if (self%mirror_slab_rank == nrank) then
+      self%c_mirror_dev = c_dev
+    else
+      call MPI_Sendrecv(c_dev, n_slab, mpi_cplx, &
+                        self%mirror_slab_rank, tag_slab, &
+                        self%c_mirror_dev, n_slab, mpi_cplx, &
+                        self%mirror_slab_rank, tag_slab, &
+                        MPI_COMM_WORLD, MPI_STATUS_IGNORE, ierr_mpi)
+      if (ierr_mpi /= MPI_SUCCESS) then
+        error stop 'The 100 mirror slab exchange failed.'
```
ia267: Could we use `MPI_Abort` here instead of `error stop`  for the CUDA synchronisation failure so that other ranks cannot remain blocked in `MPI_Sendrecv`.  
```suggestion
        write(stderr, '(a,i0)') &
           'The 100 mirror slab exchange failed: ', ierr_mpi
        call MPI_Abort(MPI_COMM_WORLD, 1, ierr_abort)
```

ia267 (changes requested):
The `MPI_Abort` stuff needs to be implemented.

#355 will add the mulit-GPU tests. But I think we should add one this to the tests so that we can get automated local testing for multip GPU? 

you can add this in `tests/CMakeLists.txt` immediately after the existing one-rank CUDA test:
`define_verification_test(verification/test_poisson_bc.f90 2 cuda)`

ia267 (approved):
Approved. This closes #355 

The remaining `MPI_Abort` changes will be handled in #356

PR #358, opened 2026-08-27
Add OpenMP CPU Support for the Cylinder (100) Case

kaanolgu:
Adds OpenMP backend support for the cylinder case with `100` boundary conditions
(Dirichlet in x, periodic in y and z), including multi-rank execution.

This is the OpenMP counterpart to #349, which added multi-GPU support for the same
case on the CUDA backend.


kaanolgu:
Thanks to @mathrack  for pointing out that decomp_2d_fft_get_ph is already in the source code of 2decomp https://github.com/xcompact3d/x3d2/pull/358/commits/5455005850ea07d575b52de325418bdc5e6b4707

kaanolgu:
@x3d2bot fix style

src/backend/omp/poisson_fft.f90
```diff
@@ -195,11 +341,44 @@ subroutine fft_postprocess_010_omp(self)
...
+    !! into this one fused pass.
     implicit none
 
     class(omp_poisson_fft_t) :: self
 
-    error stop 'OpenMP backend does not support fft_postprocess_100 yet!'
+    if (self%p_col > 1) then
+      ! Hop to the y-pencil, where dim2 is whole, pair there, and hop back.
+      call transpose_z_to_y(self%c_x, self%c_pair, self%sp)
+      call process_spectral_010( &
```
ia267, line 366: The comparison here uses local pencil index, but the `nx/2 + 1` and `nz/2 + 1` that appears in `spectral_processing.f90` lin 222:

```fortran
if (i == nx/2 + 1 .and. k == nz/2 + 1) div_u(i, j, k) = 0._dp
```
are global indices. This worked OK when it was single-rank, but the current multi-rank 100 implementation calls it with non-zero `x_sp_st / z_sp_st` (i.e. non-zero offsets). On multiple MPI ranks, this mode may not be set to zero because its local and global indices differ. We should apply the offset here:

```fortran
if (i + x_sp_st == nx/2 + 1 .and. k + z_sp_st == nz/2 + 1) div_u(i, j, k) = 0._dp
```

ia267 (changes requested):
- Could you add a OMP test case that tests this? For example `test_poisson_bc` should be added (restricting it to `--config 100`). Ideally the tests should cover:
   - a single rank 
   - [1, 1, nproc] testing for `p_col > 1`
   - [1, nproc, 1] testing for `p_row > 1`
   - Also current test hard-codes `[1, 1, nproc]`, so we never test `p_row > 1` redistribution path. 

I think the [1, nrpoc, 1] test is OMP specific (CUDA only uses `[1, 1, nproc]`) so the test will be specific to OMP.

- Re-using the `process_spectral_010` routine is great to avoid duplication, but the positional X/Y swaps makes this call hard to verify and easy to make mistakes. Could we hide the mapping behind a `process_spectral_100` wrapper and use keyword arguments internally? I think longer-term we need to add spectral-layout abstraction for the different BC orientations will make this easier to extend safely

kaanolgu:
> * Could you add a OMP test case that tests this? For example `test_poisson_bc` should be added (restricting it to `--config 100`). Ideally the tests should cover:
>   
>   * a single rank
>   * [1, 1, nproc] testing for `p_col > 1`
>   * [1, nproc, 1] testing for `p_row > 1`
>   * Also current test hard-codes `[1, 1, nproc]`, so we never test `p_row > 1` redistribution path.
> 
> I think the [1, nrpoc, 1] test is OMP specific (CUDA only uses `[1, 1, nproc]`) so the test will be specific to OMP.

I agree this is worth adding, but I'd prefer to address it in a separate PR. I'm currently planning an overhaul of test_poisson_bc

> * Re-using the `process_spectral_010` routine is great to avoid duplication, but the positional X/Y swaps makes this call hard to verify and easy to make mistakes. Could we hide the mapping behind a `process_spectral_100` wrapper and use keyword arguments internally? I think longer-term we need to add spectral-layout abstraction for the different BC orientations will make this easier to extend safely

Addressed in commit [3964735](https://github.com/xcompact3d/x3d2/pull/358/commits/39647351fcb2a94ed5da5bc8bc6d764d79e2206a)

kaanolgu:
@x3d2bot fix style

src/backend/omp/kernels/spectral_processing.f90
```diff
@@ -288,4 +289,37 @@ subroutine process_spectral_010( &
...
+    div_u, waves, nx_spec, ny_spec, nz_spec, x_sp_st, y_sp_st, z_sp_st, &
+    nx, ny, nz, ax, bx, ay, by, az, bz &
+    )
+    !! Post-process div U* in spectral space, for non-periodic BC in x-dir.
+    !!
+    !! This is the process_spectral_010 arrangement with x and y swapped:
+    !! the caller passes the transposed buffer (dim1 = y modes, dim2 = x
+    !! modes, dim3 = z), so this wrapper maps the global sizes nx <-> ny
+    !! and the coefficient pairs ax, bx <-> ay, by onto process_spectral_010's
+    !! non-periodic-x convention before delegating to it.
```
ia267: I think this should say `non-periodic-y convention`

ia267 (approved):
Approved

src/backend/omp/kernels/spectral_processing.f90
kaanolgu:
```suggestion
    !! non-periodic-y convention before delegating to it.
```

PR #360, opened 2026-08-27
Fix regression test suite portability and robustness

ia267:
The regression suite added by #341 works in the configuration CI runs, but it had surfaced some bugs which were invisible because of gaps in how CI runs tests.

#### Regression suite

- `div_u_max/div_u_mean` are essentially zero plus round-odff, so checking them against a stored number was just comparing noise. They're now checked against an upper bound
- The suit now runs in single-precision too, against the same reference file as double-precision. The only thing locking it was a hardcoded tolerance on the time column, too light for float32, which can't represent 0.05 exactly. A single-precision reference would actually be worse: 2 single-precision builds differ from each other more than either differs from the double-precision one. 
- Tightened tolerance from `1e-4` to `1e-10` in double-precision. Single-precision needs looser limits because its round-off accumulates over the run: 1e-4 for the short case and 1e-3 for the nightly.
- The hardcoded `mpirun`, which doesn't exist on Cray machines like ARCHER2. It now uses `srun` there
- Overwriting the reference ("blessing") now needs an explicit `--bless`. Before, a mistyped path was read as a request to create a new reference - so the test passed without comparing anything.

#### CI
- `ctest` now runs via `--test-dir` instead of `ccd build && ctest && cd ..`. When a test failed, the `cd ..` was skipped and the next command failed with an unrelated CMake error - so every GitLab test failure looked like a broken build
- Test results are published as a JUnit report, so failures show in the pipeline's Tests tab instead of only deep in the raw log
- Single-precision builds now run their tests; they were only being compiled before. One fix was needed first: `test_les.f90` used a hardcoded tolerance instead of  that adaps to precision

### DIFF
--- a/.github/workflows/build_and_test.yml
+++ b/.github/workflows/build_and_test.yml
@@ -74,9 +74,15 @@ jobs:
     - name: Build tests (single precision)
       run: make -C build-single-debug
 
+    - name: Run unit and verification tests (single precision)
+      run: ctest --test-dir build-single-debug -L "unit|verification" --output-on-failure
+
     - name: Configure Release build (single precision)
       run: FC=mpif90 cmake -S . -B build-single -DCMAKE_BUILD_TYPE=Release -DSINGLE_PREC=ON -DWITH_ADIOS2=ON -DUSE_SYSTEM_ADIOS2=ON
 
     - name: Build tests (single precision)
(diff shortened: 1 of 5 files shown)


kaanolgu (approved):
LGTM

PR #361, opened 2026-08-28
remove transfer of symmetric heap memory to MPI

### LINKED ISSUE #359: Multi-GPU 100 case crashes in exchange_mirror_100
The CUDA 100 case crashes on multiple ranks inside `exchange_mirror_100` (https://gitlab.ae.ic.ac.uk/xcompact3d/x3d2/-/jobs/690). Reproduced by `test_poisson_bc_cuda_2`, registered at `tests/CMakeLists.txt:223`:

```cmake
define_verification_test(verification/test_poisson_bc.f90 2 cuda --config 100)
```

#### Cause
On the multi-rank path `c_dev` is not an ordinary device array. `src/backend/cuda/poisson_fft.f90:847`:

```fortran
if (self%use_cufftmp) then
  call c_f_pointer(self%xtdesc%descriptor, descriptor)
  call c_f_pointer(descriptor%data(1), c_dev, [nx_spec, ny_spec, nz_spec])
```

`xtdesc` comes from `cufftXtMalloc` (line 421), which serves memory from cuFFTMp's NVSHMEM symmetric heap. `exchange_mirror_100` then passes that pointer straight to `MPI_Sendrecv` as the send buffer (line 982).

`uct_cma_ep_tx_error` is UCX's host shared-memory transport (CMA, process_vm_readv). UCX only selects it when it believes the buffer is host memory - the known failure mode for NVSHMEM symmetric-heap pointers that `cuPointerGetAttribute` does not classify as device memory. This is the first place in x3d2 that hands a cuFFTMp/NVSHMEM pointer to MPI. The existing CUDA MPI code uses the same assumed-shape `device` idiom and passes the 4-rank CUDA tests, but those buffers come from the CUDA allocator.

#### Fix
Stage the slab through a plain `cudaMalloc`'d device buffer. D2D copy out of the cuFFTMp descriptor. `MPI_Sendrecv` on the temporary, then copy back. One extra device copy per exchange. The same applies to the plane exchange at line 998 if `c_plane_send_dev` ever aliases descriptor memory.

kaanolgu:
Closes #359

### DIFF
--- a/src/backend/cuda/poisson_fft.f90
+++ b/src/backend/cuda/poisson_fft.f90
@@ -78,7 +78,7 @@ module m_cuda_poisson_fft
     real(dp), device, allocatable, dimension(:, :, :) :: r_dev
     !> Transposed real workspace for 110 case (nz, nx, ny)
     real(dp), device, allocatable, dimension(:, :, :) :: r_dev_110
-        !> dim2 mirror of the local spectral slab, used by the 100 case to
+    !> dim2 mirror of the local spectral slab, used by the 100 case to
     !> reach the paired split partner that lives on another rank
     complex(dp), device, allocatable, dimension(:, :, :) :: c_mirror_dev
     !> Staging copy of the local slab, sent in place of c_dev itself.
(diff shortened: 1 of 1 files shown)


kaanolgu:
@x3d2bot fix style

PR #363, opened 2026-09-08
Allow using pre-build decomp2d library

CFD-Xing:
Currently, it is not possible to use a pre-build decomp2d library as the decomp2d_install_dir is overwritten within cmake.

This can cause complication one some HPC system (e.g Archer2) when it is not possible to remotely fetch data on the compute node.


ia267 (approved):
Approved

PR #364, opened 2026-09-08
Unify device allocator unit tests

CFD-Xing:
Both CUDA and OMP TGT have provision to select the GPU by passing in index. It is possible to unify the device allocator unit test.

Note: Tested on tornado (CUDA) and Archer2 (OMP TGT)


ia267 (approved):
Approved

PR #365, opened 2026-09-08
Improve coverage of backend-agnostic test_reordering.f90 and remove test_cuda_reorder.f90

CFD-Xing:
The `tests/unit/test_cuda_reorder.f90` mostly duplicate those in `tests/unit/test_reordering.f90`, but uses low-level CUDA backend-specific kernel functions instead of the backeng-agnostic wrapper functions.

This PR update `tests/unit/test_reordering.f90` to cover additional tests covered in  `tests/unit/test_cuda_reorder.f90` and remove the later.


CFD-Xing:
Note: test_reordering is backend agnostic but the OMP_TGT is still not fully implemented.

ia267 (approved):
Approved

PR #366, opened 2026-09-09
Set OMP TGT SZ parameter value based on vendor (AMD vs. CUDA)

CFD-Xing:
CUDA GPUs have a warpsize of 32 while AMD GPUs have a wavefront size of 64. This PR set the SZ value based on the compiler type (NVHPC/PGI for CUDA and CRAY/FLANG for AMD).

Also enable OMP TGT reordering tests.


ia267 (approved):
Approved

PR #367, opened 2026-09-11
Update test_kinetic_energy.f90 to be fully backend agnostic along with adding OMP TGT backend

CFD-Xing:
Update test_kinetic_energy.f90 to be fully backend agnostic

Also use to test the GitLab Ci.


CFD-Xing:
Note: GitLab CI was not triggered due (probably) to improper permission when creating the access token. New PR will be required to check if the new token is working or not.

ia267 (approved):
Approved - could you please update the title of the PR and also the text to reflect the additional changes made in OMPTGT backend?

PR #389, opened 2026-09-21
Fix OMP TGT backend for NVHPC

CFD-Xing:


CFD-Xing:
The previous solution does not seem to be robust, it results in warning: @ia267 

```
NVFORTRAN-W-0287-Unrecognized OpenMP directive - is_device_ptr (/home/icl/x3d2/src/backend/omp/target/backend.f90: 376)
NVFORTRAN-W-0287-Unrecognized OpenMP directive - map (/home/icl/x3d2/src/backend/omp/target/backend.f90: 402)

```

